PR #69, opened 2019-05-23
Implement peer discovery

albrow:
This PR adds support for peer discovery via [go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht) and [go-libp2p-discovery](https://github.com/libp2p/go-libp2p-discovery). It also adds a new environment variable `USE_BOOTSTRAP_LIST`. If set to "true", it uses the IPFS default bootstrap list to bootstrap the DHT.

WIP right now because of some weird behavior when using the bootstrap list. If you set it to "false", you can test peer discovery by following these steps:

1. Create three 0x Mesh nodes (A, B, and C).
2. Manually connect A to B via the `AddPeer` RPC method.
3. Manually connect B to C via the `AddPeer` RPC method.
4. Observe that A is introduced to C through B and they will connect to each other.

### DIFF
--- a/blockwatch/watcher.go
+++ b/blockwatch/watcher.go
@@ -166,7 +166,7 @@ func (w *Watcher) pollNextBlock() error {
 		if err == ethereum.NotFound {
 			log.WithFields(log.Fields{
 				"blockNumber": nextBlockNumber,
-			}).Info("block header not found")
+			}).Trace("block header not found")
 			return nil // Noop and wait next polling interval
 		}
 		return err
--- a/core/node.go
+++ b/core/node.go
@@ -34,11 +34,11 @@ const (
 	// maxShareBatch is the maximum number of messages to share at once.
 	maxShareBatch = 10
 	// peerCountLow is the target number of peers to connect to at any given time.
-	peerCountLow = 10
+	peerCountLow = 100
 	// peerCountHigh is the maximum number of peers to be connected to. If the
 	// number of connections exceeds this number, we will prune connections until
 	// we reach peerCountLow.
-	peerCountHigh = 12
+	peerCountHigh = 110
 	// peerGraceDuration is the amount of time a newly opened connection is given
 	// before it becomes subject to pruning.
 	peerGraceDuration = 10 * time.Second
@@ -205,14 +205,15 @@ func (n *Node) Start() error {
 		return err
 	}
 
-	// Advertise ourselves for the purposes of peer discovery.
-	discovery.Advertise(n.ctx, n.routingDiscovery, n.config.RendezvousString, discovery.TTL(advertiseTTL))
-
+	// If needed, connect to all peers in the bootstrap list.
 	if n.config.UseBootstrapList {
 		if err := n.connectToBootstrapList(); err != nil {
 			return err
 		}
 	}
+
+	// Advertise ourselves for the purposes of peer discovery.
+	discovery.Advertise(n.ctx, n.routingDiscovery, n.config.RendezvousString, discovery.TTL(advertiseTTL))
 
 	return n.mainLoop()
 }
@@ -240,6 +241,13 @@ func (n *Node) connectToBootstrapList() error {
 		}()
 	}
 	wg.Wait()
+
+	// It is recommended to wait for 2 seconds after connecting to all the
+	// bootstrap peers to give time for the relevant notifees to trigger and the
+	// DHT to fully initialize.
+	// See: https://github.com/0xProject/0x-mesh/pull/69#discussion_r286849679
+	time.Sleep(2 * time.Second)
+
 	return nil
 }
 
--- a/core/node.go
+++ b/core/node.go
@@ -9,11 +9,16 @@ import (
 	"fmt"
 	"io"
 	mrand "math/rand"
+	"sync"
 	"time"
 
 	libp2p "github.com/libp2p/go-libp2p"
+	connmgr "github.com/libp2p/go-libp2p-connmgr"
 	crypto "github.com/libp2p/go-libp2p-crypto"
+	discovery "github.com/libp2p/go-libp2p-discovery"
 	host "github.com/libp2p/go-libp2p-host"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	p2pnet "github.com/libp2p/go-libp2p-net"
 	peer "github.com/libp2p/go-libp2p-peer"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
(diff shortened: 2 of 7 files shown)


blockwatch/watcher.go
```diff
@@ -166,7 +166,7 @@ func (w *Watcher) pollNextBlock() error {
 		if err == ethereum.NotFound {
 			log.WithFields(log.Fields{
 				"blockNumber": nextBlockNumber,
-			}).Info("block header not found")
+			}).Trace("block header not found")
```
albrow: @fabioberger This log was getting pretty noisy. I think it is always expected to happen while we are waiting for the next block to be mined. Are you okay with changing this to `Trace`?

core/node.go
```diff
@@ -253,6 +388,5 @@ func (n *Node) receive(ctx context.Context) (*Message, error) {
 // Close closes the Node and any active connections.
 func (n *Node) Close() error {
 	n.cancel()
-	n.sub.Cancel()
```
albrow: We're passing in `n.ctx` to `n.sub` during initialization so it now closes automatically when we call `n.cancel`.

core/node.go
```diff
@@ -143,9 +198,51 @@ func (n *Node) ID() peer.ID {
...
 func (n *Node) Start() error {
+	// Note: dht.Bootstrap has a somewhat confusing name. It doesn't automatically
+	// connect to the bootstrap peers. It just starts the background process of
+	// searching for new peers.
+	if err := n.dht.Bootstrap(n.ctx); err != nil {
+		return err
+	}
+
+	// Advertise ourselves for the purposes of peer discovery.
+	discovery.Advertise(n.ctx, n.routingDiscovery, n.config.RendezvousString, discovery.TTL(advertiseTTL))
```
raulk: At some point we'll probably want to add a persistent advertisement, such that libp2p manages the re-advertisement when the TTL expires.

core/node.go
```diff
@@ -143,9 +198,51 @@ func (n *Node) ID() peer.ID {
...
+
+	if n.config.UseBootstrapList {
```
raulk: You should probably move this code above the advertisement. If you're connected to nobody, we won't manage to advertise anywhere.

core/node.go
```diff
@@ -143,9 +198,51 @@ func (n *Node) ID() peer.ID {
...
+	// TODO(albrow): Replace default list with our own bootstrap peer list.
+	for _, addr := range dht.DefaultBootstrapPeers {
+		peerInfo, err := peerstore.InfoFromP2pAddr(addr)
+		if err != nil {
+			return err
+		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if err := n.host.Connect(connectCtx, *peerInfo); err != nil {
```
raulk: `Connect` will establish a connection and trigger the notifees on protocols like the DHT. Those protocols will in turn open streams to the connected peer to test whether the other party supports the protocol, and will perform any init logic in consequence (e.g. adding the peer to a routing table). Until that happens, you won't have a working DHT node. Since the notifees are triggered asynchronously, so you should add like a 2 second wait after all dials completed.

We are speccing out a local event bus (will tag you in that issue) that will allow us to be deterministic instead of probabilistic in the near future (you'll listen to an event instead of relying on a timing condition), but until that's a thing, this is probably the safest path.

core/node.go
```diff
@@ -143,9 +198,51 @@ func (n *Node) ID() peer.ID {
...
+			return err
+		}
+	}
+
 	return n.mainLoop()
 }
 
+func (n *Node) connectToBootstrapList() error {
+	log.WithField("bootstrapPeers", dht.DefaultBootstrapPeers).Info("connecting to bootstrap peers")
+	connectCtx, cancel := context.WithTimeout(n.ctx, defaultNetworkTimeout)
```
raulk: You probably want a lower timeout than 30 seconds for connections -- IIRC we tend to use 5 seconds.

blockwatch/watcher.go
fabioberger: Yeah totally, thanks for making that change. It is an expected condition.

core/notifee.go
```diff
@@ -0,0 +1,69 @@
...
+	p2pnet "github.com/libp2p/go-libp2p-net"
+	ma "github.com/multiformats/go-multiaddr"
+	log "github.com/sirupsen/logrus"
+)
+
+// localConnTag is the tag used with the Connection Manager to mark local
+// connections.
+const localConnTag = "local-connection"
+
+type notifee struct {
```
fabioberger: Can you add a short description defining what a `notifee` is?

core/node.go
albrow: Yes, the short TTL is just for testing purposes right now. We're generating a random peer ID each time you run *main.go* so with a longer TTL we would see a lot of attempts to connect to a peer that disconnected a long time ago.

core/node.go
albrow: @raulk The doc comment for `discovery.Advertise` states:

> Advertise is a utility function that persistently advertises a service through an Advertiser

"persistently" seems to imply that it will continuously advertise to connected peers. Is that not the case? Maybe we could update the docs to clarify?

core/node.go
albrow: Okay I'll add a 2 second wait after dialing all the bootstrap peers for now. Also wanted to point out that the chat-with-rendezvous example doesn't do this: https://github.com/libp2p/go-libp2p-examples/blob/master/chat-with-rendezvous/chat.go. Should we add a `time.Sleep` there as well?

core/node.go
albrow: 👍 In my experience, WebRTC connections can take longer, especially with third-party signaling. I'll reduce it for now and take another look once we plug in the WebRTC transport.

core/node.go
```diff
@@ -28,19 +33,37 @@ const (
 	maxReceiveBatch = 100
 	// maxShareBatch is the maximum number of messages to share at once.
 	maxShareBatch = 10
+	// peerCountLow is the target number of peers to connect to at any given time.
+	peerCountLow = 10
+	// peerCountHigh is the maximum number of peers to be connected to. If the
+	// number of connections exceeds this number, we will prune connections until
+	// we reach peerCountLow.
+	peerCountHigh = 12
```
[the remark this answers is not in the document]
albrow: @fabioberger the Connection Manager decides which connections to prune based on tags/scores. This is good because it lets us easily implement something very similar to the "neighborhood establishment" algorithm described in the Architecture Doc.

Here's how it works: You can use [the `TagPeer` method](https://godoc.org/github.com/libp2p/go-libp2p-connmgr#BasicConnMgr.TagPeer) to add `val` to a peer's total score. The `tag` is simply a unique identifier (e.g. "0x-mesh-weijie-sub"). You can also use [the `UntagPeer` method](https://godoc.org/github.com/libp2p/go-libp2p-connmgr#BasicConnMgr.UntagPeer) to remove the `val` associated with a particular tag from a peer's total score. When it is time to prune connections, the Connection Manager takes the sum of the `val`s from all tags for each peer and disconnects the peers with the lowest score.

> Or will we be able to prune based on neighbor reputation? 

Yes, once we implement WeijieSub and peer ranking, all we need to do is add tags for each peer according to the "peer ranking" algorithm. As long as we keep everything in sync, it will automatically prune the connections with the lowest score.

To help find the right group of peers more quickly, I'll probably also add a slight positive score when a peer indicates that they speak the 0x Mesh protocol (we can detect this via a `StreamHandler`). This means the Connection Manager will automatically disconnect from peers that don't speak the protocol (or peers that only speak an older version). Thanks to @raulk for this suggestion on Slack.

core/node.go
raulk: You’re right. I thought you were using Advertiser.Advertise.

albrow:
@fabioberger removed WIP tag. We will probably make improvements to peer discovery and peer ranking over time, but this PR is ready for final review.

albrow:
Fixes https://github.com/0xProject/0x-mesh/issues/72

core/notifee.go
```diff
@@ -0,0 +1,81 @@
...
+		"peerID":       conn.RemotePeer(),
+		"multiaddress": conn.RemoteMultiaddr(),
+	}).Trace("disconnected from peer")
+}
+
+// OpenedStream is called when a stream opened
+func (n *notifee) OpenedStream(network p2pnet.Network, stream p2pnet.Stream) {
+	go func() {
+		// HACK(albrow): When the stream is initially opened, the protocol is not set.
+		// For now, we have to manually poll until it is set.
```
fabioberger: Is there a PR or issue you could add a link to?

albrow: Just added a link to this issue: https://github.com/libp2p/go-libp2p/issues/467. I'm not 100% sure if it will help us avoid this hack; it depends on how exactly the event bus is implemented.

PR #70, opened 2019-05-23
ethereum/zeroex: Implement signing utilities

fabioberger:
This PR:
- Implements utility for signing a message via `eth_sign` and associated tests
- Add convenience method for signing a 0x order and returning the signature in the format expected by the smart contracts.

### DIFF
--- a/ethereum/sign.go
+++ b/ethereum/sign.go
@@ -1,0 +1,39 @@
+package ethereum
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+// ECSignature contains the parameters of an elliptic curve signature
+type ECSignature struct {
+	V byte
+	R []byte
+	S []byte
+}
+
+// ECSign signs an order and generates an EthSign signature type
+func ECSign(message []byte, signerAddress common.Address, rpcClient *rpc.Client) (*ECSignature, error) {
+	// Set an ETH_SIGN JSON-RPC request
+	var signatureHex string
+	if err := rpcClient.Call(&signatureHex, "eth_sign", signerAddress.Hex(), common.Bytes2Hex(message)); err != nil {
+		return nil, err
+	}
+	// eth_sign returns the signature as r+s+v
+	signatureBytes := common.Hex2Bytes(signatureHex[2:])
+	vParam := signatureBytes[64]
+	if vParam == byte(0) {
+		vParam = byte(27)
+	} else if vParam == byte(1) {
+		vParam = byte(28)
+	}
+	rParam := signatureBytes[0:32]
+	sParam := signatureBytes[32:64]
+
+	ecSignature := &ECSignature{
+		V: vParam,
+		R: rParam,
+		S: sParam,
+	}
+	return ecSignature, nil
+}
--- a/ethereum/sign.go
+++ b/ethereum/sign.go
@@ -8,8 +8,8 @@ import (
 // ECSignature contains the parameters of an elliptic curve signature
 type ECSignature struct {
 	V byte
-	R []byte
-	S []byte
+	R common.Hash
+	S common.Hash
 }
 
 // ECSign signs a message via the `eth_sign` Ethereum JSON-RPC call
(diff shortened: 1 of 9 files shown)


zeroex/order.go
```diff
@@ -43,6 +45,21 @@ type SignedOrder struct {
...
+// SignatureType values
+const (
+	IllegalSignature SignatureType = iota
+	InvalidSignature
+	EIP712Signature
+	EthSignSignature
+	WalletSignature
+	ValidatorSignature
+	PreSignedSignature
+	NSignatureTypesSignature
```
fabioberger: I had to suffix these values with `Signature` because there was already a collision between the `Invalid` value in this enum, and the `Invalid` value in the `OrderStatus` enum.

ethereum/sign_test.go
```diff
@@ -0,0 +1,31 @@
+// +build !js
```
albrow: Hmm.. What's the blocker for testing this in Node.js/Wasm? Does github.com/ethereum/go-ethereum/rpc not work out of the box?

fabioberger: It's just that the RPC call happens in a test file and Golang re-routes http requests in test files: https://github.com/golang/go/issues/31559

ethereum/sign.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+// ECSignature contains the parameters of an elliptic curve signature
+type ECSignature struct {
+	V byte
+	R []byte
```
albrow: If `V` and `R` are always a fixed length, we should use an array (e.g. `[32]byte`) instead of a slice here.

zeroex/order.go
```diff
@@ -154,6 +171,25 @@ func (s *SignedOrder) ComputeOrderHash() (common.Hash, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	ecSignature, err := ethereum.ECSign(orderHash.Bytes(), s.MakerAddress, rpcClient)
+	if err != nil {
+		return nil, err
+	}
+
+	signature := []byte{}
+	signature = append(signature, ecSignature.V)
```
albrow: Using an array instead of a slice will also make this code more efficient. Instead of making 4 calls to `append`, we can preallocate the length we need (66 bytes I think?). Something like this:

```go
signature := make([]byte, 66)
signature[0] = ecSignature.V
signature[1:33] = ecSignature.R
signature[34:66] = ecSignature.S
signature[66] = byte(EthSignSignature)
```

zeroex/order.go
```diff
@@ -154,6 +171,25 @@ func (s *SignedOrder) ComputeOrderHash() (common.Hash, error) {
 	return hash, nil
 }
 
+// ECSign signs the order with it's `makerAddress` and formats it to be a valid 0x order signature
+func (s *SignedOrder) ECSign(rpcClient *rpc.Client) ([]byte, error) {
```
albrow: Can we make a separate `Order` type to compliment the `SignedOrder` type? Then we could make `ECSign` a method of `Order` or a function that takes an `Order` as an argument. I think it would be more clear and would also allow us to catch certain bugs at compile time. The way it is currently written, passing a `SignedOrder` with a `nil` signature to functions which expect a signature can result in a runtime error.

ethereum/sign.go
fabioberger: Agreed!

zeroex/order.go
```diff
@@ -174,6 +215,19 @@ func (s *SignedOrder) ConvertToOrderWithoutExchangeAddress() wrappers.OrderWitho
 	return orderWithoutExchangeAddress
 }
 
+// ConvertToSignedOrder converts an order to a signed 0x order
+func ConvertToSignedOrder(order *Order, rpcClient *rpc.Client) (*SignedOrder, error) {
```
albrow: I don't like that there are two exported identifiers for signing an order which do relatively similar things. How about making `Order.ECSign` unexported and renaming `ConvertToSignedOrder` to `SignOrder`? Then there's only one exported function called `SignOrder` with a clear purpose.

PR #78, opened 2019-05-24
ethereum: Local eth_sign equivalent signing

fabioberger:
This PR:
- Adds a signing method that performs the identical signing logic as a call to `eth_sign` would, just locally using a private key.
- Refactors our tests to use the local signing method where sensible. Some tests (e.g., `meshdb_test.go` can now be run in WASM again 🍾 )


zeroex/order.go
```diff
@@ -215,7 +223,21 @@ func (s *SignedOrder) ConvertToOrderWithoutExchangeAddress() wrappers.OrderWitho
 	return orderWithoutExchangeAddress
 }
 
-// SignOrder converts an order to a signed 0x order
+// SignTestOrder converts a test 0x order to a signed 0x order
+func SignTestOrder(order *Order) (*SignedOrder, error) {
```
albrow: One reason I wanted to use a `Signer` interface is so that we could simplify the code for `SignOrder`. This allows us to abstract away the parts of the code that are the same (generating an order hash, signing it, adding the `Signature` field to the order) and the parts that are different (*how* exactly the order hash is signed).

What do you think of changing the API for `SignOrder` to:

```go
func SignOrder(signer ethereum.Signer, order Order) (*SignedOrder, error)
```

PR #80, opened 2019-05-24
Reorganize main and core packages

albrow:
There are two main motivations for this PR:

1. __main.go__ is becoming too big and is hard to work with.
2. __main.go__ mixes concerns by having some functionality that is browser-compatible (e.g. setting up the database and various watcher services) and some that is not (e.g. parsing environment variables and starting the RPC server).

Since we are re-organizing things, I took the time to rename/move some packages, which I have been wanting to do for a while.

Summary of major changes:

1. The `core` package has been renamed `p2p` to more accurately reflect its responsibilities. The package does not actually contain the core business logic for 0x Mesh. It is only concerned with peer discovery and sending/receiving messages.
2. Large parts of __main.go__ have been moved to a new `core` package. The new `core` package *does* contain the core business logic for 0x Mesh. It hooks everything together and is (or will be) compatible with both browsers and the standalone node.
3. The remaining parts of __main.go__ have been moved to the `cmd/mesh` package. This package is lightweight and only responsible for parsing environment variables and starting the RPC server.
4. The __cmd/__ directory is the new home for anything intended to be run from the command line, so __demo/add_order__ and __demo/add_peer__ have also been moved there.
5. I renamed the `ws` package to `rpc`. This package specifically implements a JSON RPC server+client over WebSockets, but the name `ws` implies it was WebSockets library. This also opens the door for us to support JSON RPC over HTTP or other protocols in the future.
6. There is a new Makefile target called "mesh". You can install the standalone version of 0x Mesh by running `make mesh`.


Makefile
```diff
@@ -21,3 +21,8 @@ test-wasm:
 .PHONY: lint
 lint:
 	golangci-lint run
+
+
+.PHONY: mesh
+mesh:
+	go install ./cmd/mesh
```
fabioberger: Have a slight preference for `cli` over `cmd`, it's a bit less vague imo.

Issue #74, opened 2019-05-23
Implement RPC method for subscribing to order updates


albrow:
Currently the RPC API can be used to add new orders and connect to peers, but we need to add support for subscribing to order updates. Clients should be notified when:

- An existing order becomes invalid (filled, canceled, or expired).
- A previously existing order becomes valid again (due to a block re-org)
- A new order is received from a peer.

albrow:
Implemented in #87.

PR #87, opened 2019-05-26
Add RPC method for subscribing to order events

fabioberger:
This PR: Completes: https://github.com/0xProject/0x-mesh/issues/74

One issue I ran into was that the interface for receiving order events from the OrderWatcher did not support having multiple subscribers so I modified the interface to allow it. @albrow I know we spoke about not using channels for interfaces but in this case, the `go-ethereum` `rpc` package forced my hand. I'm still down to attempt a refactor in the future, but this approach made sense for a first iteration.

Testing instructions:

In one terminal, start the Ganache snapshot for Mesh:

```
docker run -ti -p 8545:8545 0xorg/mesh-ganache-cli:5.0.0
```

In another terminal, start the Mesh node:

```
rm -rf ./0x_mesh_db/; make mesh; RPC_PORT=60556 ETHEREUM_NETWORK_ID="50" ETHEREUM_RPC_URL="http://localhost:8545" VERBOSITY=5  mesh
```

Finally, run the `add_order` client main.go:

```
cd cmd/demo/add_order
RPC_ADDRESS=ws://0.0.0.0:60556 ETHEREUM_RPC_URL=http://localhost:8545 go run main.go
```

In order to test multiple simultaneous subscriptions, replace the `add_order` `main.go` with: https://gist.github.com/fabioberger/bbd6fd03fb8020372b018bb0a534665d


rpc/service.go
```diff
@@ -18,6 +21,13 @@ type RPCHandler interface {
 	AddOrder(order *zeroex.SignedOrder) error
 	// AddPeer is called when the client sends an AddPeer request.
 	AddPeer(peerInfo peerstore.PeerInfo) error
+	// Orders is called when a client sends a Subscribe to orderStream request
+	Orders(ctx context.Context) (*rpc.Subscription, error)
+}
+
+// Orders calls rpcHandler.Orders and returns the rpc subscription.
```
albrow: The `go-ethereum/rpc` package has some very specific requirements about the type signature and name of methods which reduces the flexibility of our API a lot. The purpose of __rpc/service.go__ is to smooth out this API and expose names and signatures that make more sense in the context of the entire application. `rpcService` implements the API required by the `go-ethereum/rpc` package, but `RPCHandler` doesn't have to and is not necessarily bound by the same API restrictions.

At the very least, I think we should rename `RPCHandler.Orders` to something like `RPCHandler.SubscribeToOrders`. The name `Orders` is too generic and doesn't make sense in the context of the broader application.

We should also consider simplifying the API. For example, if we update the `core` package so that it is not RPC-aware, we could also modify `RPCHandler` to be more aligned with the `core` package API to reduce the required glue code (e.g. `RPCHandler` could have a method that accepts a `chan []*zeroex.OrderInfo` and `rpcService` can wrap this up into an `rpc.Subscription`). I'm okay with punting on changing the API until after we make progress on browser support, because that will affect what the common API in the `core` package looks like.

zeroex/orderwatch/watcher.go
```diff
@@ -184,13 +185,12 @@ func (w *Watcher) setupInMemoryOrderState(signedOrder *zeroex.SignedOrder, order
...
-	w.cachedOrderEventsMux.Lock()
-	cachedEvents := w.cachedOrderEvents
-	w.cachedOrderEvents = []*zeroex.OrderInfo{}
-	w.cachedOrderEventsMux.Lock()
-	return cachedEvents
+// Subscribe allows one to subscribe to the order events emitted by the OrderWatcher.
+// To unsubscribe, simply call `Unsubscribe` on the returned subscription.
+// The sink channel should have ample buffer space to avoid blocking other subscribers.
+// Slow subscribers are not dropped.
+func (w *Watcher) Subscribe(sink chan<- []*zeroex.OrderInfo) event.Subscription {
```
albrow: > I know we spoke about not using channels for interfaces but in this case, the go-ethereum rpc package forced my hand. I'm still down to attempt a refactor in the future, but this approach made sense for a first iteration.

I agree with you and I think it is okay for this iteration to use channels in the exported API. However, I wanted to point out that just because the `go-ethereum/rpc` package requires the use of channels, doesn't mean that we have to use channels in the other parts of our API. We could use callbacks in the `orderwatch` package and then convert to channels in the `rpc` package.

I'm going to create an issue for re-examining our asynchronous APIs to make things more consistent. I'm okay with punting for now.

fabioberger: 🙏

core/core.go
```diff
@@ -233,6 +232,39 @@ func (app *App) AddPeer(peerInfo peerstore.PeerInfo) error {
 	return app.node.Connect(ctx, peerInfo)
 }
 
+// SetupOrderStream sets up the order stream for a subscription
+func (app *App) SetupOrderStream(ctx context.Context) (*rpc.Subscription, error) {
```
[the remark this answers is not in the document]
fabioberger: I added a `SubscribeToOrderEvents` method to `core` that is _not_ RPC-related, and moved the RPC-specific `SetupOrderStream` to `cmd/mesh`.

Issue #88, opened 2019-05-26
rpc: Allow submitting multiple 0x orders in one JSON-RPC request


fabioberger:
Currently, the WS JSON-RPC interface requires that 0x orders be submitting one at a time. This is very inefficient since each request is calling `BatchValidate()` under-the-hood with a single order. If one could submit multiple orders in a single request, they could also be validated in fewer batches.

albrow:
Implemented in #89.

PR #89, opened 2019-05-27
rpc: Refactor `AddOrder` to `AddOrders`

fabioberger:
Completes: https://github.com/0xProject/0x-mesh/issues/88

Since the caller now submits multiple orders in a single call, this method's return value is of type:

```
type AddOrdersResponse struct {
	Added       []*zeroex.SuccinctOrderInfo
	Invalid     []*zeroex.SuccinctOrderInfo
	FailedToAdd []common.Hash
}
```

Why? 

- `Added`: Some orders pass the order validation test and are added to Mesh.
- `Invalid`: Some order are not added to Mesh because they fail the order validation test. By letting the client know, they can also drop those orders from their data store.
- `Failed`: Some orders are not added to Mesh if the network request to validate them fails unexpectedly. By letting the client know, they can then try and add those orders again in a new request.

### DIFF
--- a/cmd/demo/add_order/main.go
+++ b/cmd/demo/add_order/main.go
@@ -61,10 +61,10 @@ func main() {
 		log.WithError(err).Fatal("could not sign 0x order")
 	}
 
-	hash, err := client.AddOrder(signedTestOrder)
+	orderHashToSuccinctOrderInfo, err := client.AddOrders([]*zeroex.SignedOrder{signedTestOrder})
 	if err != nil {
 		log.WithError(err).Fatal("error from AddOrder")
 	} else {
-		log.Printf("added valid order: %s", hash.Hex())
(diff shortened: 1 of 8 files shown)


albrow:
Shouldn't this new endpoint and [`OrderValidator.BatchValidate`](https://github.com/0xProject/0x-mesh/blob/ac9a8dda7d97baf972b391e9f50c02bd8930c3af/zeroex/order_validator.go#L71) have the same/similar API? Can we change response type for the RPC endpoint to:

```
map[common.Hash]zeroex.SuccinctOrderInfo
```

Alternatively, if you think that `Added`, `Invalid`, `FailedToAdd` structure is an improvement, can we change the `OrderValidator.BatchValidate` API to match?

Staying consistent will make it easier to write and maintain the code and make it easier for callers to understand.

Issue #85, opened 2019-05-25
core: Add extra validation to `[]byte` fields on incoming SignedOrders received from peers


fabioberger:
Currently we simply check if an incoming message from a peer can be unmarshalled into an `SignedOrder` struct. In order to prevent the `batchValidate()` method from panicking when calling the `getOrdersAndTradersInfo()` contract method, we need to perform some additional validation on a `SignedOrder`'s `[]byte` fields.

1. Make sure the byte array representing `assetData`'s are decodable and valid.
2. Make sure the signature byte array has an even number of bytes.

If either of these conditions fail, the message should be immediately discarded and not forwarded on to the `batchValidate()` call.

fabioberger:
Closed by #92

PR #92, opened 2019-05-27
zeroex: Perform all possible off-chain 0x validation before on-chain validation

fabioberger:
This PR completes: https://github.com/0xProject/0x-mesh/issues/85 and https://github.com/0xProject/0x-mesh/issues/57https://github.com/0xProject/0x-mesh/issues/57

- Previously we were relying entirely on on-chain 0x order validation. Not only was this inefficient, the very call to the contract would fail if the supplied orders with incorrectly formatted. This PR thus adds the following off-chain validation checks:
  - Make sure maker & taker assetData are decodable and supported
  - Make sure signature is properly formatted and supported
  - Make sure order isn't expired
  - Make sure maker & taker assetAmounts are not 0
If any of these checks fail, `BatchFill()` won't include these orders in the contract call. An `OrderInfo` object for these invalid orders is still returned from `BatchFill()` however.

### DIFF
--- a/zeroex/order_validator.go
+++ b/zeroex/order_validator.go
@@ -233,8 +233,6 @@ func (o *OrderValidator) BatchOffchainValidation(signedOrders []*SignedOrder) (m
 		if err != nil {
 			log.Panic("Computing the orderHash failed unexpectedly")
 		}
-		// TODO(fabio): Allow Mesh operator to specify an expirationBuffer (+/-)
-		// that they wish to tolerate. Use the same expirationBuffer as the ExpirationWatcher
 		now := big.NewInt(time.Now().UTC().Unix())
 		if signedOrder.ExpirationTimeSeconds.Cmp(now) == -1 {
 			orderHashToInfo[orderHash] = &OrderInfo{
--- a/zeroex/order_validator.go
+++ b/zeroex/order_validator.go
@@ -233,7 +233,7 @@ func (o *OrderValidator) BatchOffchainValidation(signedOrders []*SignedOrder) (m
 		if err != nil {
 			log.Panic("Computing the orderHash failed unexpectedly")
 		}
-		now := big.NewInt(time.Now().UTC().Unix())
+		now := big.NewInt(time.Now().Unix())
 		if signedOrder.ExpirationTimeSeconds.Cmp(now) == -1 {
 			orderHashToInfo[orderHash] = &OrderInfo{
 				OrderHash:                orderHash,
--- a/zeroex/order_validator.go
+++ b/zeroex/order_validator.go
@@ -244,7 +244,7 @@ func (o *OrderValidator) BatchOffchainValidation(signedOrders []*SignedOrder) (m
 			continue
 		}
 
-		if signedOrder.MakerAssetAmount == big.NewInt(0) {
+		if signedOrder.MakerAssetAmount.Cmp(big.NewInt(0)) == 0 {
 			orderHashToInfo[orderHash] = &OrderInfo{
 				OrderHash:                orderHash,
 				SignedOrder:              signedOrder,
@@ -253,7 +253,7 @@ func (o *OrderValidator) BatchOffchainValidation(signedOrders []*SignedOrder) (m
 			}
 			continue
 		}
-		if signedOrder.TakerAssetAmount == big.NewInt(0) {
+		if signedOrder.TakerAssetAmount.Cmp(big.NewInt(0)) == 0 {
 			orderHashToInfo[orderHash] = &OrderInfo{
 				OrderHash:                orderHash,
 				SignedOrder:              signedOrder,
--- a/zeroex/order_validator.go
+++ b/zeroex/order_validator.go
@@ -60,7 +60,8 @@ type OrderHashToSuccinctOrderInfo map[common.Hash]*SuccinctOrderInfo
 
 // OrderValidator validates 0x orders
 type OrderValidator struct {
-	orderValidator *wrappers.OrderValidator
+	orderValidator   *wrappers.OrderValidator
+	assetDataDecoder *AssetDataDecoder
 }
 
 // NewOrderValidator instantiates a new order validator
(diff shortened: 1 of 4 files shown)


zeroex/order_validator.go
```diff
@@ -197,6 +207,155 @@ func (o *OrderValidator) BatchValidate(signedOrders []*SignedOrder) map[common.H
...
 }
 
+// BatchOffchainValidation performs all off-chain validation checks on a batch of 0x orders.
+// These checks include:
+// - `MakerAssetAmount` and `TakerAssetAmount` cannot be 0
+// - `AssetData` fields contain properly encoded, and currently supported assetData (ERC20 & ERC721 for now)
+// - `Signature` contains a properly encoded 0x signature
+// - Validate that order isn't expired
+// Returns an orderHashToInfo mapping with all invalid orders added to it, and an array of the valid signedOrders
+func (o *OrderValidator) BatchOffchainValidation(signedOrders []*SignedOrder) (map[common.Hash]*OrderInfo, []*SignedOrder) {
```
albrow: When we change the API for `BatchValidate`, we should take another look at this method as well. I'm okay with leaving it as is for now.

zeroex/order_validator_test.go
```diff
@@ -17,68 +18,150 @@ import (
...
+}
 
-	orderHash, err := order.ComputeOrderHash()
-	require.NoError(t, err)
+func TestBatchValidateMalformedSignature(t *testing.T) {
+	signedOrder := &SignedOrder{
+		Order: &testOrder,
+	}
+	// Incorrectly formatted signature
+	signedOrder.Signature = malformedSignature
```
albrow: Is there a reason the malformed signature case can't be covered by a test case in `TestBatchValidateOffChainCases`?

zeroex/order_validator_test.go
```diff
@@ -240,3 +310,21 @@ func TestCalculateRemainingFillableTakerAmount(t *testing.T) {
 	remainingFillableTakerAssetAmount := calculateRemainingFillableTakerAmount(signedOrder, orderInfo, traderInfo)
 	assert.Equal(t, new(big.Int).Div(takerAssetAmount, big.NewInt(2)), remainingFillableTakerAssetAmount)
 }
+
+func modifyOrder(t *testing.T, order Order, field string, value interface{}) Order {
```
albrow: Using a `string` to identify struct fields and an `interface{}` for values makes me nervous. I would rather have more lines of type-safe code than fewer lines of type-unsafe code. How do you feel about changing/removing this?

We could still use a type-safe `copyOrder` function so we don't have to create a new order from scratch every time: 

```go
// works because function arguments are passed as a copy and the Order type has no pointer fields.
func copyOrder(order Order) Order {
   return order
}
```

zeroex/orderwatch/expirationwatch.go
```diff
@@ -134,7 +134,7 @@ func (e *ExpirationWatcher) Receive() <-chan []ExpiredOrder {
 // to the caller
 func (e *ExpirationWatcher) prune() []ExpiredOrder {
 	pruned := []ExpiredOrder{}
-	currentTimestamp := time.Now().Unix()
+	currentTimestamp := time.Now().UTC().Unix()
```
albrow: [`Time.Unix`](https://golang.org/pkg/time/#Time.Unix) already converts to UTC so you don't have to do this.

> Unix returns t as a Unix time, the number of seconds elapsed since January 1, 1970 UTC. The result does not depend on the location associated with t.

zeroex/order_validator_test.go
fabioberger: It was because all the test cases were for fields on `Orders`, and this changes the `SignedOrder`. I've refactored the test cases to use SignedOrders.

zeroex/order_validator_test.go
```diff
@@ -22,88 +22,90 @@ var malformedAssetData = []byte("9HJhsAAAAAAAAAAAAAAAAInSSmtMyxtvqiYl")
...
-	TakerAssetData:        common.Hex2Bytes("f47261b00000000000000000000000000b1ba0af832d7c05fd64161e0db78e85978e8082"),
-	Salt:                  big.NewInt(1548619145450),
-	MakerFee:              big.NewInt(0),
-	TakerFee:              big.NewInt(0),
-	MakerAssetAmount:      big.NewInt(1000),
-	TakerAssetAmount:      big.NewInt(2000),
-	ExpirationTimeSeconds: big.NewInt(time.Now().Add(48 * time.Hour).Unix()),
-	ExchangeAddress:       constants.NetworkIDToContractAddresses[constants.TestNetworkID].Exchange,
+var testSignedOrder = SignedOrder{
+	Order: &Order {
```
albrow: I'm sorry I'm just now noticing this. The embedded `Order` in `SignedOrder` should probably not be a pointer. We already pass around `SignedOrder` by reference pretty much everywhere so the additional indirection is not necessary and just introduces the possibility of bugs where `Order` could be nil.

Don't need to address now. I'll create a separate issue for it.

zeroex/order_validator_test.go
```diff
@@ -291,23 +265,52 @@ func TestCalculateRemainingFillableTakerAmount(t *testing.T) {
 	assert.Equal(t, new(big.Int).Div(takerAssetAmount, big.NewInt(2)), remainingFillableTakerAssetAmount)
 }
 
-func orderWithCustomMakerAssetAmount(order Order, makerAssetAmount *big.Int) Order {
-	order.MakerAssetAmount = makerAssetAmount
-	return order
+func signedOrderWithCustomMakerAssetAmount(t *testing.T, signedOrder SignedOrder, makerAssetAmount *big.Int) SignedOrder {
+	signedOrderCopy := copySignedOrder(signedOrder)
```
albrow: If we change `SignedOrder.Order` so that it's not a pointer, it would also simplify this code a lot. The `Order` would also be copied so you could remove the `copyOrder` and `copySignedOrder` functions.

PR #98, opened 2019-05-28
Improve db errors

### LINKED ISSUE #28: db: Improve error messages.
We should probably have typed errors for common cases like `NotFoundError` and `TypeError`.

albrow:
Fixes #28.

I only added a few typed errors for now. It is probably not necessary to create a new error type for every possible error that might be returned, but if we need more in the future we can always add them.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -7,9 +7,8 @@ import (
 	"context"
 	"time"
 
-	"github.com/0xProject/0x-mesh/db"
-
 	"github.com/0xProject/0x-mesh/blockwatch"
+	"github.com/0xProject/0x-mesh/db"
 	"github.com/0xProject/0x-mesh/ethereum"
 	"github.com/0xProject/0x-mesh/meshdb"
 	"github.com/0xProject/0x-mesh/p2p"
--- a/core/core.go
+++ b/core/core.go
@@ -5,8 +5,9 @@ package core
 
 import (
 	"context"
-	"strings"
 	"time"
+
+	"github.com/0xProject/0x-mesh/db"
 
 	"github.com/0xProject/0x-mesh/blockwatch"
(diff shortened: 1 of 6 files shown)


core/core.go
```diff
@@ -179,14 +179,14 @@ func (app *App) Start() error {
 	return nil
 }
 
-// TODO(albrow): Either use the Exists method or check for a typed error after
-// updating the db package.
+// TODO(albrow): Uset the more efficient Exists method instead of FindByID.
```
```suggestion
// TODO(albrow): Use the more efficient Exists method instead of FindByID.
```

db/collection.go
```diff
@@ -170,21 +169,17 @@ func (c *Collection) Update(model Model) error {
...
 	if err := c.checkModelType(model); err != nil {
 		return err
 	}
-	data, err := c.db.ldb.Get(key, nil)
+	pk := c.primaryKeyForID(id)
+	data, err := c.db.ldb.Get(pk, nil)
 	if err != nil {
+		if err == leveldb.ErrNotFound {
+			return NotFoundError{ID: id}
+		}
```
fabioberger: Nice! 🚀

db/collection.go
```diff
@@ -170,21 +169,17 @@ func (c *Collection) Update(model Model) error {
...
 		return err
 	}
-	if data == nil {
-		return errors.New("model not found")
-	}
```
fabioberger: Why is this conditional no longer needed?

albrow: It was probably never needed. If the key isn't found in the database it returns a `leveldb.ErrNotFound` instead of `nil`.

If you are using the API exposed in the db package, data cannot be nil. Technically, someone could manually add a key with a nil value, but there are so many other things you could break by manually setting keys that I think it would be too hard to consider.

PR #99, opened 2019-05-29
Implement a max message size for orders

albrow:
We've always wanted to perform ETH balance checks to limit the capacity for malicious actors to flood the network. The original plan was to calculate `(ETH balance of maker) / (total bytes of storage allocated to maker)`. However, we've discovered that it is almost impossible to determine the number of bytes of storage used by a given maker due to the fact that LevelDB uses opaque compression techniques. Even if we could do it, it would likely be computationally expensive and likely require iterating over a large number of keys.

As a compromise, we've chosen to use a less precise calculation: `(ETH balance of maker) / (total number of orders stored for that maker)`. Since orders can be theoretically unlimited in size, we've also decided to also add a max size for order messages. This PR implements that limit.

Still WIP while we figure out some interface design questions. I also want to update peer scores in the Connection Manager in this PR before we merge.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -7,14 +7,12 @@ import (
 	"context"
 	"time"
 
-	"github.com/0xProject/0x-mesh/db"
 	"github.com/0xProject/0x-mesh/ethereum"
 	"github.com/0xProject/0x-mesh/ethereum/blockwatch"
 	"github.com/0xProject/0x-mesh/meshdb"
 	"github.com/0xProject/0x-mesh/p2p"
 	"github.com/0xProject/0x-mesh/zeroex"
 	"github.com/0xProject/0x-mesh/zeroex/orderwatch"
-	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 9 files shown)


albrow:
Still leaving this WIP. Functionally everything is there, but I'm not happy with the current structure of the code and want to see if we can clean it up before merging.

albrow:
@fabioberger removed WIP tag. This PR is ready for review.

It does two main things:

1. Implements a peer scoring system. The `p2p` package exposes some methods for updating peer scores and the `core` package uses them to update scores when certain events occur. Right now this is only used by the Connection Manager to periodically disconnect from peers with the lowest scores. In the future this same scoring mechanism could be used as part of WeijieSub.
2. Validates the size of messages received from peers and orders received via RPC. When a peer sends us a message that is too big, we give them a negative score (same for other cases where we received an invalid message/order).

In retrospect, these two things could have probably been implemented in two separate PRs, but it was hard for me to see how all the pieces fit together without working on both.

Issue #94, opened 2019-05-28
zeroex: Add Ethereum txHash to Order events emitted


fabioberger:
Some of the order events emitted by Mesh were triggered by the EventWatcher, and in those cases, the state change (e.g., filled -> cancelled) can be directly linked to a specific Ethereum transaction. Both to allow consumers to verify the validity of a specific order event, and for them to pull additional information about the transaction that caused that state change, we should return the corresponding Ethereum transaction when available.

New OrderInfo struct:
```
type OrderInfo struct {
    OrderHash                          common.Hash
    SignedOrder                       *SignedOrder
    FillableTakerAssetAmount  *big.Int
    OrderStatus                        OrderStatus
    TxHash                                common.Hash
}
```

Since `common.Hash` is an alias for `[32]byte`, if there is no corresponding `TxHash` for an event (e.g., new order discovered in P2P network), the zero-value is set.

fabioberger:
Closed by #101

PR #101, opened 2019-05-29
orderwatch: Add TxHash to Order events emitted by OrderWatcher 

fabioberger:
Addresses: https://github.com/0xProject/0x-mesh/issues/94

### DIFF
--- a/zeroex/order_validator.go
+++ b/zeroex/order_validator.go
@@ -43,7 +43,8 @@ type OrderInfo struct {
 	SignedOrder              *SignedOrder
 	FillableTakerAssetAmount *big.Int
 	OrderStatus              OrderStatus
-	TxHash                   common.Hash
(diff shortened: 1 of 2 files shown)


zeroex/order_validator.go
```diff
@@ -37,12 +37,13 @@ type getOrdersAndTradersInfoParams struct {
...
 }
 
-// OrderInfo represents the order information returned from OrderValidator methods.
+// OrderInfo represents the order information emitted from Mesh
 type OrderInfo struct {
 	OrderHash                common.Hash
 	SignedOrder              *SignedOrder
 	FillableTakerAssetAmount *big.Int
 	OrderStatus              OrderStatus
+	TxHash                   common.Hash
```
albrow: Can we document what `TxHash` means here? It's not completely obvious, but I think it is something like "the hash of the transaction that caused the order status to change"?

PR #103, opened 2019-05-29
zeroex: Make `GetOrdersAndTradersInfo()` network request failure logs less noisy

fabioberger:
This PR edits the logs emitted on `GetOrdersAndTradersInfo()` request failures since it was too noisy and of limited use. Most of the time the failure has nothing to do with the orders submitted and more to do with other network disruptions.

### DIFF
--- a/zeroex/order_validator.go
+++ b/zeroex/order_validator.go
@@ -147,7 +147,7 @@ func (o *OrderValidator) BatchValidate(signedOrders []*SignedOrder) map[common.H
 				results, err := o.orderValidator.GetOrdersAndTradersInfo(opts, params.Orders, params.Signatures, params.TakerAddresses)
 				if err != nil {
 					log.WithFields(log.Fields{
-						"err":       err.Error(),
+						"error":     err.Error(),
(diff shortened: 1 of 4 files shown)


zeroex/order_validator.go
```diff
@@ -147,18 +147,16 @@ func (o *OrderValidator) BatchValidate(signedOrders []*SignedOrder) map[common.H
 				results, err := o.orderValidator.GetOrdersAndTradersInfo(opts, params.Orders, params.Signatures, params.TakerAddresses)
 				if err != nil {
 					log.WithFields(log.Fields{
-						"err":            err.Error(),
-						"attempt":        b.Attempt(),
-						"orders":         params.Orders,
-						"takerAddresses": params.TakerAddresses,
+						"err":       err.Error(),
```
albrow: This key should be `error` instead of `err`. It's going to be really important when we set up log monitoring.

fabioberger: Found this same issue in a bunch of places, changed all of them.

albrow: 🙏

PR #104, opened 2019-05-29
orderwatch: Avoid attempting to remove orders from ExpirationWatcher multiple times

fabioberger:
When an expired order is discovered by the ExpirationWatcher it is immediately removed from it in order to continue validating other orders. The `unwatchOrder()` method in OrderWatcher was blindly telling the ExpirationWatcher to remove all orders, even those that should be unwatched because they have expired. This PR fixes the issue by discerning between `unwatch` requests because of expiry or other.

### DIFF
--- a/zeroex/orderwatch/watcher.go
+++ b/zeroex/orderwatch/watcher.go
@@ -467,7 +467,7 @@ func (w *Watcher) generateOrderEventsIfChanged(orders []*meshdb.Order) {
 		if order.FillableTakerAssetAmount.Cmp(orderInfo.FillableTakerAssetAmount) != 0 {
 			isOrderUnfillable := orderInfo.FillableTakerAssetAmount.Cmp(big.NewInt(0)) == 0
 			if isOrderUnfillable {
-				didExpire := true
+				didExpire := false
 				w.unwatchOrder(order, didExpire)
 			} else {
 				w.rewatchOrder(order, orderInfo)
--- a/zeroex/orderwatch/watcher.go
+++ b/zeroex/orderwatch/watcher.go
@@ -506,9 +506,8 @@ func (w *Watcher) unwatchOrder(order *meshdb.Order, didExpire bool) {
 		}).Error("Failed to update order")
 	}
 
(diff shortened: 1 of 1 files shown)


zeroex/orderwatch/watcher.go
```diff
@@ -492,7 +494,7 @@ func (w *Watcher) rewatchOrder(order *meshdb.Order, orderInfo *zeroex.OrderInfo)
 	w.expirationWatcher.Add(order.SignedOrder.ExpirationTimeSeconds.Int64(), order.Hash)
 }
 
-func (w *Watcher) unwatchOrder(order *meshdb.Order) {
+func (w *Watcher) unwatchOrder(order *meshdb.Order, didExpire bool) {
```
albrow: We should be careful not to fall into the [boolean flag trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap). I'm okay with it for now because it is fairly isolated in scope.

zeroex/orderwatch/watcher.go
```diff
@@ -504,7 +506,12 @@ func (w *Watcher) unwatchOrder(order *meshdb.Order) {
 		}).Error("Failed to update order")
 	}
 
-	w.expirationWatcher.Remove(order.SignedOrder.ExpirationTimeSeconds.Int64(), order.Hash)
+	// If the order is being unwatched because it has expired, it will have been removed
+	// from the ExpirationWatcher immediately so that it can continue checking the remaining
+	// orders immediately
+	if !didExpire {
```
albrow: "If the order is being unwatched because it has expired" is followed by `if !didExpire`. The comment is describing the opposite condition of the if statement. If this is intentional and not a bug, maybe we can reword the comment? Something like:

```
// If we unwatched the order but it didn't expire, we need to remove it from the expiration watcher. If
// it did expire, it will have already been removed so we don't need to remove it again.
```

fabioberger: Not a bug, just badly worded. I like your wording, thanks for that! 🙏

Issue #84, opened 2019-05-25
Use new Query.Count method to optimize GetMessagesToShare


albrow:
See https://github.com/0xProject/0x-mesh/blob/9c12e0bed28814e1b675f489751a493988b18654/main.go#L197

Once #83 is merged, we should now be able to address this TODO and optimize the query.

PR #107, opened 2019-05-29
Implement Query.Offset method

albrow:
`Query.Offset` is an analog to the [`OFFSET` keyword in SQL](https://www.postgresql.org/docs/current/queries-limit.html). (Just as `Query.Max` is the analog to `LIMIT` in SQL).

This feature is important for addressing https://github.com/0xProject/0x-mesh/issues/84 and will likely be useful for future optimizations as well.

### DIFF
--- a/db/query.go
+++ b/db/query.go
@@ -15,6 +15,7 @@ type Query struct {
 	col     *Collection
 	filter  *Filter
 	max     int
+	offset  int
 	reverse bool
 }
 
@@ -29,7 +30,9 @@ type Filter struct {
 // a query will return all models that match the filter in ascending byte order
 // according to their index values. The query offers methods that can be used to
 // change this (e.g. Reverse and Max). The query is lazily executed, i.e. it
-// does not actually touch the database until you call Run.
+// does not actually touch the database until they are run. In general, queries
+// have a runtime of O(N) where N is the number of models that match the query,
+// but using some features may significantly change this.
 func (c *Collection) NewQuery(filter *Filter) *Query {
 	return &Query{
 		col:    c,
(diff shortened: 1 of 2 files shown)


db/query.go
```diff
@@ -50,6 +53,15 @@ func (q *Query) Reverse() *Query {
 	return q
 }
 
+// Offset causes the query to skip offset models when iterating through models
+// that match the query. Note that queries which use an offset have a runtime of
+// O(offset + N), where N is the number of models that match the query. This is
```
fabioberger: Shouldn't it be `O(N - offset)`? Since you don't process the offset records?

albrow: It depends on what you mean by "process". The key thing that I am trying to communicate here is that we have to actually iterate over `offset` keys before we do anything else. You can't just skip over `offset` keys for free. Imagine you have a linked list and need to find the nth item in the list. That's sort of what we need to do here.

So if `offset` is high, then that will be the dominating factor in the runtime. You could have a query that returns nothing but if `offset` is exceedingly high it will still take a long time to run. On the other hand, if `offset` is low and `N` is high, then `N` will be the dominating factor. 
One thing I technically didn't account for in this notation is the fact that `offset` won't have any additional affect on runtime if it is higher than the total number of keys in the index. But it might be more confusing to try to express that.

I suppose I should also clarify what I mean by "the number of models that match the query". I intended for this number to account for other parts of the query like `Max`. So if you set `Max` to 5, the number of models that match the query cannot exceed 5. I could change the wording a bit to make this more clear.

How do you feel about this explanation instead?

```
// Note that queries which use an offset have a runtime of O(max(K, offset) + N), where N is
// the number of models returned by the query and K is the total number of keys in the index.
// Queries with a high offset can take a long time to run, regardless of the number of
// models returned. This is due to limitations of the underlying database.
```

albrow: We might also describe this as `O(offset + K + N)` since typically in big-O notation `O(max(a, b)) == O(a + b)` but I also think that might be confusing.

fabioberger: Thanks for elaborating. That makes a lot more sense now. I prefer your new proposed comment. Let's also add a line about how this implements the same behavior as SQL's OFFSET, and link to it (you put it in the PR description, but think having this in the code is better for posterity's sake).

PR #110, opened 2019-05-30
rpc/core/orderwatcher/zeroex: Refactor BatchValidate() & AddOrders() interfaces

fabioberger:
After discussions with Alex about the current interface to `BatchValidate` and in anticipation of the needs for further Mesh order validation to communicate different kinds of order rejections to clients calling `AddOrders`, we've decided to perform this interface refactor.

This PR:
- Refactors `BatchValidate` to return a `ValidationResults` struct.
- Change the code that calls and processes the return value of `BatchValidate`
- Introduced the `OrderEvent` type, which is emitted from the OrderWatcher
- Introduce two more enums to complement `OrderStatus`: `RejectedOrderCode` and `OrderEventKind`
- Removed the `didExpire` conditional in OW since it ultimately didn't make sense. The race-condition is attempts to prevent is not preventable with the current async design of our code.


zeroex/order_validator.go
```diff
@@ -47,16 +41,105 @@ type OrderInfo struct {
...
+	case ROUnfunded:
+		return EKOrderBecameUnfunded, true
+	default:
+		// Catch-all returns Invalid OrderEventKind
+		return EKInvalid, false
+	}
+}
+
+// GetMessageForRejectOrderCode returns the corresponding message for the RejectedOrderCode
+func GetMessageForRejectOrderCode(rejectedOrderCode RejectedOrderCode) string {
```
albrow: Maybe instead of using this helper function we could add the message itself to the `RejectedOrderCode` type?

Something like:

```go
type RejectedOrderCode struct {
    Code string
    Message string
}

var (
    RORequestFailed RejectedOrderCode = {
        Code: "RequestFailed"
        Message: "network request to Ethereum RPC endpoint failed"
    },
    // ...
)
```

This has two benefits: (1) It puts the definition of the code and its message closer together and (2) it makes it nearly impossible to forget to associate a code with a message.

Maybe if we go with this approach we also rename the `RejectedOrderCode` type? Something like `RejectedOrderStatus`?

zeroex/order_validator.go
```diff
@@ -225,80 +337,92 @@ func (o *OrderValidator) BatchValidate(rawSignedOrders []*SignedOrder) map[commo
 // - `Signature` contains a properly encoded 0x signature
 // - Validate that order isn't expired
 // Returns an orderHashToInfo mapping with all invalid orders added to it, and an array of the valid signedOrders
-func (o *OrderValidator) BatchOffchainValidation(signedOrders []*SignedOrder) (map[common.Hash]*OrderInfo, []*SignedOrder) {
-	orderHashToInfo := map[common.Hash]*OrderInfo{}
-	validSignedOrders := []*SignedOrder{}
+func (o *OrderValidator) BatchOffchainValidation(signedOrders []*SignedOrder) ([]*RejectedOrderInfo, []*SignedOrder) {
```
albrow: optional: I think it would be more natural to re-arrange the return values so we return the valid orders first and the invalid ones second. This maps more nicely to the familiar `(result, error)` return values in Go.

zeroex/order_validator.go
```diff
@@ -47,16 +41,105 @@ type OrderInfo struct {
...
+type RejectedOrderKind uint8
+
+// RejectedOrderKind values
+const (
+	ZeroExValidation RejectedOrderKind = iota
+	MeshError
+)
+
+// ValidationResults defines the validation results returned from BatchValidate
+type ValidationResults struct {
```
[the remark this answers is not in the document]
fabioberger: Added to this comment.

rpc/client.go
```diff
@@ -30,17 +30,17 @@ func NewClient(addr string) (*Client, error) {
 
 // AddOrders adds orders to the 0x Mesh node and broadcasts them throughout the
 // 0x Mesh network.
-func (c *Client) AddOrders(orders []*zeroex.SignedOrder) (*AddOrdersResponse, error) {
-	var addOrdersResponseJSON string
-	if err := c.rpcClient.Call(&addOrdersResponseJSON, "mesh_addOrders", orders); err != nil {
+func (c *Client) AddOrders(orders []*zeroex.SignedOrder) (*zeroex.ValidationResults, error) {
+	var validationResultsJSON string
+	if err := c.rpcClient.Call(&validationResultsJSON, "mesh_addOrders", orders); err != nil {
```
albrow: The `go-ethereum/rpc` package already does automatic marshaling/unmarshaling. I think you can just write it this way?

```go
var validationResults zeroex.ValidationResults
if err := c.rpcClient.Call(&validationResults, "mesh_addOrders", orders); err != nil {
    // ...
}
```

zeroex/order_validator.go
```diff
@@ -47,16 +41,110 @@ type OrderInfo struct {
...
-type OrderHashToSuccinctOrderInfo map[common.Hash]*SuccinctOrderInfo
+// RejectedOrderStatus enumerates all the unique reasons for an orders rejection
+type RejectedOrderStatus struct {
+	Code    string
+	Message string
+}
+
+// RejectedOrderStatus values
+var (
+	RORequestFailed RejectedOrderStatus = RejectedOrderStatus{
```
albrow: Go can infer the type here so you can leave it out to make this cleaner:

```go
var (
	RORequestFailed = RejectedOrderStatus{
		Code:    "NetworkRequestFailed",
		Message: "network request to Ethereum RPC endpoint failed",
	},
        // ...
)
```

PR #114, opened 2019-05-30
Use GossipSub instead of FloodSub

albrow:
This PR changes our our main mechanism for communicating orders to peers from FloodSub to GossipSub. GossipSub is still not perfect for our use case, but it will dramatically reduce thrashing that can occur in FloodSub whenever new messages are sent to all peers. It should lead to more efficient bandwidth usage overall.

The change itself was really easy, but because GossipSub incurs higher latency, I had to make significant changes to our tests in order to get them working again.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -52,7 +52,7 @@ const (
 	advertiseTTL = 5 * time.Minute
 	// pubsubProtocolID is the protocol ID to use for pubsub.
 	// TODO(albrow): Is there a way to use a custom protocol ID with GossipSub?
-	// pubsubProtocolID = protocol.ID("/0x-mesh-floodsub/0.0.1")
+	// pubsubProtocolID = protocol.ID("/0x-mesh-gossipsub/0.0.1")
 	pubsubProtocolID = pubsub.GossipSubID
 )
 
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -186,8 +186,6 @@ func New(config Config) (*Node, error) {
 	routingDiscovery := discovery.NewRoutingDiscovery(kadDHT)
 
 	// Set up pubsub.
-	pubsub.GossipSubHistoryGossip = 10
-	pubsub.GossipSubHistoryLength = 20
 	ps, err := pubsub.NewGossipSub(nodeCtx, basicHost)
 	if err != nil {
 		cancel()
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -23,7 +23,6 @@ import (
 	p2pnet "github.com/libp2p/go-libp2p-net"
 	peer "github.com/libp2p/go-libp2p-peer"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
-	protocol "github.com/libp2p/go-libp2p-protocol"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/multiformats/go-multiaddr"
 	ma "github.com/multiformats/go-multiaddr"
@@ -52,7 +51,9 @@ const (
 	// advertiseTTL is the TTL for our announcement to the discovery network.
 	advertiseTTL = 5 * time.Minute
 	// pubsubProtocolID is the protocol ID to use for pubsub.
-	pubsubProtocolID = protocol.ID("/0x-mesh-floodsub/0.0.1")
+	// TODO(albrow): Is there a way to use a custom protocol ID with GossipSub?
+	// pubsubProtocolID = protocol.ID("/0x-mesh-floodsub/0.0.1")
+	pubsubProtocolID = pubsub.GossipSubID
 )
 
 // bootstrapPeers is a list of peers to use for bootstrapping the DHT. Based on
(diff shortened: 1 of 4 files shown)


albrow:
To clarify, here are some issues with the way GossipSub fits in with 0x Mesh right now:


1. GossipSub maintains a message cache for the purposes of `IWANT` and `IHAVE` control messages. This cache is in-memory, so it might not be able to hold all orders that a peer knows about.
2. GossipSub is fundamentally about sharing new messages and has no mechanism for re-sharing old messages (after they have fallen out of the cache). This means we still need to use the `GetMessagesToShare` method and manually re-share old messages.
3.  The message cache implementation uses a combination of `seqno` and `from` to create a unique ID for each message. This means the same order might have a different message ID, which can cause the same order to be sent multiple times redundantly.
4. GossipSub currently does not allow us to specify our own protocol ID which means that we might waste time connecting to peers that speak GossipSub but don't know anything about 0x and won't contribute to the 0x Mesh network.

None of these are big enough problems to prevent 0x Mesh from working, but they do reduce the efficiency of the network as a whole. I think it will be good enough for the beta.

opaolini:
Hi,

I have started to implement `WeijieSub` router and the Neighbour Ranking algorithm for `0xMesh` based on your specification here: https://drive.google.com/file/d/1dAVTEND7e1sISO9VZSOou0DN-igoUi9z/view

I needed to fork https://github.com/libp2p/go-libp2p-pubsub as it currently does not allow to implement routers outside of the context of the `pubsub` package (issue ->  https://github.com/libp2p/go-libp2p-pubsub/issues/187) , but that may change "in the not so distant future" as the author said.

I should do a working "WIP" PR over the weekend, if you're interested of course :)

albrow:
@opaolini thank you for reaching out!

In general, we are extremely excited to see PRs from external contributors 😄However, WeijieSub might not be the best thing to work on right now. We are currently thinking that WeijieSub will not be part of the initial beta launch for 0x Mesh. There are a few reasons for this:

1. As you discovered, there are some technical hurdles for implementing WeijieSub today. https://github.com/libp2p/go-libp2p-pubsub does not have an API that is conducive to writing your own router. We have been talking to some folks at libp2p/Protocol Labs and they are working to improve the situation. One concern we have is that if we fork go-libp2p-pubsub ourselves we will be taking on technical debt and will have to rewrite WeijieSub when new versions of go-libp2p-pubsub are released upstream with a different API.
2. We've realized that WeijieSub as it is described in the Architecture Doc has a major downside. It incentivizes participants in the network to share spam orders to increase their rank. Spam orders are ones that look valid but will never be filled (for example orders that are far away from the market price or ones that involve non-transferrable assets). We're worried that this version of WeijieSub will turn out to be worse than existing solutions in libp2p like FloodSub or GossipSub (where participants can still spam the network but at least they are not *incentivized* to do so). We have some ideas for how to improve WeijieSub to address this but there are a lot of angles to consider and it will take some time to formalize.
3. An efficient pubsub routing mechanism will be important if 0x Mesh grows the way that we want it to, but it will likely not be necessary early on. Today, even if you were to combine the orders from every relayer into one giant order book, the size of that global order book would still be be relatively small (we think on the order of megabytes). It would be smaller than some of the individual files that are stored and transferred in protocols like BitTorrent or IPFS. While it is important to plan for growth, we also want to avoid premature optimization, especially at a time where our efforts could be spent on testing and improving other components that *will* be necessary for the beta.

If you have any interest in tackling [a different issue or feature](https://github.com/0xProject/0x-mesh/issues), we would love to talk about it! Feel free to leave a comment on GitHub or join us on Discord, where we just created [a channel specifically for 0x Mesh](https://discordapp.com/channels/435912040142602260/583062159962013737).

p2p/node.go
```diff
@@ -52,7 +51,9 @@ const (
 	// advertiseTTL is the TTL for our announcement to the discovery network.
 	advertiseTTL = 5 * time.Minute
 	// pubsubProtocolID is the protocol ID to use for pubsub.
-	pubsubProtocolID = protocol.ID("/0x-mesh-floodsub/0.0.1")
+	// TODO(albrow): Is there a way to use a custom protocol ID with GossipSub?
+	// pubsubProtocolID = protocol.ID("/0x-mesh-floodsub/0.0.1")
```
fabioberger:
```suggestion
	// pubsubProtocolID = protocol.ID("/0x-mesh-gossipsub/0.0.1")
```

p2p/node.go
```diff
@@ -185,8 +186,9 @@ func New(config Config) (*Node, error) {
 	routingDiscovery := discovery.NewRoutingDiscovery(kadDHT)
 
 	// Set up pubsub.
-	// TODO: Replace with WeijieSub. Using FloodSub for now.
-	ps, err := pubsub.NewFloodsubWithProtocols(nodeCtx, basicHost, []protocol.ID{pubsubProtocolID})
+	pubsub.GossipSubHistoryGossip = 10
+	pubsub.GossipSubHistoryLength = 20
```
fabioberger: Can you add comments explaining what these configurations mean?

albrow: They affect the size of the internal message cache used by GossipSub.

I actually just decided to remove them. The idea was that since we are sharing old messages, we should have a larger message cache. But after I wrote this code I realized old messages get a new message ID so increasing the size of the message cache doesn't help us at all.

Issue #108, opened 2019-05-30
zeroex: Marshal []byte & number values in SignedOrder to hexidecimal


fabioberger:
Currently, when sending a 0x order to Mesh via `mesh_addOrders`, the JSON-RPC request looks like this:

```
{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "mesh_addOrders",
    "params": [
        [
            {
                "makerAddress": "0x6440b8c5f5a3c725eb394c7c40994afaf50a0d39",
                "makerAssetData": "9HJhsAAAAAAAAAAAAAAAAMAqqjmyI/6NCg5cTyfq2Qg8dWzC",
                "makerAssetAmount": 1233400000000000,
                "makerFee": 0,
                "takerAddress": "0x0000000000000000000000000000000000000000",
                "takerAssetData": "9HJhsAAAAAAAAAAAAAAAAA2HdfZIQwZ5pwnpjSsMtiUNKIfv",
                "takerAssetAmount": 1233400000000000,
                "takerFee": 0,
                "senderAddress": "0x0000000000000000000000000000000000000000",
                "exchangeAddress": "0x4f833a24e1f95d70f028921e27040ca56e09ab0b",
                "feeRecipientAddress": "0xa258b39954cef5cb142fd567a46cddb31a670124",
                "expirationTimeSeconds": 1560917245,
                "salt": 1545196045897,
                "signature": "G2pJMCd0sLDhTvWekfz5UN+321cFrmkp4GGYUYsRBTAdTvlLG0dg5VA3i7W3dGsaKcF0KQr+lEgyTO9BEt0D16ED"
            }
        ]
    ]
}
```

There are two issues to address here:

- The `[]byte` fields in the `SignedOrder` need to be sent over-the-wire in Base64 encoding. This is different from what is expected from Ethereum developers (`0x`prefixed and hex-encoded), and I feel strongly we  should adhere to these developer norms.
- The order submission fails if any of the amounts submitted are larger then 64 bits long (int64). This is probably why Ethereum JSON-RPC avoids passing any `uint256` values as numbers. Instead, they always return amounts hex-encoded (e.g., https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gasprice). We might need to do the same.

PR #120, opened 2019-05-31
zeroex: Implement custom JSON Marshal/Unmarshal methods for SignedOrder

fabioberger:
Fixes: #108

And fixes some whitespace issues.

### DIFF
--- a/zeroex/order.go
+++ b/zeroex/order.go
@@ -1,12 +1,14 @@
 package zeroex
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math/big"
 
 	"github.com/0xProject/0x-mesh/ethereum"
 	"github.com/0xProject/0x-mesh/ethereum/wrappers"
+	"github.com/0xProject/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/common"
 	signer "github.com/ethereum/go-ethereum/signer/core"
 	"golang.org/x/crypto/sha3"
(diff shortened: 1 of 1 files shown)


zeroex/order.go
```diff
@@ -208,9 +210,9 @@ func SignOrder(signer ethereum.Signer, order *Order) (*SignedOrder, error) {
 }
 
 // SignTestOrder signs the 0x order with the local test signer
-func SignTestOrder(order *Order) (*SignedOrder, error) {
+func SignTestOrder(order Order) (*SignedOrder, error) {
```
albrow: Is there a reason to use `Order` here instead of `*Order`? The latter is slightly more memory efficient and more closely matches the API of `SignOrder`.

Note that for `Order` struct inside `SignedOrder`, I wanted to change it to not be a pointer because `SignedOrder` is already passed around as a pointer pretty much everywhere. You don't gain any memory efficiency by having a pointer within a pointer. It's also easy to forget that embedded structs are pointer types they can be `nil`.

zeroex/order.go
```diff
@@ -237,6 +239,86 @@ func (s *SignedOrder) ConvertToOrderWithoutExchangeAddress() wrappers.OrderWitho
 	return orderWithoutExchangeAddress
 }
 
+type SignedOrderJSON struct {
```
albrow: I think this type should be unexported?

zeroex/order.go
fabioberger: Yeah agreed, reverted these changes.

PR #128, opened 2019-06-03
Added multi-stage Dockerfile, added a test with no race in Makefile

opaolini:
# About

As part of this PR I've:
* Added multi-stage Dockerfile 
* Added a task that tests with no race flag in `Makefile`.

The end image contains only the statically linked binary -> no dependency on `musl` / `glibc`

I am aware that there is a second branch here, specifically aiming to add the `Dockerfile` so please let me know if the target of this PR is wrong.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+# mesh-builder produces a statically linked binary
+FROM golang:1.12.1-alpine3.9 as mesh-builder
+
+
+RUN apk update && apk add ca-certificates nodejs-current npm make git dep
+RUN npm install -g yarn
+
+WORKDIR /go/src/github.com/0xProject/0x-mesh
+
+ADD . ./
+
+
+RUN make deps
+
+# RUN CGO_ENABLED=0 make test-all
+RUN CGO_ENABLED=0 go build ./cmd/mesh
+
+# Final Image
+FROM alpine:3.9
+
+RUN apk update && apk add ca-certificates --no-cache
+
+WORKDIR /usr/mesh
+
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+COPY --from=mesh-builder /go/src/github.com/0xProject/0x-mesh/mesh /usr/mesh/mesh
+
+
+ENV RPC_PORT=60557
+EXPOSE 60557
+
+RUN chmod +x ./mesh
+
+ENTRYPOINT ./mesh
```
fabioberger: Thoughts on using an SH script here that will continually re-boot Mesh if it panics?

```
#!/bin/sh

RC=1
while [ $RC -ne 0 ]; do
  ./mesh
  RC=$?
```

Mesh is built to resume where it left off after a crash.

opaolini: My initial thoughts would be to not do it via an SH script, and instead resort to some `recover()` mechanic inside mesh itself, where we specifically want to recover (goroutines, `JRPC` or `net/http` stuff).

However, we should allow the it to crash. If mesh will be running in a containerised environment, the operators can resort to either:
* a Docker restart policy: https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy
* K8s Pod lifecycle / restart policy: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

fabioberger: That makes sense. Thanks for the explanation.

opaolini:
Superseded by https://github.com/0xProject/0x-mesh/pull/137

Issue #86, opened 2019-05-26
rpc: Add paginated get all orders in DB method


fabioberger:
We have designed Mesh such that if it crashes for any reason, it resumes from where it left off upon re-start. This means clients will receive all order events despite the disruption. But what if the client is disrupted? The order event subscriptions won't queue up until their return.

In order to handle this scenario, we will add a paginated method for fetching all orders currently in the Mesh database. This will allow a client that has crashed to update all the orders in their DB/in-memory store and recover to an up-to-date snapshot of the valid orders stored by the Mesh node.

fabioberger:
Addressed in #198

Issue #22, opened 2019-04-25
db: Implement transactions


albrow:
We just need to build an API on top of [leveldb.Transaction](https://godoc.org/github.com/syndtr/goleveldb/leveldb#Transaction) that exposes transaction functionality to callers.

albrow:
Marking this for the Beta Release milestone since we realized it's required for #73.

PR #132, opened 2019-06-03
db: Add support for snapshots

albrow:
This is a prerequisite for https://github.com/0xProject/0x-mesh/issues/86. Without the ability to take a snapshot, we can't guarantee a consistent view of the current set of orders for RPC clients.

For the implementation, I first recognized that there are some common methods between [`leveldb.Snapshot`](https://godoc.org/github.com/syndtr/goleveldb/leveldb#Snapshot) and [`leveldb.DB`](https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB). Specifically, the `Get`, `Has`, and `NewIterator` methods. These are the methods that we rely on throughout the codebase (e.g. for creating and running queries). I encapsulated the common functionality in an interface called `dbReader`, and we now use `dbReader` instead of `*db.DB` wherever possible. This abstraction allows us to use the same underlying code for running queries, regardless of whether or not we are running them against a snapshot.

I also went ahead and abstracted the methods related to writing/updating the database. These will be useful when we add support for transactions in #22.

### DIFF
--- a/db/collection.go
+++ b/db/collection.go
@@ -13,7 +13,12 @@ import (
 
 // Collection represents a set of a specific type of model.
 type Collection struct {
-	db        *DB
+	readOnlyCollection
+	writerTransactor dbWriterTransactor
+}
+
+type readOnlyCollection struct {
+	reader    dbReader
 	name      string
 	modelType reflect.Type
 	indexes   []*Index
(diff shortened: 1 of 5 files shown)


db/collection.go
```diff
@@ -13,46 +13,69 @@ import (
 
 // Collection represents a set of a specific type of model.
 type Collection struct {
-	db        *DB
+	*readOnlyCollection
```
albrow: @fabioberger in one of your PRs, I suggested that we don't need to use a pointer for embedded fields. Usually, this is true. However, in this specific case, we need to mutate `readOnlyCollection` and `writeOnlyCollection` when calling embedded methods and this is the only way to make it work.

fabioberger:
What are the rules-of-thumb in deciding when to use a snapshot and when not?

albrow:
> What are the rules-of-thumb in deciding when to use a snapshot and when not?

You need to use snapshots whenever you are running more than one query and you don't want the state to change in between.

For example, for #86, we need to offer the ability to send multiple requests over RPC to paginate through the existing orders in the database. What happens if a new order is added to the database while you're in the middle of paginating? You might miss certain orders or see duplicates.

With a snapshot, we can guarantee that we return a consistent view of the orders that existed in the database at a particular point in time.

PR #136, opened 2019-06-05
Use a snapshot for GetMessagesToShare

albrow:
Without snapshots, there was a slight chance that we would return the wrong number of orders if the db state changed in between our two queries.


core/message_handler.go
```diff
@@ -140,7 +142,7 @@ func (app *App) HandleMessages(messages []*p2p.Message) error {
 		log.WithFields(map[string]interface{}{
 			"rejectedOrderInfo": rejectedOrderInfo,
 			"from":              msg.From.String(),
-		}).Warn("not storing rejected order received from peer")
+		}).Trace("not storing rejected order received from peer")
```
albrow: I also changed the verbosity for this message since it comes up a lot.

PR #137, opened 2019-06-05
Add multi-stage Dockerfile

opaolini:
# About

This PR adds a multi-stage `Dockerfile` that builds mesh without CGO.

The end alpine image contains only the statically linked binary -> no dependency on `musl` / `glibc`.

Supersedes https://github.com/0xProject/0x-mesh/pull/128 - removed the change to Makefile, and this PR is against master.


fabioberger:
Thanks for the contribution @opaolini!

Issue #93, opened 2019-05-27
zeroex/orderwatch/core: Add `EXPIRATION_BUFFER` env var


fabioberger:
Deciding when a 0x order is expired can be a tricky thing. Comparing the `ExpirationTimeSeconds` field with the current UTC time doesn't provide the full picture. The reason for this, is that the order's `ExpirationTimeSeconds` field is actually going to be compared against the `blockTime` of the block within which the miner is trying to include the fill transaction. 

Since Ethereum is a decentralized network of nodes which do not attempt to synchronize clocks, the only hard requirement for `blockTime`'s is that it must be greater then it's parent block's `blockTime`. As you might have already realized, that means the `blockTime` set by a miner might be _before_ or _after_ your Mesh node's current UTC time. 😢 

Let's recap: You can't assume current UTC time is what the miner is using for `blockTime`, and you have no way of knowing what `blockTime` they _are_ using. So what is there to do?

You try your best. And depending on what you are using 0x Mesh for, you might want to tweak the `EXPIRATION_BUFFER` when runnning your node. 

TODOs
- [x] Add `EXPIRATION_BUFFER` env var to Mesh configs and hook up to `OrderWatcher`
- [x] Decide on a sane default for `EXPIRATION_BUFFER`, so that orders don't end up expired by the time our neighbors validate the ones we send them

PR #140, opened 2019-06-05
core/orderwatch: Add OrderExpirationBuffer config to Mesh

fabioberger:
Fixes: https://github.com/0xProject/0x-mesh/issues/93

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -47,9 +47,9 @@ type Config struct {
 	// UseBootstrapList is whether to use the predetermined list of peers to
 	// bootstrap the DHT and peer discovery.
 	UseBootstrapList bool `envvar:"USE_BOOTSTRAP_LIST" default:"false"`
-	// OrderExpirationBuffer is the number of seconds before the order's stipulated expiration time
+	// OrderExpirationBuffer is the amount of time before the order's stipulated expiration time
 	// that you'd want it pruned from the Mesh node.
-	OrderExpirationBuffer int64 `envvar:"ORDER_EXPIRATION_BUFFER" default:"10"`
+	OrderExpirationBuffer time.Duration `envvar:"ORDER_EXPIRATION_BUFFER" default:"10s"`
 }
 
 type App struct {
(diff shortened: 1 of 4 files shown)


core/core.go
```diff
@@ -51,6 +51,9 @@ type Config struct {
 	// used for signing messages and generating a peer ID. If empty, a randomly
 	// generated key will be used.
 	PrivateKeyPath string `envvar:"PRIVATE_KEY_PATH" default:"./0x_mesh/key/privkey"`
+	// OrderExpirationBuffer is the number of seconds before the order's stipulated expiration time
+	// that you'd want it pruned from the Mesh node.
+	OrderExpirationBuffer int64 `envvar:"ORDER_EXPIRATION_BUFFER" default:"10"`
```
albrow: `ExpirationBuffer` should probably have type `time.Duration` instead of `int64`. The env var package we're using supports any type which implements `TextMarshalerUnmarshaler`, so if we changed the type to `time.Duration`, it would accept values like `10s`, `1h`, or `2h20m`.

If you want to keep the env var as `int64` I'm okay with it, but at the very least we should convert to `time.Duration` after parsing the env var.

What do you think?

fabioberger: Agreed! It actually lead to a necessary cleanup.

zeroex/orderwatch/expiration_watcher_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func TestPrunesExpiredOrders(t *testing.T) {
-	var expirationBuffer int64
+	expirationBuffer := 0 * time.Second
 	watcher := NewExpirationWatcher(expirationBuffer)
```
albrow: You don't have to add `* time.Second` here. Go can infer the type and 0 seconds is equal to zero of any other unit of time (0 * time.Second == 0 * time.Nanosecond). You can just do `NewExpirationWatcher(0)` or if you want to keep the separate variable declaration:

```
expirationBuffer := 0
watcher := NewExpirationWatcher(expirationBuffer)
```

PR #147, opened 2019-06-06
Fix bug where snapshot was not correctly isolated

albrow:
I discovered a bug in the snapshot feature introduced in #132. The tests were not sufficient to catch the bug, so this PR also improves the snapshot tests.

WIP because there is still an issue where `Delete` is not correctly isolated. When you call `Delete` after a snapshot is taken, the model also appears to be deleted in the snapshot.

### DIFF
--- a/db/collection.go
+++ b/db/collection.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sync"
 
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/iterator"
@@ -25,6 +26,7 @@ type readOnlyCollection struct {
 	name      string
 	modelType reflect.Type
 	indexes   []*Index
+	indexMut  sync.RWMutex
 }
 
 // writeableCollection is an extension of readonlyCollection which adds support
@@ -296,6 +298,8 @@ func (c *writeableCollection) Delete(id []byte) error {
 // deleteIndexesForModel deletes any indexes computed from the given model. It
 // *doesn't* discard the transaction if there is an error.
 func (c *writeableCollection) deleteIndexesForModel(txn *leveldb.Transaction, model Model) error {
+	c.indexMut.RLock()
+	defer c.indexMut.RUnlock()
 	for _, index := range c.indexes {
 		keys := index.keysForModel(model)
 		for _, key := range keys {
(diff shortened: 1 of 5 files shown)


db/collection.go
```diff
@@ -25,6 +26,7 @@ type readOnlyCollection struct {
 	name      string
 	modelType reflect.Type
 	indexes   []*Index
+	indexMut  sync.RWMutex
```
albrow: Since we're now copying the indexes whenever we create a snapshot, we need a mutex to protect them.

db/collection.go
```diff
@@ -25,6 +26,8 @@ type readOnlyCollection struct {
 	name      string
 	modelType reflect.Type
 	indexes   []*Index
+	// indexMut protects the indexes slice.
+	indexMut sync.RWMutex
```
albrow: Since we're now copying the indexes whenever we create a snapshot, we need a mutex to protect them.

albrow:
Found the last remaining bug and fixed it. Removing WIP tag. This PR is now ready for review.

db/snapshot.go
```diff
@@ -18,9 +18,18 @@ func (c *Collection) GetSnapshot() (*Snapshot, error) {
...
 		return nil, err
 	}
+	c.indexMut.RLock()
+	indexes := make([]*Index, len(c.indexes))
+	copy(indexes, c.indexes)
+	defer c.indexMut.RUnlock()
 	return &Snapshot{
-		readOnlyCollection: c.readOnlyCollection,
-		snapshot:           snapshot,
+		readOnlyCollection: &readOnlyCollection{
```
albrow: The main cause of the bug was that I forgot to create a new `readOnlyCollection` here with the `snapshot` as the reader.

db/snapshot_test.go
```diff
@@ -33,18 +33,28 @@ func TestSnapshot(t *testing.T) {
...
 	defer snapshot.Release()
 
-	// Any models we add after taking the snapshot should not be included in the
-	// query.
+	// Any models we add after taking the snapshot should not affect the query.
 	for i := 0; i < 5; i++ {
 		model := &testModel{
 			Name: "OtherPerson_" + strconv.Itoa(i),
-			Age:  i,
+			Age:  42,
```
albrow: There was also a bug in the tests. These models weren't showing up in the query results because the query is specifically looking for models where `Age == 42`.

db/snapshot.go
```diff
@@ -18,9 +18,18 @@ func (c *Collection) GetSnapshot() (*Snapshot, error) {
 	if err != nil {
 		return nil, err
 	}
+	c.indexMut.RLock()
+	indexes := make([]*Index, len(c.indexes))
+	copy(indexes, c.indexes)
+	defer c.indexMut.RUnlock()
```
fabioberger: Nit: Shouldn't this simply unlock here since we're done copying the indexes?

albrow: Nice catch! Yes that was my intention.

PR #152, opened 2019-06-06
Implement database transactions

### LINKED ISSUE #22: db: Implement transactions

albrow:
Fixes #22.

There are a lot of changes here. As part of implementing this new feature (and because I'm worried that #147 was a consequence of too much magic), I decided to simplify the interfaces we're using internally and reduce some of the magic.

Instead of relying on embedded structs like `readOnlyCollection` and `writableCollection`, I've decided to opt for a simpler approach. The main logic for most operations now exists as functions in the __operations.go__ file. Each of these functions accept the pieces that they need as arguments (usually `colInfo` and either `dbReader` or `dbWriter` depending on the type of operation).

Functions which use transactions internally are now split up into two parts. For example, the old `Collection.Update` method is now split into two functions: `update` and `updateWithTransaction`. `update` automatically creates and commits/discards the transaction, whereas `updateWithTransaction` accepts a transaction as an argument instead of creating one. This allows us to share a lot of the same underlying code between the `Collection` and `Transaction` types.

@fabioberger the commits in this PR are well-organized and isolated. If it makes it easier, you can review one commit at a time.

### DIFF
--- a/db/transaction.go
+++ b/db/transaction.go
@@ -1,0 +1,34 @@
+package db
+
+// type Transaction struct {
+// 	*writeableCollection
+// 	transaction *leveldb.Transaction
+// }
+
+// func (c *Collection) OpenTransaction() (*Transaction, error) {
+// 	snapshot, err := c.ldb.OpenTransaction()
+// 	if err != nil {
+// 		return nil, err
+// 	}
+// 	c.indexMut.RLock()
+// 	indexes := make([]*Index, len(c.indexes))
+// 	copy(indexes, c.indexes)
+// 	c.indexMut.RUnlock()
+// 	return &Transaction{
+// 		writeableCollection: &writeableCollection{
+// 			readOnlyCollection: &readOnlyCollection{
+// 				reader:    snapshot,
+// 				name:      c.name,
+// 				modelType: c.modelType,
+// 				indexes:   indexes,
+// 			},
+// 		},
+// 	}, nil
+// }
+
+// // Release releases the snapshot. This will not release any ongoing queries,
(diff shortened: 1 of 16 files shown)


albrow:
I updated the implementation to use a combination of `leveldb.Batch` and our own in-process global write lock. We might be able to improve the throughput further by only acquiring a write lock on the key or range of keys involved in the transaction, but I think this version should be good enough.

This changes one major property of transactions. In the new implementation, no writes/updates occur until the transaction is committed. I think this compromise is worth it and won't affect our use case.

db/transaction.go
```diff
@@ -1,89 +1,128 @@
...
-	return txn.txn.Commit()
+	txn.mut.Lock()
+	if err := txn.unsafeCheckState(); err != nil {
+		txn.mut.Unlock()
+		return err
+	}
+	txn.committed = true
+	txn.mut.Unlock()
+	defer txn.colInfo.writeMut.Unlock()
+	return txn.batchWriter.Write(txn.readerWithBatch.batch, nil)
```
fabioberger: You write after releasing the locks?

albrow: There are two mutexes. `colInfo.writeMux` is a per-collection global write lock. txn.mut` just protects the `committed` and `discarded` properties of the transaction. I thought it was okay set `committed = true` and release `txn.mut` before we actually write, but this is technically not correct. I updated the code so that `committed = true` is not set and the write lock is not released unless the commit was successful.

PR #156, opened 2019-06-09
Removed butt URL for mesh-ganache-cli reference

amaurer:
Removed butt URL for mesh-ganache-cli reference and replaced with existing DockerHub link.


albrow:
😑 

This is a result of a [Chrome extension](https://chrome.google.com/webstore/detail/cloud-to-butt-plus/apmlngnhgbnjpajelfkmabhkfapgnoai?hl=en) that I installed 8 years ago and forgot about. Thanks for the fix!

PR #157, opened 2019-06-10
zeroex/orderwatch: Validate order networkId & avoid sending entry order event payloads

### LINKED ISSUE #149: zeroex: Validate that `exchangeAddress` in order corresponds to the networkId

### LINKED ISSUE #150: rpc: Empty subscription RPC payloads being sent
Not sure what's causing this but I do see some  empty subscription payloads being sent to the JS WS client. 

```
{
	"subscription": "0x1068abf307ca9c6e57046b6b0764e39c",
	"result": []
}
```

On a pretty regular interval. Perhaps we are still emitting order events from the OrderWatcher with 0 events? If so, we should not be doing this.

fabioberger:
Fixes #149
Fixes #150

### DIFF
--- a/core/validation.go
+++ b/core/validation.go
@@ -5,6 +5,7 @@ package core
 import (
 	"fmt"
 
+	"github.com/0xProject/0x-mesh/constants"
 	"github.com/0xProject/0x-mesh/db"
 	"github.com/0xProject/0x-mesh/meshdb"
 	"github.com/0xProject/0x-mesh/p2p"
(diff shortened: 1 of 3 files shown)


core/validation.go
```diff
@@ -33,6 +34,10 @@ var (
 		Code:    "OrderAlreadyStored",
 		Message: "order is already stored",
 	}
+	ROIncorrectNetwork = zeroex.RejectedOrderStatus{
+		Code:    "OrderForIncorrectNetwork",
+		Message: "order was created for a different network then the one this Mesh node is configured to support",
```
albrow:
```suggestion
		Message: "order was created for a different network than the one this Mesh node is configured to support",
```

PR #159, opened 2019-06-11
core: Check for max expiration of orders

fabioberger:
Add a max order expiration check to our validation and set it to 30 days.

@weijiewu8 @albrow do you both agree with the 30 day expiration maximum?

### DIFF
--- a/core/validation.go
+++ b/core/validation.go
@@ -4,6 +4,8 @@ package core
 
 import (
 	"fmt"
+	"math/big"
+	"time"
 
 	"github.com/0xProject/0x-mesh/constants"
 	"github.com/0xProject/0x-mesh/db"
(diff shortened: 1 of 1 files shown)


albrow:
@fabioberger apologies if we talked about it before, but why should we have a max expiration?

weijiewu8:
> @fabioberger apologies if we talked about it before, but why should we have a max expiration?

I guess to prevent idle orders from staying in the mesh forever (e.g., someone proposed an order, set the expiration to be infinite, then price changed and no one is going to take this order, but the maker left or forgot the private key so that the order always stays there). If idle orders accumulate, they will eventually occupy a lot of storage space.

fabioberger:
@albrow it's just a final backstop against a useless order that still passes validation sticking around forever.

PR #162, opened 2019-06-13
rpc/mesh: Add heartbeat subscription

fabioberger:
Fixes: https://github.com/0xProject/0x-mesh/issues/170

The MARKET Protocol team requested the addition of a heartbeat subscription to the Mesh WS interface so that they could properly handle lost connections that might otherwise be hard to detect.

### DIFF
--- a/rpc/client_server_test.go
+++ b/rpc/client_server_test.go
@@ -24,9 +24,10 @@ import (
 // dummyRPCHandler is used for testing purposes. It allows declaring handlers
 // for some requests or all of them, depending on testing needs.
 type dummyRPCHandler struct {
-	addOrdersHandler         func(orders []*zeroex.SignedOrder) (*zeroex.ValidationResults, error)
-	addPeerHandler           func(peerInfo peerstore.PeerInfo) error
-	subscribeToOrdersHandler func(ctx context.Context) (*rpc.Subscription, error)
+	addOrdersHandler            func(orders []*zeroex.SignedOrder) (*zeroex.ValidationResults, error)
+	addPeerHandler              func(peerInfo peerstore.PeerInfo) error
+	subscribeToOrdersHandler    func(ctx context.Context) (*rpc.Subscription, error)
+	subscribeToHeartbeatHandler func(ctx context.Context) (*rpc.Subscription, error)
 }
 
 func (d *dummyRPCHandler) AddOrders(orders []*zeroex.SignedOrder) (*zeroex.ValidationResults, error) {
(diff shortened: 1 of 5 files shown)


rpc/client_server_test.go
```diff
@@ -196,3 +204,28 @@ func TestOrdersSubscription(t *testing.T) {
 	// The WaitGroup signals that AddOrder was called on the server-side.
 	wg.Wait()
 }
+
+func TestHeartbeatSubscription(t *testing.T) {
+	ctx := context.Background()
+
+	// Set up the dummy handler with a subscribeToHeartbeatHandler
+	wg := &sync.WaitGroup{}
```
albrow: Can we use a channel to signal when the `subscribeToHeartbeatHandler` was called instead of a `sync.WaitGroup`? It would improve the failure mode a lot. Right now the test will hang for 10 minutes (the default test timeout in Go) and eventually timeout, but we could easily fail faster. Something like:

```go
rpcHandlerCalled := make(chan struct{})
rpcHandler := &dummyRPCHandler{
    subscribeToHeartbeatHandler: func(ctx context.Context) (*rpc.Subscription, error) {
        rpcHandlerCalled <- struct{}{}
        return nil, nil
    },
}

// The select statement allows us to timeout sooner than the default test timeout.
select {
    case <-time.After(5 * time.Second):
        t.Error("timed out waiting for rpcHandler to be called")
    case <-rpcHandlerCalled:
        break
}
```

rpc/service.go
```diff
@@ -21,15 +21,22 @@ type RPCHandler interface {
...
 }
 
-// Orders calls rpcHandler.Orders and returns the rpc subscription.
+// Orders calls rpcHandler.SubscribeToOrders and returns the rpc subscription.
 func (s *rpcService) Orders(ctx context.Context) (*rpc.Subscription, error) {
 	return s.rpcHandler.SubscribeToOrders(ctx)
 }
 
+// Heartbeat calls rpcHandler.SubscribeToHeartbeat and returns the rpc subscription.
+func (s *rpcService) Heartbeat(ctx context.Context) (*rpc.Subscription, error) {
```
albrow: I think we should consider implementing `Heartbeat` directly in `rpcService` instead of making it part of the `RPCHandler` interface. The other methods require knowledge of the database, watchers, and other packages; it would break separation of concerns to implement them in `rpcService`. But the `Heartbeat` method doesn't depend on the database, watchers, or any other packages. `Heartbeat` is really just a concern of the `rpc` package itself. What do you think?

fabioberger: Yep, makes sense. Moved it.

rpc/client_server_test.go
```diff
@@ -50,6 +51,13 @@ func (d *dummyRPCHandler) SubscribeToOrders(ctx context.Context) (*rpc.Subscript
 	return d.subscribeToOrdersHandler(ctx)
 }
 
+func (d *dummyRPCHandler) SubscribeToHeartbeat(ctx context.Context) (*rpc.Subscription, error) {
```
albrow: We don't need a `SubscribeToHeartbeat` method on `dummyRPCHandler` anymore since it is not part of the `RPCHandler`.

PR #167, opened 2019-06-14
WIP: check ETH backing when inserting orders

albrow:
WIP for now. Pausing for a bit while we figure out if there is a simpler way to do this.


albrow:
Okay most of the implementation is in place at this point. I still need to add more tests and implement updating ETH backings when we receive balance changes from the ETH watcher.

meshdb/meshdb.go
```diff
@@ -201,18 +269,154 @@ func (m *MeshDB) FindLatestMiniHeader() (*MiniHeader, error) {
 	return miniHeaders[0], nil
 }
 
+func (m *MeshDB) findEthBackingsWithLessEthPerOrder(target *big.Rat, max int) ([]*ETHBacking, error) {
```
fabioberger:
```suggestion
func (m *MeshDB) findEthBackingsWithLeastEthPerOrder(target *big.Rat, max int) ([]*ETHBacking, error) {
```
?

meshdb/meshdb.go
```diff
@@ -201,18 +269,154 @@ func (m *MeshDB) FindLatestMiniHeader() (*MiniHeader, error) {
...
+		return fmt.Errorf("invalid database state: total number of orders (%d) is greater than max orders (%d)", totalExistingOrders, m.maxOrders)
+	}
+
+	var ethBackingForIncomingOrder ETHBacking
+	if err := m.ETHBackings.FindByID(order.SignedOrder.MakerAddress.Bytes(), &ethBackingForIncomingOrder); err != nil {
+		if _, ok := err.(db.NotFoundError); ok {
+			// This should never happen because the ETHWatcher should have already
+			// been notified about this incoming order. We can't fix the problem here
+			// because meshdb doesn't know about the ETHWatcher (if it did it would
+			// violate separation of concerns).
```
fabioberger: Sure you're talking about ETHWatcher here? I think you mean `core`'s validation logic (which inserts EthBackings)

core/validation.go
```diff
@@ -166,3 +180,139 @@ func (app *App) orderAlreadyStored(orderHash common.Hash) (bool, error) {
...
+			ordersByMakerAddress[order.MakerAddress] = []*zeroex.SignedOrder{order}
+		}
+	}
+	var makerAddresses []common.Address
+	for makerAddress := range ordersByMakerAddress {
+		makerAddresses = append(makerAddresses, makerAddress)
+	}
+
+	// Get the current ETH balance for each maker.
+	addressToBalance, failedAddresses := app.ethWathcher.Add(makerAddresses)
```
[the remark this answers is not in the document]
0xpeter: I think we are going to end up with the conclusion that anything that is rebroadcast will need to be stored in some way. Otherwise we will be able to suck up bandwidth by repeatedly forcing rebroadcast of the same set of bogus messages. It doesn't need to be the full message, just a hash will do.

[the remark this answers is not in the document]
albrow: > This is a bit strange. We are going to be tracking makerAddresses, even of orders that won't ultimately get stored... Is that desirable?

@fabioberger I think it is desirable. Even in the current implementation which doesn't store anything for invalid orders, caching these balances will make it more efficient to check incoming orders from makers we have already seen.

> Also, `ethWathcher.Add` as it's currently written only returns the balances of previously unknown ETH addresses passed in. We should probably refactor so it returns balances for all passed in addresses.

@fabioberger good point. Yeah we will need to address this.

> I think we are going to end up with the conclusion that anything that is rebroadcast will need to be stored in some way. Otherwise we will be able to suck up bandwidth by repeatedly forcing rebroadcast of the same set of bogus messages. It doesn't need to be the full message, just a hash will do.

@0xpeter we *probably* want to do this to some extent but not in this PR.

meshdb/meshdb.go
albrow: You're right; this comment should be updated. I wrote it before implementing the validation logic in the `core` package.

albrow:
We now have an algorithm for validating ETH backings for a batch of orders which runs in approximately linear time with respect to the number of incoming orders. However benchmarks reveal that even though it scales nearly linearly it's still too slow:

```
BenchmarkValidateETHBackings1Account100Orders-12       	    1000	   2210667 ns/op
BenchmarkValidateETHBackings1Account1000Orders-12      	      50	  23444738 ns/op
BenchmarkValidateETHBackings1Account10000Orders-12     	       5	 219615431 ns/op
BenchmarkValidateETHBackings5Accounts100Orders-12      	     300	   4359196 ns/op
BenchmarkValidateETHBackings5Accounts1000Orders-12     	      30	  33445863 ns/op
BenchmarkValidateETHBackings5Accounts10000Orders-12    	       5	 228234687 ns/op
```

These benchmarks are only testing the core part of the algorithm which keeps everything in memory and does not query the database or send any network requests.

The algorithm takes about 23-33ms to validate 1000 orders, so we should expect to see about 11-16ms for our current maximum batch size of 500 orders. This is probably not fast enough for something that is going to happen frequently and requires a global database lock. I'm looking into ways to optimize it starting by looking at the `big` and `heap` packages.

meshdb/meshdb.go
```diff
@@ -201,18 +265,142 @@ func (m *MeshDB) FindLatestMiniHeader() (*MiniHeader, error) {
...
 }
 
+// InsertOrder atomically inserts the given order and updates any relevant ETH
+// backings. It also removes the order with the lowest ETH backing if needed in
+// order to make room. If order storage is full and the given order does not
+// have a high enough ETH backing, it will not be stored (this is not considered
+// an error). InsertOrder will return an error if any corresponding ETH backings
+// cannot be found.
+// TODO: return hash of order that was deleted.
+func (m *MeshDB) InsertOrder(order *Order) error {
```
albrow: @fabioberger we need to eventually return the order hash for any order that was deleted so that we can send an update to RPC clients. Are you okay with punting on this for now? It likely won't come up in the near future due to the fact that the total number of current open order book orders is much lower than the default limit.

albrow:
Almost done. The last known thing that I need to do is update the corresponding `ETHBacking` whenever `OrderWatcher` deletes an order.

albrow:
Latest benchmarks for reference:

```
BenchmarkValidateETHBackings1Address100Orders-12          	  100000	     12281 ns/op
BenchmarkValidateETHBackings1Address1000Orders-12         	   10000	    113263 ns/op
BenchmarkValidateETHBackings10Addresses100Orders-12       	   50000	     26999 ns/op
BenchmarkValidateETHBackings10Addresses1000Orders-12      	   10000	    189115 ns/op
BenchmarkValidateETHBackings100Addresses100Orders-12      	  100000	     18083 ns/op
BenchmarkValidateETHBackings100Addresses1000Orders-12     	    2000	    605147 ns/op
BenchmarkValidateETHBackings1000Addresses100Orders-12     	   30000	     57250 ns/op
BenchmarkValidateETHBackings1000Addresses1000Orders-12    	   10000	    165459 ns/op
```

It's hard to do an exact comparison because I increased the maximum number of addresses covered in the latest benchmarks. The original implementation in this PR took 23-33ms to validate 1000 orders across 5 addresses. This would not have been feasible for a greater number of addresses (we expect a maximum of 1,000 addresses per batch). The current implementation takes ~165 microseconds to validate 1,000 orders across 1,000 addresses. Due to having a greater number of orders per maker, it is actually slower with 1,000 orders across 100 addresses, at ~600 microseconds. In any case, it several orders of magnitude faster than the original implementation and is definitely feasible for production use.

meshdb/meshdb.go
fabioberger: Yep, we can punt but can you please open an issue so we don't forget?

albrow:
Given the high number of remaining issues (some of which have game theoretic implications that need to be worked out), we have decided to delay this feature until after the beta release. Leaving this PR open in the meantime.

albrow:
This PR is really out of date at this point so I'm going to go ahead and close it. We might still refer to it/borrow some code from it if we ever implement this feature in the future, but it's going to be a while before that happens.

Issue #168, opened 2019-06-15
ethereum: Refactor ethWatcher interface


fabioberger:
Speaking with @albrow about the implementation of #73, we realized the interface for `ethWatcher` isn't ideal. We should make the following changes:

- Let someone add an array of Ethereum addresses
- Upon adding Ethereum addresses, `ethWatcher` should fetch their Ethereum balances and return them to the caller.
- Update the interval in which ETH balances are updated to every minute.

PR #175, opened 2019-06-17
ethereum: Refactor EthWatcher

fabioberger:
Fixes: #168

### DIFF
--- a/ethereum/blockchain_lifecycle.go
+++ b/ethereum/blockchain_lifecycle.go
@@ -2,9 +2,10 @@ package ethereum
 
 import (
 	"context"
-	"fmt"
+	"testing"
 
 	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/stretchr/testify/require"
 )
 
 type BlockchainLifecycle struct {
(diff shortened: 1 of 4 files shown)


ethereum/eth_watcher.go
```diff
@@ -134,14 +133,41 @@ func (e *ETHWatcher) Receive() <-chan Balance {
 	return e.balanceChan
 }
 
-func (e *ETHWatcher) updateBalances() error {
+func (e *ETHWatcher) updateBalances() {
 	e.addressToBalanceMu.Lock()
+	e.addressToBalanceMu.Unlock()
```
albrow: I think you meant:

```suggestion
	defer e.addressToBalanceMu.Unlock()
```

ethereum/eth_watcher.go
```diff
@@ -134,14 +133,41 @@ func (e *ETHWatcher) Receive() <-chan Balance {
...
 	}
-	e.addressToBalanceMu.Unlock()
+	// Intentionally ignore addresses we failed to fetch balances for
+	// and simply attempt them again at the next polling interval
+	addressToBalance, _ := e.getBalances(addresses)
+	for address, newBalance := range addressToBalance {
+		if cachedBalance, ok := e.addressToBalance[address]; ok {
+			if cachedBalance.Cmp(newBalance) != 0 {
+				e.addressToBalance[address] = newBalance
+				updatedBalance := Balance{
```
albrow: nit: The names `updatedBalance` and `newBalance` seem like synonyms to me but they have different purposes here. The fact that the `Balance` type has a field called `Balance` is also sort of confusing.

What do think about this naming scheme instead?

```go
type Balance struct {
    Address common.Address
    Amount *big.Int
}
```

The idea here is that an `Amount` by itself is just a number. We only call it a `Balance` when it is associated with an account.

This would lend clarity to the code I mentioned by changing it to:

```go
for address, newAmount := range addressToAmount {
    // ...
    updatedBalance := Balance{
        Address: address,
        Amount: newAmount,
    }
}
```

What do you think?

fabioberger: I like it! Much clearer. Thanks for the suggestion.

ethereum/blockchain_lifecycle.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/stretchr/testify/require"
+)
+
+type BlockchainLifecycle struct {
+	rpcClient       *rpc.Client
+	snapshotIdStack []string
+}
+
+func NewBlockchainLifecycle(rpcURL string) (*BlockchainLifecycle, error) {
```
albrow: Sorry I forgot earlier. We should probably add some documentation to this file. I recognize this from the TypeScript version, but other people reading/contributing to the code might not get it right away.

Issue #178, opened 2019-06-19
mesh: fillableTakerAssetAmount is a number


dekz:
```
{
  "jsonrpc": "2.0",
  "method": "mesh_subscription",
  "params": {
    "subscription": "0xdb7aae70131ae373240aca8fb3e3669",
    "result": [
      {
        "orderHash": "0x4bcabce9c0e4a05251caec5e3e69ae98970b1fe758d8d835ef496a2cdcd23d03",
        "signedOrder": {
          "makerAssetAmount": "8313702600",
          "makerFee": "0",
          "takerAssetAmount": "8400000000000000000000",
          "takerFee": "0",
...
        },
        "kind": "EXPIRED",
        "fillableTakerAssetAmount": 0,
        "txHash": "0x0000000000000000000000000000000000000000000000000000000000000000"
      }
    ]
  }
}
```

All other amounts are strings. This may occur all the time or may only occur when the `fillableTakerAssetAmount` is 0.

PR #179, opened 2019-06-19
Fix `fillableTakerAssetAmount` in order event subscription payload

fabioberger:
Fixes: #178 

Add custom MarshalJSON method for OrderEvent so that `fillableTakerAssetAmount` is sent over the wire as a stringified amount and never in scientific notation.

### DIFF
--- a/zeroex/order.go
+++ b/zeroex/order.go
@@ -81,6 +81,17 @@ type OrderEvent struct {
 	TxHash common.Hash `json:"txHash"`
 }
 
+// MarshalJSON implements a custom JSON marshaller for the SignedOrder type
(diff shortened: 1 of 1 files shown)


albrow:
@fabio what do you think about implementing our own wrapper type around `big.Int` as [described here](https://stackoverflow.com/questions/53991835/how-to-marshal-and-unmarshal-big-int-in-json)? If `big.Int` is the only type causing us trouble when marshaling to JSON this could save us some lines of code. I'm also okay with merging this for now and creating a separate issue.

fabioberger:
That is a great idea. I love it. I'll create an issue for doing this refactor and will tackle it in a subsequent PR.

PR #181, opened 2019-06-19
blockwatch/core: Fix Kovan incorrect block hash bug

fabioberger:
MarkM mentioned Mesh not firing events when connected to Kovan on our discord.

Turns out Kovan uses a different structure for block headers then mainnet that isn't `rlp` compatible and since we are using go-ethereum's `EthClient`, it is generating incorrect block hashes for Kovan blocks. More details on this discrepancy between mainnet and Kovan here: https://github.com/ethereum/go-ethereum/pull/18166.

The fix is to use a raw JSON RPC request to `eth_blockByNumber` which returns a pre-computed block hash that is always properly formatted instead of re-computing the block hash given the block header struct returned from `EthClient`'s `BlockByNumber()` method.

Additionally, this PR also logs all BlockWatcher errors, which were not being logged previously.

### DIFF
--- a/ethereum/blockwatch/block_watcher.go
+++ b/ethereum/blockwatch/block_watcher.go
@@ -118,8 +118,8 @@ func (w *Watcher) startPollingLoop() {
 	}
 }
 
-// StopPolling stops the block poller
-func (w *Watcher) StopPolling() {
+// stopPolling stops the block poller
+func (w *Watcher) stopPolling() {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	w.isWatching = false
@@ -132,7 +132,7 @@ func (w *Watcher) StopPolling() {
 // Stop stops the BlockWatcher
 func (w *Watcher) Stop() {
 	if w.isWatching {
-		w.StopPolling()
+		w.stopPolling()
 	}
 	close(w.Errors)
 }
--- a/ethereum/blockwatch/block_watcher.go
+++ b/ethereum/blockwatch/block_watcher.go
@@ -126,6 +126,7 @@ func (w *Watcher) StopPolling() {
 	if w.ticker != nil {
 		w.ticker.Stop()
 	}
+	close(w.Errors)
 	w.ticker = nil
 }
 
--- a/ethereum/blockwatch/block_watcher.go
+++ b/ethereum/blockwatch/block_watcher.go
@@ -126,8 +126,15 @@ func (w *Watcher) StopPolling() {
 	if w.ticker != nil {
 		w.ticker.Stop()
 	}
-	close(w.Errors)
 	w.ticker = nil
+}
+
+// Stop stops the BlockWatcher
+func (w *Watcher) Stop() {
+	if w.isWatching {
+		w.StopPolling()
+	}
+	close(w.Errors)
 }
 
 // Subscribe allows one to subscribe to the block events emitted by the Watcher.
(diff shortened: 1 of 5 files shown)


ethereum/blockwatch/block_watcher.go
```diff
@@ -129,6 +129,14 @@ func (w *Watcher) StopPolling() {
 	w.ticker = nil
 }
 
+// Stop stops the BlockWatcher
+func (w *Watcher) Stop() {
```
albrow: Should `StopPolling` be unexported? It looks like things will break if you call it directly instead of via `Stop`.

ethereum/blockwatch/client.go
```diff
@@ -22,29 +26,64 @@ type Client interface {
...
+func NewRpcClient(rpcURL string, requestTimeout time.Duration) (*RpcClient, error) {
+	ethClient, err := ethclient.Dial(rpcURL)
+	if err != nil {
+		return nil, err
+	}
+	rpcClient, err := rpc.Dial(rpcURL)
+	if err != nil {
+		return nil, err
+	}
+	return &RpcClient{rpcClient, ethClient, requestTimeout}, nil
```
albrow: We should use explicit field names here. Otherwise the compiler won't help us if we add new fields in the future.

```suggestion
	return &RpcClient{rpcClient: rpcClient, client: ethClient, requestTimeout: requestTimeout}, nil
```

ethereum/blockwatch/client.go
```diff
@@ -22,29 +26,64 @@ type Client interface {
...
+}
+
+type GetBlockByNumberResponse struct {
+	Hash       common.Hash `json:"hash"`
```
albrow: For future reference: if you're just unmarshaling, you don't need to add these `json` struct tags. The `encoding/json` package will account for the fact that cases might be different in Go vs. JSON and will automatically unmarshal `hash` into `Hash`. It also doesn't break anything to include them so it's fine for now.

ethereum/blockwatch/client.go
```diff
@@ -22,29 +26,64 @@ type Client interface {
...
+		blockParam = "latest"
+	} else {
+		blockParam = fmt.Sprintf("0x%s", common.Bytes2Hex(number.Bytes()))
+	}
+	shouldIncludeTransactions := false
+
+	// Note(fabio): We use a raw RPC call here instead of `EthClient`'s `BlockByNumber()` method because block
+	// hashes are computed differently on Kovan vs. mainnet, resulting in the wrong block hash being returned by
+	// `BlockByNumber` when using Kovan. By doing a raw RPC call, we can simply use the blockHash returned in the
+	// RPC response rather then re-compute it from the block header.
```
albrow:
```suggestion
	// RPC response rather than re-compute it from the block header.
```

ethereum/blockwatch/block_watcher_test.go
```diff
@@ -85,7 +85,7 @@ func TestWatcherStartStop(t *testing.T) {
 	}
 	watcher := New(config)
 	require.NoError(t, watcher.StartPolling())
-	watcher.StopPolling()
+	watcher.stopPolling()
```
albrow: I think we should be calling `Stop` here now instead of `stopPolling`?

HACKNATHON001:
not kovan issue mainet ethereum do not licennse spam issue..if im closed this project same stuck overall

PR #185, opened 2019-06-19
blockwatch/core: Fix block not found, make polling interval configurable

fabioberger:
- Make block polling interval a config.
- Fix edge-case where block not found.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -24,12 +24,11 @@ import (
 )
 
 const (
-	blockWatcherPollingInterval = 2 * time.Second
-	blockWatcherRetentionLimit  = 20
-	ethereumRPCRequestTimeout   = 30 * time.Second
-	ethWatcherPollingInterval   = 1 * time.Minute
-	peerConnectTimeout          = 60 * time.Second
-	checkNewAddrInterval        = 20 * time.Second
+	blockWatcherRetentionLimit = 20
+	ethereumRPCRequestTimeout  = 30 * time.Second
+	ethWatcherPollingInterval  = 1 * time.Minute
+	peerConnectTimeout         = 60 * time.Second
+	checkNewAddrInterval       = 20 * time.Second
 )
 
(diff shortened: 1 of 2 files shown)


ethereum/blockwatch/client.go
```diff
@@ -75,6 +76,10 @@ func (rc *RpcClient) HeaderByNumber(number *big.Int) (*meshdb.MiniHeader, error)
 	if err != nil {
 		return nil, err
 	}
+	// If it returned an empty struct, return NotFound error
+	if reflect.DeepEqual(header, reflect.Zero(reflect.TypeOf(header)).Interface()) {
```
albrow: We don't need to use `reflect` to create the zero value if we already know the type ahead of time.

```suggestion
	if reflect.DeepEqual(header, GetBlockByNumberResponse{}) {
```

But even better, if there are some fields of `GetBlockByNumberResponse` that we know should always be present, we could just check those fields and avoid using reflect altogether. For example:

```go
if header.Number == "" {
    // ...
}
```

core/core.go
```diff
@@ -57,6 +56,11 @@ type Config struct {
 	// OrderExpirationBuffer is the amount of time before the order's stipulated expiration time
 	// that you'd want it pruned from the Mesh node.
 	OrderExpirationBuffer time.Duration `envvar:"ORDER_EXPIRATION_BUFFER" default:"10s"`
+	// BlockPollingInterval is the polling interval to wait before checking for a new Ethereum block
+	// that might contain transactions that impact the fillability of orders stored by Mesh. Different
+	// networks have different block producing intervals: POW networks are typicall slower (e.g., Mainnet)
```
```suggestion
	// networks have different block producing intervals: POW networks are typically slower (e.g., Mainnet)
```

ethereum/blockwatch/client.go
fabioberger: Agreed. Let's go with `if header.Number == "" {` here.

Issue #188, opened 2019-06-21
orderwatch: Dedup orders to revalidate from a batch of events


fabioberger:
Currently, we process the block logs (events) sequentially and each one might cause us to re-validate an order. Multiple events could cause the same order to get re-validated. What we should do instead is compute the set of orders to revalidate given an array of events, and only once we've generated this deduped set, go on to the next step of re-validating the orders.

Issue #56, opened 2019-05-15
blockwatch: Implement fast-sync in BlockWatcher


fabioberger:
Currently, every time a Mesh node is re-started, it does two things:
1. Starts fetching blocks from where it last left off and begins the process of catching up with the latest block, processing all events sequentially. 
2. Loads all orders from the DB into the OrderWatcher and starts re-validating the orders as the contract events are emitted.

If it's been a very long time since the node was last online however, it could take a while for the node to catch up with the latest block. Instead, we should check if we've fallen behind on start-up and fast-sync using `eth_getLogs` block range queries to catch up quickly with the latest block. We are able to fetch and process up to 60 blocks worth of events per query.

fabioberger:
In additional to a `max blocks elapsed` condition, if it's more then ~10 blocks behind, we should perform a fast-sync that fetches the logs for multiple blocks in batch queries and more quickly catches the node back up with the latest block. This limits the time after boot-up that the node will be out-of-sync with the tip of the chain.

PR #189, opened 2019-06-24
blockwatch: Implement fast-sync

fabioberger:
Fixes: #56  
Fixes: #188 

This PR:
- Implements fast-sync which checks if the Mesh node has fallen behind the latest block since the last time it was started, and if so, does limited-concurrency batch requests to `eth_getLogs` (60 blocks at a time) and catches up with the latest block. It then processes all the events found and re-validates any orders impacted by them.
- Improves the OrderWatcher's efficiency by processing all events emitted by BlockWatcher at once and computes a de-duped list of orders whose validity might be impacted. These orders are then batch re-validated.
- Refactors blockwatch.FakeClient and added blockwatch.FakeLogClient in order to write more robust unit tests for BlockWatcher.


fabioberger:
Update: Parity, Geth and Alchemy have much higher limits on the number of logs they return in a single request when compared with Infura. They should all be able to handle the 60 block range query without issue.

core/core.go
```diff
@@ -194,14 +210,20 @@ func (app *App) Start() error {
...
-	}
-	log.Info("started block watcher")
 	if err := app.orderWatcher.Start(); err != nil {
 		return err
 	}
 	log.Info("started order watcher")
+	if err := app.blockWatcher.Start(); err != nil {
+		return err
+	}
+	log.Info("started block watcher")
```
fabioberger: We actually want to start `BlockWatcher` after `OrderWatcher` so that `OrderWatcher` is already setup to receive the block events when `BlockWatcher` starts.

ethereum/blockwatch/block_watcher_test.go
```diff
@@ -67,25 +72,398 @@ func TestWatcher(t *testing.T) {
...
+				"10-20": filterLogsResponse{
+					Err: errUnexpected,
+				},
+			},
+			Err: errUnexpected,
+		},
+	}
+
+	meshDB, err := meshdb.NewMeshDB("/tmp/leveldb_testing/" + uuid.New().String())
+	require.NoError(t, err)
```
albrow: This is far from the only place this happens, but we should remember to call `Close` when instantiating a new `MeshDB` in our tests. I created https://github.com/0xProject/0x-mesh/issues/209 to address this.

ethereum/blockwatch/block_watcher_test.go
```diff
@@ -67,25 +72,398 @@ func TestWatcher(t *testing.T) {
...
+		watcher := New(config)
+
+		logs, furthestBlockProcessed := watcher.getLogsInBlockRange(int64(testCase.From), int64(testCase.To))
+		require.Equal(t, testCase.FurthestBlockProcessed, furthestBlockProcessed, testCase.Label)
+		require.Equal(t, testCase.Logs, logs, testCase.Label)
+		assert.Equal(t, len(testCase.RangeToFilterLogsResponse), fakeLogClient.Count())
+	}
+}
+
+func toR(from, to int) string {
```
albrow: Would help make this function a little more clear.

```suggestion
func toRange(from, to int) string {
```

ethereum/blockwatch/block_watcher_test.go
```diff
@@ -67,25 +72,398 @@ func TestWatcher(t *testing.T) {
...
+			},
+			Logs:                   []types.Log{logStub},
+			FurthestBlockProcessed: int64(to),
+		},
+		logsInBlockRangeTestCase{
+			Label: "SPLIT_REQUEST_BY_MAX_BLOCKS_IN_QUERY",
+			From:  from,
+			To:    from + maxBlocksInQuery + 10,
+			RangeToFilterLogsResponse: map[string]filterLogsResponse{
+				toR(from, from+maxBlocksInQuery-1): filterLogsResponse{
```
albrow: We should consider using constants in the range here just to make sure that the result is exactly what we expect. Using arithmetic runs the risk of accidentally introducing a bug in the test cases.

ethereum/blockwatch/fake_log_client.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type filterLogsResponse struct {
+	Logs []types.Log
+	Err  error
+}
+
+// fakeLogClient is a fake Client for testing purposes.
```
albrow: Can you include in this comment how `fakeLogClient` differs from `fakeClient`?

ethereum/blockwatch/fake_log_client.go
```diff
@@ -0,0 +1,65 @@
...
+func (fc *fakeLogClient) HeaderByHash(hash common.Hash) (*meshdb.MiniHeader, error) {
+	return nil, errors.New("NOT_IMPLEMENTED")
+}
+
+// FilterLogs returns the logs that satisfy the supplied filter query
+func (fc *fakeLogClient) FilterLogs(q ethereum.FilterQuery) ([]types.Log, error) {
+	// Add a slight delay to simulate an actual network request. This also gives
+	// BlockWatcher.getLogsInBlockRange multi-requests to hit the concurrent request
+	// limit semaphore and simulate more realistic conditions.
+	<-time.Tick(5 * time.Millisecond)
```
albrow:
```suggestion
	time.Sleep(5 * time.Millisecond)
```

zeroex/order.go
```diff
@@ -77,18 +77,23 @@ type OrderEvent struct {
...
 	Kind                     OrderEventKind `json:"kind"`
 	FillableTakerAssetAmount *big.Int       `json:"fillableTakerAssetAmount"`
-	// The hash of the Ethereum transaction that caused the order status to change
-	TxHash common.Hash `json:"txHash"`
+	// The hashes of the Ethereum transactions that caused the order status to change.
+	// Could be because of multiple transactions, not just a single transaction.
+	TxHashes []common.Hash `json:"txHashes"`
 }
 
 // MarshalJSON implements a custom JSON marshaller for the SignedOrder type
```
albrow:
```suggestion
// MarshalJSON implements a custom JSON marshaller for the OrderEvent type
```

albrow: Sorry I missed this earlier.

zeroex/orderwatch/order_watcher.go
```diff
@@ -392,7 +406,11 @@ func (w *Watcher) setupEventWatcher() {
 								w.handleDecodeErr(err, eventType)
 								continue
 							}
-							w.findOrderAndGenerateOrderEvents(exchangeFillEvent.OrderHash, log.TxHash)
+							orders = []*meshdb.Order{}
```
albrow: `orders` has already been declared and initialized above, so we don't need to redeclare/re-initialize it.

zeroex/orderwatch/order_watcher.go
```diff
@@ -410,69 +432,89 @@ func (w *Watcher) setupEventWatcher() {
...
+	if len(signedOrders) == 0 {
+		return // Noop
 	}
 	validationResults := w.orderValidator.BatchValidate(signedOrders)
 
 	orderEvents := []*zeroex.OrderEvent{}
 	for _, acceptedOrderInfo := range validationResults.Accepted {
-		order := orderHashToOrder[acceptedOrderInfo.OrderHash]
+		orderWithTxHashes := hashToOrderWithTxHashes[acceptedOrderInfo.OrderHash]
+		txHashes := []common.Hash{}
```
albrow: If we know the length of `txHashes` ahead of time, we should initialize using `make`:

```suggestion
		txHashes := make([]common.Hash{}, len(orderWithTxHashes.TxHashes))
```


Then:

```go
for i, txHash := range orderWithTxHashes.TxHashes {
    txHashes[i] = txHash
}
```

zeroex/orderwatch/order_watcher.go
```diff
@@ -533,12 +576,16 @@ func (w *Watcher) generateOrderEventsIfChanged(orders []*meshdb.Order, txHash co
 				if !ok {
 					logger.WithField("rejectedOrderStatus", rejectedOrderInfo.Status).Panic("No OrderEventKind corresponding to RejectedOrderStatus")
 				}
+				txHashes := []common.Hash{}
```
albrow: Same here re: `make`.

core/core.go
```diff
@@ -61,6 +62,12 @@ type Config struct {
 	// networks have different block producing intervals: POW networks are typically slower (e.g., Mainnet)
 	// and POA networks faster (e.g., Kovan) so one should adjust the polling interval accordingly.
 	BlockPollingInterval time.Duration `envvar:"BLOCK_POLLING_INTERVAL" default:"5s"`
+	// DropDB instructs Mesh to drop the database before starting up. If it's been a long time since you ran
+	// Mesh and there've been 10,000's of Ethereum blocks mined since then, many of your orders are probably no
+	// longer valid and it's maybe not worth fast-syncing 10,000's of blocks to emit order events for them.
+	// Instead, you might want to drop the DB and start from scratch. This feature is also useful during testing
+	// when you don't want state to persist between runs.
+	DropDB bool `envvar:"DROP_DB" default:"false"`
```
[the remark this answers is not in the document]
fabioberger: Yeah, I agree. Let's remove it.

ethereum/blockwatch/block_watcher.go
```diff
@@ -2,18 +2,29 @@ package blockwatch
...
 )
 
+// Number of concurrent JSON RPC requests to allow while fast-syncing logs
+const concurrencyLimit = 3
+
+// maxBlocksInGetLogsQuery is the max number of blocks to fetch logs for in a single query. There is
+// a hard limit of 10,000 logs returned by a single `eth_getLogs` query by Infura's Ethereum nodes so
+// we need to try and stay below it. Parity, Geth and Alchemy all have much higher limits (if any) on
+// the number of logs returned so Infura is by far the limiting factor.
+var maxBlocksInGetLogsQuery = int64(60)
```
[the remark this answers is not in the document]
fabioberger: Ok. I was using int64 because that's the type `big.Int` expects and returns.

ethereum/blockwatch/block_watcher.go
```diff
@@ -270,3 +289,257 @@ func (w *Watcher) addLogs(header *meshdb.MiniHeader) (*meshdb.MiniHeader, error)
...
+	FromBlock int64
+	ToBlock   int64
+}
+
+// getBlockRangeChunks breaks up the block range into chunks of chunkSize. `eth_getLogs`
+// requests are inclusive to both the start and end blocks specified and so we need to
+// make the ranges exclusive of one another to avoid fetching the same blocks' logs twice.
+func (w *Watcher) getBlockRangeChunks(from, to, chunkSize int64) []*blockRange {
+	chunks := []*blockRange{}
+	numBlocksLeft := int64(to - from)
```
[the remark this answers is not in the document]
fabioberger: Fixed.

ethereum/blockwatch/block_watcher.go
```diff
@@ -270,3 +289,257 @@ func (w *Watcher) addLogs(header *meshdb.MiniHeader) (*meshdb.MiniHeader, error)
...
+		}(i, chunk)
+	}
+
+	// Wait for all log requests to complete
+	wg.Wait()
+
+	logs := []types.Log{}
+	for i := range chunks {
+		logRequestResult := indexToLogResult[i]
+		if logRequestResult.Err != nil {
```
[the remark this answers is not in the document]
fabioberger: Yes, you are correct. `i` might fail, and `i+1` might succeed. We intentionally discard the results in `i+1` because we need the contiguous history of events and cannot be missing gaps. Because of this, we only process up until the first error (if one exists), and discard any responses returned beyond that point.

zeroex/order.go
fabioberger: Fixed.

ethereum/blockwatch/block_watcher_test.go
```diff
@@ -67,25 +72,398 @@ func TestWatcher(t *testing.T) {
...
+					FromBlock: 10,
+					ToBlock:   10,
+				},
+			},
+		},
+		blockRangeChunksTestCase{
+			from: 10,
+			to:   16,
+			expectedBlockRanges: []*blockRange{
+				&blockRange{
```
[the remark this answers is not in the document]
fabioberger: Good catch. This was indeed a bug. Fixed.

ethereum/blockwatch/block_watcher_test.go
fabioberger: Although I generally agree, I feel inclined to keep these tests this way. I feel it's much clearer _what_ we are testing by keeping these as simple arithmetic of well-named constants. Much more so then e.g., `"10-71"`

albrow: Fair enough. I'm okay with it.

ethereum/blockwatch/block_watcher_test.go
```diff
@@ -67,25 +73,410 @@ func TestWatcher(t *testing.T) {
...
+	To                        int
+	RangeToFilterLogsResponse map[string]filterLogsResponse
+	Logs                      []types.Log
+	FurthestBlockProcessed    int
+}
+
+func TestGetLogsInBlockRange(t *testing.T) {
+	from := 10
+	to := 20
+	maxBlocksInQuery := maxBlocksInGetLogsQuery
```
albrow: It looks like we don't need this variable and can simply refer to `maxBlocksInGetLogsQuery` directly. (Previously it involved a cast to `int` but that has since been removed).

ethereum/blockwatch/fake_client.go
```diff
@@ -79,7 +79,23 @@ func (fc *fakeClient) HeaderByHash(hash common.Hash) (*meshdb.MiniHeader, error)
 // FilterLogs returns the logs that satisfy the supplied filter query.
 func (fc *fakeClient) FilterLogs(q ethereum.FilterQuery) ([]types.Log, error) {
 	// NOT IMPLEMENTED SINCE NOT USED IN TESTS
```
albrow: I think this comment is outdated?

fabioberger: Yes, fixed.

PR #194, opened 2019-06-27
expirationwatch: Make ExpirationWatcher generic to any common.Hash and expiry

fabioberger:
Prerequisite for: https://github.com/0xProject/0x-mesh/issues/86


expirationwatch/expiration_watcher.go
```diff
@@ -0,0 +1,162 @@
...
+
+// ExpiredItem represents an expired item returned from the Watcher
+type ExpiredItem struct {
+	ExpirationTimeSeconds int64
+	Hash                  common.Hash
+}
+
+// Watcher watches the expiration of items
+type Watcher struct {
+	expiredOrders    chan []ExpiredItem
```
albrow:
```suggestion
	expiredItems    chan []ExpiredItem
```

expirationwatch/expiration_watcher.go
```diff
@@ -0,0 +1,162 @@
...
+func (w *Watcher) prune() []ExpiredItem {
+	pruned := []ExpiredItem{}
+	for {
+		w.rbTreeMu.RLock()
+		key, value := w.rbTree.Min()
+		w.rbTreeMu.RUnlock()
+		if key == nil {
+			break
+		}
+		expirationTimeSeconds := reflect.ValueOf(key).Elem().Int()
```
albrow: The reflect package is mostly for when you don't know types ahead of time, but I don't think we need it here. We know that `key` is of type `*rbt.Int64Key` so you can just do a type cast. I think it would be something like:

```
expirationTimeSeconds := *(*rbt.Int64Key)(key)
```

expirationwatch/expiration_watcher.go
```diff
@@ -0,0 +1,162 @@
...
+	rbTree := rbt.NewRbTree()
+	return &Watcher{
+		expiredOrders:    make(chan []ExpiredItem, 10),
+		rbTree:           rbTree,
+		expirationBuffer: expirationBuffer,
+	}
+}
+
+// Add adds a new item identified by a hash to the expiration watcher
+func (w *Watcher) Add(expirationTimeSeconds int64, hash common.Hash) {
```
albrow: Accepting the time as `int64` is unidiomatic and can lead to bugs/confusion because it is not always clear what unit of time is represented. It also restricts us unnecessarily to second precision (made sense when this package was used only for watching orders, but not when it is more generic). And finally, it doesn't allow us to take advantage of Go's builtin [monotonic clocks](https://golang.org/pkg/time/#hdr-Monotonic_Clocks) (without this, you can see bugs that occur whenever the clock is rolled back or jumps forward).

Can we change this type to `time.Duration`?

expirationwatch/expiration_watcher.go
```diff
@@ -0,0 +1,162 @@
...
+	for {
+		w.rbTreeMu.RLock()
+		key, value := w.rbTree.Min()
+		w.rbTreeMu.RUnlock()
+		if key == nil {
+			break
+		}
+		expirationTimeSeconds := reflect.ValueOf(key).Elem().Int()
+		currentTimePlusBuffer := time.Now().Add(w.expirationBuffer)
+		if expirationTimeSeconds > currentTimePlusBuffer.Unix() {
```
albrow: If we use `time.Time` we should also use `time.After` instead of doing a direct comparison here to take full advantage of monotonic clocks.

expirationwatch/expiration_watcher.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ocdogan/rbt"
+	log "github.com/sirupsen/logrus"
+)
+
+// ExpiredItem represents an expired item returned from the Watcher
+type ExpiredItem struct {
+	ExpirationTimeSeconds int64
+	Hash                  common.Hash
```
albrow: If we want this to be generic, `Hash` should just be `[]byte` or `string` and I would also suggest renaming it to `ID`.

expirationwatch/expiration_watcher.go
fabioberger: These are actually UNIX timestamps passed around as `int64` because they are stored as Big.Ints in the `Order` struct type. We could change this to try `time.Time`, but `time.Duration` doesn't make sense here.

expirationwatch/expiration_watcher.go
fabioberger: Yep, if we do the refactor mentioned above, we could make this change as well.

expirationwatch/expiration_watcher.go
albrow: Yes, sorry, I meant to suggest `time.Time`, not `time.Duration`. The `time` package makes a lot of considerations under the hood (e.g. timezones, monotonic clocks, different units of time) that we often don't get right when trying to do things manually. It's a huge win for preventing bugs and I think it would be a mistake not to use it.

> Perhaps we should refactor the Order type to represent ExpirationTimeSeconds using time.Time?

I don't think we need to do that. It's more convenient to make the `Order` type closely match the JSON-encoding and ABI-encoding for orders in the 0x protocol. However, the expiration watcher is not just for watching orders anymore so I think it should use the more generic `time.Time` type.

That said, you know better than I do how this generic expiration watcher is going to be used in the near future. If you think it's easier/faster to leave it as is, then I'm okay with it.

expirationwatch/expiration_watcher.go
albrow: @fabioberger did you mean to resolve this? The code is still the same.

expirationwatch/expiration_watcher_test.go
```diff
@@ -0,0 +1,160 @@
...
+	watcher.Add(expiryEntryOne.ExpirationTimeSeconds, expiryEntryOne.ID)
+
+	expiryEntryTwo := ExpiredItem{
+		ExpirationTimeSeconds: expiration,
+		ID:                    "0x12ab7edd34515025d0c34aa61a0d1156a631248a4318576a2ce0fb408d3bee521",
+	}
+	watcher.Add(expiryEntryTwo.ExpirationTimeSeconds, expiryEntryTwo.ID)
+
+	pruned := watcher.prune()
+	assert.Len(t, pruned, 2, "Pruned the expired item")
```
albrow: This is a nitpick and you don't have to change it right now, but it would be easier to read test failures if the custom messages passed to `assert` or `require` more closely matched the grammar and structure of the default messages. The default messages describe _why_ the test failed or _how_ the expected result did not match the actual result, not _what_ you were testing. (e.g. "The two words should be the same" from the [docs page](https://godoc.org/github.com/stretchr/testify/assert) or "[1, 2, 3] should have 4 items but had 3" from [`assert.Len`](https://github.com/stretchr/testify/blob/34c6fa2dc70986bccbbffcc6130f6920a924b075/assert/assertions.go#L581)).

So instead of "Pruned the expired item" it should be something like "The expired item should be pruned but it was not". As it is currently written, the error message could be interpreted to mean that the expected behavior was that the item _should not_ be pruned but it actually was.

expirationwatch/expiration_watcher.go
fabioberger: Yes, thanks for pointing this out. I've made this change.

expirationwatch/expiration_watcher.go
fabioberger: Np, refactored to time.Time.

PR #198, opened 2019-07-02
rpc: Implement GetOrders RPC method

### LINKED ISSUE #86: rpc: Add paginated get all orders in DB method

fabioberger:
Fixes #86

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -277,7 +277,7 @@ func (e ErrSnapshotNotFound) Error() string {
 // GetOrders retrieves paginated orders from the Mesh DB at a specific snapshot in time. Passing an empty
 // string as `snapshotID` creates a new snapshot and returns the first set of results. To fetch all orders,
 // continue to make requests supplying the `snapshotID` returned from the first request. After 1 minute of not
-// received further requests referencing a specific snapshot, the snapshot expires and can no longer be used.
+// receiving further requests referencing a specific snapshot, the snapshot expires and can no longer be used.
 func (app *App) GetOrders(page, perPage int, snapshotID string) (*rpc.GetOrdersResponse, error) {
 	ordersInfos := []*zeroex.AcceptedOrderInfo{}
 	if perPage <= 0 {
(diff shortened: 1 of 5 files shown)


core/core.go
```diff
@@ -233,6 +265,86 @@ func (app *App) periodicallyCheckForNewAddrs(startingAddrs []ma.Multiaddr) {
...
+	id string
+}
+
+func (e ErrSnapshotNotFound) Error() string {
+	return fmt.Sprintf("No snapshot found with id: %s. To create a new snapshot, send a request with an empty snapshotID", e.id)
+}
+
+// GetOrders retrieves paginated orders from the Mesh DB at a specific snapshot in time. Passing an empty
+// string as `snapshotID` creates a new snapshot and returns the first set of results. To fetch all orders,
+// continue to make requests supplying the `snapshotID` returned from the first request.
```
albrow: Maybe we should add to this doc comment the fact the snapshots expire after 1 minute of receiving no requests for that snapshot?

core/core.go
```diff
@@ -6,18 +6,23 @@ package core
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
+	dbPkg "github.com/0xProject/0x-mesh/db"
```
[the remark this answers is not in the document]
fabioberger: There was a collision with a variable names `db`, but on second thought, I decided to rename the variable and leave the package name as `db`.

core/core.go
```diff
@@ -233,6 +265,87 @@ func (app *App) periodicallyCheckForNewAddrs(startingAddrs []ma.Multiaddr) {
...
+}
+
+func (e ErrSnapshotNotFound) Error() string {
+	return fmt.Sprintf("No snapshot found with id: %s. To create a new snapshot, send a request with an empty snapshotID", e.id)
+}
+
+// GetOrders retrieves paginated orders from the Mesh DB at a specific snapshot in time. Passing an empty
+// string as `snapshotID` creates a new snapshot and returns the first set of results. To fetch all orders,
+// continue to make requests supplying the `snapshotID` returned from the first request. After 1 minute of not
+// received further requests referencing a specific snapshot, the snapshot expires and can no longer be used.
```
```suggestion
// receiving further requests referencing a specific snapshot, the snapshot expires and can no longer be used.
```

Issue #131, opened 2019-06-03
Validate order schemas


albrow:
AFAICT, we don't currently do schema validations for orders received from peers (i.e., making sure they have the correct structure and all the required fields). We could use JSON Schemas for this. However, we will probably eventually switch to a more efficient encoding over the wire, in which case, some kind of schema validation that matches our encoding format would be preferred.

fabioberger:
@albrow We already attempt unmarshalling the JSON payload into a`SignedOrder` struct.

1. That process will fail if the JSON is malformed
2. If it's proper JSON but has none of the required fields, the resultant `SignedOrder` will contain all the zero-values.
3. `BatchOffchainValidation` already validates that the `assetData` fields include parsable/supported `assetData` values and that the `Signature` includes a properly encoded 0x signature.

Thus the `BatchValidate()` method already reject orders that are malformed prior to any `eth_call`. Now we might want to validate for schema earlier simply to notify the developer that the JSON they are sending is not properly formatted. I'd be down to add it for this reason. Is this the reason you had in mind?

fabioberger:
I'm open to adding support for this. Should we use JSON Schemas for now?

albrow:
> Just to confirm that this is simply to warn developers who are sending an incorrect SignedOrder payload?

I was thinking it would reply to both orders received via RPC and orders received from peers.

> Should we use JSON Schemas for now? Or did you want to punt until we settle on a more efficient wire encoding?

Let's punt for now. If we use Protobufs there is already some built in schema validation if I understand correctly.

fabioberger:
Decided to move forward with JSON schema validation for the Beta.

PR #200, opened 2019-07-03
core/rpc/zeroex: Implement SignedOrder schema validation

fabioberger:
Fixes: #131

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -278,7 +278,7 @@ func (app *App) AddOrders(signedOrdersRaw []*json.RawMessage) (*zeroex.Validatio
 			if err != nil {
 				signedOrder = nil
 			}
-			log.WithField("signedOrderRaw", string(signedOrderBytes)).Info("Unexpected error while attempted to validate signedOrderJSON against schema")
+			log.WithField("signedOrderRaw", string(signedOrderBytes)).Info("Unexpected error while attempting to validate signedOrderJSON against schema")
 			allValidationResults.Rejected = append(allValidationResults.Rejected, &zeroex.RejectedOrderInfo{
 				SignedOrder: signedOrder,
 				Kind:        MeshValidation,
--- a/core/core.go
+++ b/core/core.go
@@ -282,14 +282,19 @@ func (app *App) AddOrders(signedOrdersRaw []*json.RawMessage) (*zeroex.Validatio
 			allValidationResults.Rejected = append(allValidationResults.Rejected, &zeroex.RejectedOrderInfo{
 				SignedOrder: signedOrder,
 				Kind:        MeshValidation,
-				Status:      ROInvalidSchema,
+				Status: zeroex.RejectedOrderStatus{
+					Code:    ROInvalidSchemaCode,
+					Message: "order did not pass JSON-schema validation: Malformed JSON or empty payload",
+				},
 			})
 			continue
 		}
 		if !result.Valid() {
 			log.WithField("signedOrderRaw", string(signedOrderBytes)).Info("Order failed schema validation")
-			status := ROInvalidSchema
-			status.Message = fmt.Sprintf("%s: %s", status.Message, result.Errors())
+			status := zeroex.RejectedOrderStatus{
+				Code:    ROInvalidSchemaCode,
+				Message: fmt.Sprintf("order did not pass JSON-schema validation: %s", result.Errors()),
+			}
 			signedOrder := &zeroex.SignedOrder{}
 			if err := signedOrder.UnmarshalJSON(signedOrderBytes); err != nil {
 				signedOrder = nil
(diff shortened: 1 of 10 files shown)


core/validation.go
```diff
@@ -48,13 +49,58 @@ var (
...
+	}
+	schema, err := sl.Compile(signedOrderSchemaLoader)
+	if err != nil {
+		return nil, err
+	}
+	return schema, nil
+}
+
+func (app *App) schemaValidateOrder(o []byte) (*gojsonschema.Result, error) {
+	orderLoader := gojsonschema.NewStringLoader(string(o))
```
albrow:
```suggestion
	orderLoader := gojsonschema.NewBytesLoader(o)
```

Casting between `[]byte` and `string` types in Go is not free and allocates new memory under the hood. It's best to avoid it when possible. In this case it's really easy because there is a `NewBytesLoader` method.

core/core.go
```diff
@@ -146,6 +153,7 @@ func New(config Config) (*App, error) {
 		orderWatcher:   orderWatcher,
 		ethWathcher:    ethWatcher,
 		orderValidator: orderValidator,
+		jsonSchema:     jsonSchema,
```
```suggestion
		orderJSONSchema:     orderJSONSchema,
```

Isn't this a JSON Schema specifically for orders?

core/validation_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+var (
```
albrow:
```suggestion
const (
```

core/validation_test.go
```diff
@@ -0,0 +1,53 @@
...
+	missingFieldSignedOrderJSON    = `{"takerAddress":"0x0000000000000000000000000000000000000000","makerAssetAmount":"100000000000000000000","takerAssetAmount":"100000000000000000000000","expirationTimeSeconds":"1559856615025","makerFee":"0","takerFee":"0","feeRecipientAddress":"0x0000000000000000000000000000000000000000","senderAddress":"0x0000000000000000000000000000000000000000","salt":"46108882540880341679561755865076495033942060608820537332859096815711589201849","makerAssetData":"0xf47261b0000000000000000000000000e41d2489571d322189246dafa5ebde1f4699f498","takerAssetData":"0xf47261b0000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2","exchangeAddress":"0x4f833a24e1f95d70f028921e27040ca56e09ab0b","signature":"0x1c52f75daa4bd2ad9e6e8a7c35adbd089d709e48ae86463f2abfafa3578747fafc264a04d02fa26227e90476d57bca94e24af32f1cc8da444bba21092ca56cd85603"}`
+)
+
+func TestValidateOrderSchema(t *testing.T) {
+	schema, err := setupSchemaValidator()
+	require.NoError(t, err)
+
+	orderLoader := gojsonschema.NewStringLoader(validSignedOrderJSON)
```
albrow: This would be a good candidate for table driven testing. Something like:

```go
const testCases := []struct{
    input string
    expectedErrs []string
}{
    {
        input: misspelledFieldSignedOrderJSON,
        expectedErrors: []string{
            "makerAddress is required",
            // ...
        }
    }
    // ...
}

for _, testCase := range testCases {
    // For each test case, call Validate and check the results.
}
```

core/validation_test.go
```diff
@@ -0,0 +1,53 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, true, result.Valid())
+
+	orderLoader = gojsonschema.NewStringLoader(misspelledFieldSignedOrderJSON)
+	result, err = schema.Validate(orderLoader)
+	require.NoError(t, err)
+	assert.Equal(t, false, result.Valid())
+	errs := result.Errors()
+	require.Len(t, errs, 2)
+	assert.Equal(t, "makerAddress is required", errs[0].Description())
```
albrow: If we expect 2 errors, we should check both of them instead of just the first one.

zeroex/order.go
```diff
@@ -280,83 +290,128 @@ func (s *SignedOrder) ConvertToOrderWithoutExchangeAddress() wrappers.OrderWitho
...
 	if err != nil {
 		return err
 	}
-	s.MakerAddress = signedOrderJSON.MakerAddress
-	s.MakerAssetData = common.Hex2Bytes(signedOrderJSON.MakerAssetData[2:])
+	if len(signedOrderJSON.MakerAddress) >= addressHexLength && signedOrderJSON.MakerAddress[:2] == "0x" {
+		s.MakerAddress = common.HexToAddress(signedOrderJSON.MakerAddress)
+	}
+	if len(signedOrderJSON.MakerAssetData) >= 2 && signedOrderJSON.MakerAssetData[:2] == "0x" {
+		s.MakerAssetData = common.Hex2Bytes(signedOrderJSON.MakerAssetData[2:])
```
albrow: The [`common.FromHex` function](https://github.com/ethereum/go-ethereum/blob/7fd82a0e3e87aaf26198a6cf2de242c30fa7c2ab/common/bytes.go#L45) already checks for the "0x" prefix if you just want to use that instead.

core/message_handler.go
```diff
@@ -83,6 +84,25 @@ func (app *App) HandleMessages(messages []*p2p.Message) error {
...
 			continue
 		}
+
+		result, err := app.schemaValidateOrder(msg.Data)
+		if err != nil {
+			log.WithFields(map[string]interface{}{
+				"error": err,
+				"from":  msg.From,
+			}).Trace("could not schema validate message")
+			app.handlePeerScoreEvent(msg.From, psInvalidMessage)
```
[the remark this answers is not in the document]
fabioberger: If the sender sent a message containing malformed JSON or nothing at all, `schemaValidateOrder` returns an error. I do think we should penalize the sender in both of these cases.

core/core.go
fabioberger: Yep, renamed.

zeroex/order.go
fabioberger: Hm, it's not as strict as I would like. I don't think we should allow clients to send `MakerAssetData` without the `0x` prefix, or with a `0X` prefix. Let's leave it as is for now.

albrow: > I don't think we should allow clients to send MakerAssetData without the 0x prefix

Shouldn't that be part of the JSON Schema? Or is it not possible to enforce that constraint?

albrow: I just checked and it looks like it is already enforced by the JSON schema.

Issue #191, opened 2019-06-26
zeroex: Can 500 orders be reliably validated in a single RPC call?


fabioberger:
Bumping up my SRA scraping script to send 500 orders at a time to Mesh caused the call to `orderValidationUtils.GetOrderRelevantStates` to sometimes return `request failed or timed out`. We should investigate this failure and potentially use a similar `split payload on error` approach as https://github.com/0xProject/0x-mesh/pull/189.

fabioberger:
Further investigation:

**Note:** All orders used were non-multiAsset orders. The size of orders is variable, so these values are meant as approximations, not absolutes.

Found a [WebSocket payload limit of 32MB](https://github.com/golang/net/commit/6dba816f1056709e29a1c442883cab1336d3c083) which is around ~510 non-multiAsset orders.

Calling `orderValidationUtils.GetOrderRelevantStates` via different Ethereum node implementations: 

- Infura: "request failed or timed out" error if sending > ~325 orders
- Alchemy: Tested up to 840 orders without any issue
- Parity: Tested up to 876 orders without issue. Request did take longer, but nowhere near the request timeout (currently 30sec)
- Geth: "413 Request Entity Too Large content length too large" error if sending > ~325 orders

PR #207, opened 2019-07-07
core/zeroex: Batch validate orders based on payload size, not number of orders

fabioberger:
Fixes: #191 

Currently, we simply batching 300 orders into each batch validation RPC call. There are two issues with this:
1. Some Ethereum clients have limits on the `Content-Length` of the JSON RPC payloads they will accept. Depending on which one you use, you might want to enforce a different max `Content-Length` (i.e., batch validate differing amounts of orders).
2. 0x orders aren't all a uniform size. Namely, orders involving `multiAssetData` can have any arbitrary size. As such, 300 orders will not always map to the same request `Content-Length`. Most orders do not involve `multiAssetData` and are of a uniform size.

Because of these two issues, this PR batches orders based on a desired max `Content-Length` rather then a specific number of orders. This `EthereumRPCMaxContentLength` is configurable via an env config and defaults to the max `Content-Length` supported by Geth & Infura. The algorithm works as follows:

1. We compute the number of "basic" orders we expect to fit into `EthereumRPCMaxContentLength` number of bytes, where "basic" orders are orders without `multiAssetData`. 
2. We check if encoding that number of orders into a JSON RPC payload results in a `Content-Length` less than the `EthereumRPCMaxContentLength`. If yes, we create a batch or orders of that size. If not, we use a limited-depth binary-search algorithm to figure out how many orders can fit into the max `Content-Length`.

This implementation allows Mesh operators using Parity / Alchemy as their backing Ethereum node to send larger batches of orders for validation. It also handles any arbitrary sized orders and avoids overstepping the max content length settings of the backing Ethereum node.


zeroex/order_validator.go
```diff
@@ -654,6 +685,108 @@ func (o *OrderValidator) isSupportedAssetData(assetData []byte) bool {
 	return true
 }
 
+// findChunkSize uses a limited-depth binary-search to find a chunk size that is less then the
+// max content length
+func (o *OrderValidator) findChunkSize(signedOrders []*SignedOrder) int64 {
```
albrow: Unless I'm mistaken, it looks like it is possible that the binary search will never reach `hi` with the way it is currently implemented, even in cases where all the given `signedOrders` could be included in the payload.

For example, consider a case where we have 500 orders and all of them could theoretically be included in the payload without hitting the size limit. The for loop will run 7 times, starting with `lo = 0`, `hi = 500`. On each iteration it will increase `mid`.

0. (0+500) / 2 = 250
1. (250+500) / 2 = 375
2. (375+500) / 2 = 428
3. (428+500) / 2 = 464
4. (464+500) / 2 = 482
5. (482+500) / 2 = 491
6. (491+500) / 2 = 495
7. (495+500) / 2 = 497

In fact, due to rounding, it is actually not possible to hit `max` even with infinite iterations in this case. I would expect that cases where all the given orders could fit within size constraints are fairly common. Maybe we can adjust the algorithm to check if `hi` fits within the size constraints directly?

zeroex/order_validator.go
```diff
@@ -654,6 +685,108 @@ func (o *OrderValidator) isSupportedAssetData(assetData []byte) bool {
...
+	// Determine the lowest and highest possible signedOrders limits to binary search in between
+	lo := int64(0)
+	hi := int64(len(signedOrders))
+	min := int64(1)
+
+	// Execute the binary search and hone in on an executable payload size of signedOrders
+	maxBinarySearchDepth := 7
+	depth := 0
+	for lo+1 < hi {
```
albrow: Optionally, you could write the for loop this way:

```suggestion
	for depth := 0; lo+1 < hi && depth <= maxBinarySearchDepth; depth++ {
```

zeroex/order_validator.go
```diff
@@ -654,6 +685,108 @@ func (o *OrderValidator) isSupportedAssetData(assetData []byte) bool {
...
+// findChunkSize uses a limited-depth binary-search to find a chunk size that is less then the
+// max content length
+func (o *OrderValidator) findChunkSize(signedOrders []*SignedOrder) int64 {
+	// Determine the lowest and highest possible signedOrders limits to binary search in between
+	lo := int64(0)
+	hi := int64(len(signedOrders))
+	min := int64(1)
+
+	// Execute the binary search and hone in on an executable payload size of signedOrders
+	maxBinarySearchDepth := 7
```
albrow: This is a constant that could be declared outside of the function scope.

zeroex/order_validator_test.go
```diff
@@ -251,6 +252,154 @@ func TestBatchValidateCoordinatorSoftCancels(t *testing.T) {
...
+
+	expectedChunkSize := int64(2)
+	chunkSize := orderValidator.findChunkSize(signedOrders)
+	require.NoError(t, err)
+	assert.Equal(t, expectedChunkSize, chunkSize)
+}
+
+// TODO: Catch throw?
+func TestFindChunkSizeFails(t *testing.T) {
+	defer func() {
```
albrow: You can use [`assert.Panics`](https://godoc.org/github.com/stretchr/testify/assert#Panics) or `assert.PanicsWithValue` here.

zeroex/order_validator_test.go
```diff
@@ -251,6 +252,154 @@ func TestBatchValidateCoordinatorSoftCancels(t *testing.T) {
...
+	require.NoError(t, err)
+
+	signedOrders := []*SignedOrder{}
+	for i := 0; i < 5; i++ {
+		signedOrders = append(signedOrders, signedOrder)
+	}
+
+	expectedChunkSize := int64(0)
+	chunkSize := orderValidator.findChunkSize(signedOrders)
+	assert.Equal(t, expectedChunkSize, chunkSize)
```
albrow: This assertion will never actually run if `findChunkSize` panics. You can just remove it.

zeroex/order_validator.go
```diff
@@ -654,6 +685,108 @@ func (o *OrderValidator) isSupportedAssetData(assetData []byte) bool {
...
+const abiHeadByteLen = int64(64)
+
+// abiArrayLenByteLen is the number of bytes taken up to describe the length of an encoded array
+// Source: https://solidity.readthedocs.io/en/v0.5.10/abi-spec.html#use-of-dynamic-types
+const abiArrayLenByteLen = int64(64)
+
+// computeNumBasicOrdersEncodable calculates the number of "basic" (e.g., non-multiAsset) 0x orders
+// that can be sent in a payload of max content size. Unlike "basic" orders, those involving the
+// MultiAssetProxy can be of arbitrary size.
+func (o *OrderValidator) computeNumBasicOrdersEncodable() int64 {
```
albrow: It looks like all of the things we are encoding here are known ahead of time. Can we replace this method with a constant? (e.g., `const numBasicOrdersEncodable = whatever the result is`)

zeroex/order_validator.go
```diff
@@ -654,6 +685,108 @@ func (o *OrderValidator) isSupportedAssetData(assetData []byte) bool {
...
+
+	payloadSize, err := o.computePayloadSize(getOrderRelevantStatesMethodName, orders, signatures)
+	if err != nil {
+		return false
+	}
+	return payloadSize < o.maxRequestContentLength
+}
+
+func (o *OrderValidator) computePayloadSize(method string, params ...interface{}) (int64, error) {
+	data, err := o.orderValidationUtilsABI.Pack(method, params...)
```
albrow: This is probably okay for now but some of what we're encoding here is known ahead of time. For example, we know the rest of the JSON RPC payload will always be the same and only the size of the `data` field will change.

constants/constants.go
```diff
@@ -47,3 +47,6 @@ var GanacheAccountToPrivateKey = map[common.Address][]byte{
 
 // ErrInternal is used whenever we don't wish to expose internal errors to a client
 var ErrInternal = errors.New("internal error")
+
+// TestMaxContentLength is the max Ethereum RPC Content-Length used in tests
+var TestMaxContentLength = int64(1024 * 512)
```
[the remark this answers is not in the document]
fabioberger: Yep, my bad. Was thinking it terms of Solidity where `int` is `int8` 🤦‍♂

zeroex/order_validator.go
fabioberger: The one item that is not known ahead of time is the `maxRequestContentLength` configured by the operator. We could make the rest constant but I feel it's less of a black-box to spell out all the components involved.

zeroex/order_validator.go
fabioberger: That's true. I like keeping it because it's more straightforward to refactor if the form/fields of the payload ever change, whereas if we just had a constant number of bytes here, it wouldn't be as clear how it should be changed. Open to getting rid of this though if you feel strongly.

fabioberger: Agreed. Replaced with a constant.

core/core.go
```diff
@@ -124,38 +135,39 @@ func New(config Config) (*App, error) {
...
+	orderWatcher, err := orderwatch.New(db, blockWatcher, orderValidator, config.EthereumNetworkID, config.OrderExpirationBuffer)
 	if err != nil {
 		return nil, err
 	}
-	// TODO(albrow): Call Add for all existing makers/signers in the database.
 
-	// Initialize the order validator
-	orderValidator, err := zeroex.NewOrderValidator(ethClient, config.EthereumNetworkID)
+	// Initialize the ETH balance watcher (but don't start it yet).
+	ethWatcher, err := ethereum.NewETHWatcher(ethWatcherPollingInterval, ethClient, config.EthereumNetworkID)
```
fabioberger: Annoying that the diff is messing up here. I literally just move the `OrderValidator` instantiation above the `OrderWatcher` instantiation since we now pass it into the `OrderWatcher` instead of re-instatiating it within it.

albrow: 🤷‍♂thanks for the heads up.

zeroex/order_validator.go
```diff
@@ -654,6 +660,83 @@ func (o *OrderValidator) isSupportedAssetData(assetData []byte) bool {
...
+	}
+	encodedSignedOrderByteLength := len(encodedData) - emptyGetOrderRelevantStatesCallDataByteLength
+	return encodedSignedOrderByteLength, nil
+}
+
+// computeOptimalChunkSizes splits the signedOrders into chunks where the payload size of each chunk
+// is beneath the maxRequestContentLength. It does this by implementing a greedy algorithm which ABI
+// encodes signedOrders one at a time until the computed payload size is as close to the
+// maxRequestContentLength as possible.
+func (o *OrderValidator) computeOptimalChunkSizes(signedOrders []*SignedOrder) []int {
```
albrow: 🎉Really like the new implementation.

core/core.go
```diff
@@ -64,18 +64,25 @@ type Config struct {
 	// networks have different block producing intervals: POW networks are typically slower (e.g., Mainnet)
 	// and POA networks faster (e.g., Kovan) so one should adjust the polling interval accordingly.
 	BlockPollingInterval time.Duration `envvar:"BLOCK_POLLING_INTERVAL" default:"5s"`
+	// EthereumRPCMaxContentLength is the maximum request Content-Length accepted by the backing Ethereum RPC
+	// endpoint used by Mesh. Geth & Infura both limit a request's content length to 1024 * 512 Bytes. Parity
+	// and Alchemy have much higher limits. When batch validating 0x orders, we will fit as many orders into a
+	// request without crossing the max content length. The default value is appropriate for operators using Geth
+	// or Infura. If using Alchemy or Parity, feel free to double the default max in order to reduce the
+	// number of RPC calls made by Mesh.
+	EthereumRPCMaxContentLength int `envvar:"ETHEREUM_RPC_MAX_CONTENT_LENGTH" default:"524288"`
```
albrow: Let's add this new env var to __DOCUMENTATION.md__ (or __DEPLOYMENT.md__ if #218 is merged first).

PR #213, opened 2019-07-08
Consolidate database and private key paths into a single env var

albrow:
Taking a page [from Geth's playbook](https://github.com/ethereum/go-ethereum/wiki/Command-Line-Options), I think it makes more sense to move the database file and private key file into a single directory called `DATA_DIR`. This reduces the number of env vars required and makes it easier to reset Mesh/remove all data associated with it.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -9,6 +9,7 @@ Some resources:
 - [Github repo](https://github.com/0xProject/0x-mesh/)
 
 ### Supported networks
+
 - Mainnet
 - Kovan
 - Ropsten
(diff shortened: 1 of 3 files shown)


DOCUMENTATION.md
```diff
@@ -60,8 +62,9 @@ All possible env vars are detailed in the [Config](https://godoc.org/github.com/
 type Config struct {
     // Verbosity is the logging verbosity: 0=panic, 1=fatal, 2=error, 3=warn, 4=info, 5=debug 6=trace
     Verbosity int `envvar:"VERBOSITY" default:"2"`
-    // DatabaseDir is the directory to use for persisting the database.
-    DatabaseDir string `envvar:"DATABASE_DIR" default:"./0x_mesh/db"`
+	// DataDir is the directory to use for persisting all data, including the
+	// database and private key files.
+	DataDir string `envvar:"DATA_DIR" default:"0x_mesh"`
```
fabioberger: Is the alignment off here or is it just the diff that makes it look like that?

albrow: It's off. Fixed!

PR #216, opened 2019-07-09
Fix bug when unmarshaling OrderEvent

albrow:
Without this fix, attempting to subscribe to orders will result in an error:

```
RPC_ADDRESS="ws://localhost:60147" ETHEREUM_RPC_URL="http://localhost:8545" go run cmd/demo/subscribe_to_orders/main.go
panic: math/big: cannot unmarshal "\"2000\"" into a *big.Int

goroutine 1 [running]:
main.main()
	/Users/alex/programming/go/src/github.com/0xProject/0x-mesh/cmd/demo/subscribe_to_orders/main.go:46 +0x53e
exit status 2
```

This is happening because we implemented a custom `MarshalJSON` method for `OrderEvent` but never implemented the corresponding `UnmarshalJSON` method. I also added a test in __order_test.go__ to cover this case.

### DIFF
--- a/zeroex/order.go
+++ b/zeroex/order.go
@@ -81,15 +81,42 @@ type OrderEvent struct {
 	TxHash common.Hash `json:"txHash"`
 }
 
+type orderEventJSON struct {
+	OrderHash                string         `json:"orderHash"`
+	SignedOrder              *SignedOrder   `json:"signedOrder"`
+	Kind                     OrderEventKind `json:"kind"`
+	FillableTakerAssetAmount string         `json:"fillableTakerAssetAmount"`
+	TxHash                   string         `json:"txHash"`
+}
+
 // MarshalJSON implements a custom JSON marshaller for the SignedOrder type
-func (o *OrderEvent) MarshalJSON() ([]byte, error) {
-	return json.Marshal(map[string]interface{}{
-		"orderHash":                o.OrderHash.Hex(),
-		"signedOrder":              o.SignedOrder,
-		"kind":                     o.Kind,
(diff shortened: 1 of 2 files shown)


zeroex/order.go
```diff
@@ -298,7 +325,7 @@ type signedOrderJSON struct {
 }
 
 // MarshalJSON implements a custom JSON marshaller for the SignedOrder type
-func (s *SignedOrder) MarshalJSON() ([]byte, error) {
+func (s SignedOrder) MarshalJSON() ([]byte, error) {
```
albrow: Non-pointer receiver types are safer for `MarshalJSON`. Without this, passing a `SignedOrder` (not a pointer) to `json.Marshal` will happily work but it will *not* use our custom method.

zeroex/order_test.go
```diff
@@ -11,54 +13,57 @@ import (
...
-	}
+var testOrder = &Order{
+	MakerAddress:          constants.GanacheAccount0,
+	TakerAddress:          constants.NullAddress,
+	SenderAddress:         constants.NullAddress,
+	FeeRecipientAddress:   constants.NullAddress,
+	MakerAssetData:        constants.NullAddress.Bytes(),
+	TakerAssetData:        constants.NullAddress.Bytes(),
+	ExchangeAddress:       fakeExchangeContractAddress,
+	Salt:                  big.NewInt(200),
```
albrow: I changed these numbers just in case. It's hypothetically easier to get a false positive when testing using zeroes for all the numbers.

PR #217, opened 2019-07-09
Update Dockerfile

albrow:


Dockerfile
```diff
@@ -3,14 +3,12 @@ FROM golang:1.12.1-alpine3.9 as mesh-builder
 
 
 RUN apk update && apk add ca-certificates nodejs-current npm make git dep gcc build-base musl linux-headers
-RUN npm install -g yarn
```
albrow: We only need yarn and JavaScript dependencies for running the tests in a Node.js/Wasm environment. We don't need it here so I removed this step and also added a new `make deps-go-no-lockfile` so we can skip installing JS dependencies.

Issue #227, opened 2019-07-10
orderwatch: cleanup job crashes with db error


fabioberger:
Seeing this failure after running Mesh for 5hrs:

```
{"error":"leveldb: not found","lastUpdatedCutOff":"2019-07-10T18:15:44.597809441Z","level":"panic","msg":"Failed to find orders by LastUpdatedBefore","time":"2019-07-10T18:45:45Z"}
panic: (*logrus.Entry) (0xe8ce40,0xc009d250a0)

goroutine 52 [running]:
github.com/0xProject/0x-mesh/vendor/github.com/sirupsen/logrus.Entry.log(0xc000070180, 0xc0009d3920, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/0xProject/0x-mesh/vendor/github.com/sirupsen/logrus/entry.go:239 +0x2ce
github.com/0xProject/0x-mesh/vendor/github.com/sirupsen/logrus.(*Entry).Log(0xc009d25030, 0xc000000000, 0xc0000b7ce0, 0x1, 0x1)
	/go/src/github.com/0xProject/0x-mesh/vendor/github.com/sirupsen/logrus/entry.go:268 +0xed
github.com/0xProject/0x-mesh/vendor/github.com/sirupsen/logrus.(*Entry).Panic(0xc009d25030, 0xc0000b7ce0, 0x1, 0x1)
	/go/src/github.com/0xProject/0x-mesh/vendor/github.com/sirupsen/logrus/entry.go:306 +0x55
github.com/0xProject/0x-mesh/zeroex/orderwatch.(*Watcher).startCleanupWorker.func1(0xc00030dba0)
	/go/src/github.com/0xProject/0x-mesh/zeroex/orderwatch/order_watcher.go:223 +0x4f9
created by github.com/0xProject/0x-mesh/zeroex/orderwatch.(*Watcher).startCleanupWorker
	/go/src/github.com/0xProject/0x-mesh/zeroex/orderwatch/order_watcher.go:200 +0x3f
```

PR #228, opened 2019-07-11
db: Add more detail to error when index looks outdated

albrow:
This might help us track down the root cause of #https://github.com/0xProject/0x-mesh/issues/227. 

The db package is supposed to maintain consistency between indexes and model data, even in the case of crashes or failed writes. It does this by using transactions under the hood for inserts, updates, and deletions. If the error is really occurring here it means something went seriously wrong and we lost consistency between the indexes and model data.


albrow:
We decided we're going to handle this case differently. New PR coming soon.

Issue #223, opened 2019-07-10
mesh: Add bootstrap mode


fabioberger:

albrow:
For clarification, "bootstrap mode" will mean the node is solely responsible for peer discovery and helping other peers find each other on the network. It won't store or validate any orders. This is useful because otherwise the bootstrap node could become resource constrained and peer discovery would be hampered.

PR #230, opened 2019-07-11
Update the bootstrap list and use it in mesh-bootstrap

fabioberger:
Fixes: #223 

This PR updates the bootstrap list to include our own dedicated bootstrap nodes.


albrow:
@fabioberger I hijacked this PR and added the other features I needed to it. We now make it so that each bootstrap node maintains an active connection to the other bootstrap nodes.

fabioberger:
LGTM

PR #235, opened 2019-07-12
Create a separate executable for bootstrap nodes

### LINKED ISSUE #223: mesh: Add bootstrap mode

albrow:
Fixes #223.

After investigating this, I decided it was cleaner to create a separate executable rather than modify the existing code at __p2p/node.go__. Technically there is some duplicated code but a lot of the configuration is different and we are able to skip almost all the steps in `p2p.Node.Start`.

### DIFF
--- a/cmd/mesh-bootstrap/main.go
+++ b/cmd/mesh-bootstrap/main.go
@@ -1,0 +1,158 @@
+// +build !js
+
+// mesh-bootstrap is a separate executable for bootstrap nodes. Bootstrap nodes
+// will not share or receive any orders and its sole responsibility is to
+// facilitate peer discovery.
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/0xProject/0x-mesh/keys"
+	libp2p "github.com/libp2p/go-libp2p"
+	connmgr "github.com/libp2p/go-libp2p-connmgr"
+	p2pcrypto "github.com/libp2p/go-libp2p-crypto"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	p2pnet "github.com/libp2p/go-libp2p-net"
+	ma "github.com/multiformats/go-multiaddr"
+	"github.com/plaid/go-envvar/envvar"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// peerCountLow is the target number of peers to connect to at any given time.
+	peerCountLow = 1000
+	// peerCountHigh is the maximum number of peers to be connected to. If the
+	// number of connections exceeds this number, we will prune connections until
+	// we reach peerCountLow.
+	peerCountHigh = 1100
+	// peerGraceDuration is the amount of time a newly opened connection is given
+	// before it becomes subject to pruning.
+	peerGraceDuration = 10 * time.Second
+	// defaultNetworkTimeout is the default timeout for network requests (e.g.
+	// connecting to a new peer).
+	defaultNetworkTimeout = 30 * time.Second
+)
+
+// Config contains configuration options for a Node.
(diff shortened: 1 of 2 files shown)


cmd/mesh-bootstrap/main.go
```diff
@@ -0,0 +1,158 @@
...
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	p2pnet "github.com/libp2p/go-libp2p-net"
+	ma "github.com/multiformats/go-multiaddr"
+	"github.com/plaid/go-envvar/envvar"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// peerCountLow is the target number of peers to connect to at any given time.
+	peerCountLow = 1000
```
albrow: Bootstrap nodes have a higher peer count and a longer default network timeout.

cmd/mesh-bootstrap/main.go
```diff
@@ -0,0 +1,158 @@
...
+	basicHost, err := libp2p.New(ctx, opts...)
+	if err != nil {
+		log.WithField("error", err).Fatal("could not create host")
+	}
+
+	// Set up the notifee.
+	basicHost.Network().Notify(&notifee{})
+
+	// Set up DHT for peer discovery.
+	kadDHT, err := dht.New(ctx, basicHost)
```
albrow: Bootstrap nodes do not use any rendezvous points. All they do is initialize the DHT and follow the `/ipfs/kad/1.0.0` protocol. In effect this allows bootstrap nodes to support peer discovery for all Ethereum network ids simultaneously and also combines our DHT for all networks. It's not 100% clear if this is desirable but it does have some positive effects in terms of increasing robustness and redundancy of the network. It is still the case that peers who are using different Ethereum network IDs will not share messages with one another because they use different pubsub topics. This just means that they will temporarily connect to one another for the purposes of peer discovery.

cmd/mesh-bootstrap/main.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/0xProject/0x-mesh/keys"
+	libp2p "github.com/libp2p/go-libp2p"
+	connmgr "github.com/libp2p/go-libp2p-connmgr"
+	p2pcrypto "github.com/libp2p/go-libp2p-crypto"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	p2pnet "github.com/libp2p/go-libp2p-net"
```
raulk: you should use `go-libp2p-core/network` here ;-)

cmd/mesh-bootstrap/main.go
```diff
@@ -0,0 +1,158 @@
...
+	// Set up the notifee.
+	basicHost.Network().Notify(&notifee{})
+
+	// Set up DHT for peer discovery.
+	kadDHT, err := dht.New(ctx, basicHost)
+	if err != nil {
+		log.WithField("error", err).Fatal("could not create DHT")
+	}
+
+	if err := kadDHT.Bootstrap(ctx); err != nil {
```
raulk: Since we have no nodes to bootstrap against, the routing tables of these peers will remain empty until the first peers connect to them -- which I guess it's the behaviour you're looking for. Just wanted to point it out.

albrow:
Going to go ahead and merge this for now. An upcoming PR will make the bootstrap nodes connect to one another and tag them via the connection manager so they stay connected. Thanks for the suggestion, @raulk!

cmd/mesh-bootstrap/main.go
albrow: We're planning to update all of our libp2p libraries in one fell swoop. Likely next week.

PR #242, opened 2019-07-12
Temporarily don't accept orders with a senderAddress

albrow:
The comment in the code explains why we want to do this. We plan to add support for some known sender addresses out-of-the-box and will eventually allow users to pass in custom validation logic (which can include additional third-party sender addresses).

### DIFF
--- a/core/validation.go
+++ b/core/validation.go
@@ -162,7 +162,7 @@ func (app *App) validateOrders(orders []*zeroex.SignedOrder) (*zeroex.Validation
 			continue
 		}
 		// Note(albrow): Orders with a sender address can be canceled or invalidated
-		// off-chain which is difficult to suppport since we need to prune
+		// off-chain which is difficult to support since we need to prune
 		// canceled/invalidated orders from the database. We can special-case some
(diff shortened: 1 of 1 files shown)


core/validation.go
```diff
@@ -156,6 +161,21 @@ func (app *App) validateOrders(orders []*zeroex.SignedOrder) (*zeroex.Validation
 			})
 			continue
 		}
+		// Note(albrow): Orders with a sender address can be canceled or invalidated
+		// off-chain which is difficult to suppport since we need to purne
```
fabioberger:
```suggestion
		// off-chain which is difficult to suppport since we need to prune
```

core/validation.go
```diff
@@ -156,6 +161,21 @@ func (app *App) validateOrders(orders []*zeroex.SignedOrder) (*zeroex.Validation
...
+		// canceled/invalidated orders from the database. We can special case some
```
fabioberger:
```suggestion
		// canceled/invalidated orders from the database. We can special-case some
```

Issue #256, opened 2019-07-15
mesh: mesh_getOrders fillableTakerAssetAmount is a number


dekz:
Looks like the return data from the `mesh_getOrders` contains a `fillableTakerAssetAmount` encoded as a number rather than a string.

#178
#179

PR #259, opened 2019-07-15
zeroex: Marshal AcceptedOrderInfo properly

fabioberger:
This PR makes sure to Marshal the `fillableTakerAssetAmount` field in AcceptedOrderInfo into a string.

Fixes: #256


albrow:
@fabioberger I think you need to add an `UnmarshalJSON` method like in https://github.com/0xProject/0x-mesh/pull/216.

PR #269, opened 2019-07-17
Introduce a temporary hack to discover public IP address

albrow:
As a side-effect this change makes `P2P_LISTEN_PORT` required.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"time"
 
 	libp2p "github.com/libp2p/go-libp2p"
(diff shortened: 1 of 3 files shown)


p2p/node.go
```diff
@@ -107,6 +108,22 @@ func New(config Config) (*Node, error) {
...
 	}
 
+	// HACK(albrow): As a workaround for AutoNAT issues, ping ifconfig.me to
+	// determine our public IP address on boot. This will work for nodes that
+	// would be reachable via a public IP address but don't know what it is (e.g.
+	// because they are running in a Docker container).
+	publicIP, err := getPublicIP()
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("could not get public IP address: %s", err.Error())
```
fabioberger: If a node doesn't have a public IP address, shouldn't it still boot up and try to use AutoRelay?

albrow: ifconfig.me will always return an IP address. It just might not necessarily be an IP address that you are actually publicly reachable at.

fabioberger: Gotcha.

PR #275, opened 2019-07-17
Add type suffixes to all keys when logging

albrow:
This PR will help fix field mapping conflicts in Elasticsearch. When we have two separate logs with the same field name but different types, Elasticsearch cannot correctly index them, which often results in log messages being lost. If we add a type suffix, it will no longer be possible to have the same field name but different types.

### DIFF
--- a/loghooks/key_suffix_hook.go
+++ b/loghooks/key_suffix_hook.go
@@ -1,0 +1,92 @@
+package loghooks
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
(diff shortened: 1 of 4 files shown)


loghooks/key_suffix_hook.go
```diff
@@ -0,0 +1,116 @@
...
+		delete(entry.Data, key)
+		entry.Data[newKey] = value
+	}
+	return nil
+}
+
+// getTypeForValue returns a string representation of the type of the given val.
+func getTypeForValue(val interface{}) (string, error) {
+	if _, ok := val.(json.Marshaler); ok {
+		// If val implements json.Marhsler, return the type of json.Marshal(val)
```
fabioberger:
```suggestion
		// If val implements json.Marshaler, return the type of json.Marshal(val)
```

PR #276, opened 2019-07-18
Reduce AdvertiseBootDelay for bootstrap nodes

albrow:


cmd/mesh-bootstrap/main.go
```diff
@@ -59,6 +60,13 @@ type Config struct {
 	PrivateKeyPath string `envvar:"PRIVATE_KEY_PATH" default:"0x_mesh/keys/privkey"`
 }
 
+func init() {
+	// Since we know that the bootstrap nodes are more stable, we can
+	// safely reduce AdvertiseBootDelay. This will allow the bootstrap nodes to
+	// advertise themselves as relays sooner.
+	relay.AdvertiseBootDelay = 30 * time.Second
+}
+
```
fabioberger: Where is `init` called?

albrow: [`init`](https://golang.org/doc/effective_go.html#init) is called automatically after any global variables are initialized but before `main`.

Issue #278, opened 2019-07-18
fatal error: concurrent map read and map write


dekz:
After running mesh for an extended period I encountered the following crash:

```
{"direction":1,"level":"debug","msg":"found peer who speaks our protocol","peerID":"QmVGX47BzePPqEzpkTwfUJogPZxHcifpSXsGdgyHjtk5t7","protocol":"/meshsub/1.0.0","time":"2019-07-17T10:25:10Z"}
{"count":6,"level":"debug","msg":"received AddOrders request via RPC","time":"2019-07-17T10:25:10Z"}
{"count":6,"level":"debug","msg":"received AddOrders request via RPC","time":"2019-07-17T10:25:13Z"}
{"count":6,"level":"debug","msg":"received AddOrders request via RPC","time":"2019-07-17T10:25:16Z"}
{"count":6,"level":"debug","msg":"received AddOrders request via RPC","time":"2019-07-17T10:25:19Z"}
fatal error: concurrent map read and map write

goroutine 64 [running]:
runtime.throw(0xeb74c4, 0x21)
    /usr/local/go/src/runtime/panic.go:617 +0x72 fp=0xc0013e50d8 sp=0xc0013e50a8 pc=0x433a42
runtime.mapaccess2(0xd563e0, 0xc0001b43c0, 0xc0013e5228, 0x199dfe0, 0xc001fdc600)
    /usr/local/go/src/runtime/map.go:472 +0x284 fp=0xc0013e5120 sp=0xc0013e50d8 pc=0x413f34
github.com/0xProject/0x-mesh/zeroex/orderwatch.(*Decoder).FindEventType(0xc0000f6000, 0xc10d63adca56bc67, 0xe9c8ad40a5149b71, 0xc3258396, 0xc000c6ba80, 0x3, 0x4, 0xc001fddb20, 0x20, 0x20, ...)
    /go/src/github.com/0xProject/0x-mesh/zeroex/orderwatch/event_decoder.go:253 +0x570 fp=0xc0013e5220 sp=0xc0013e5120 pc=0xc15340
github.com/0xProject/0x-mesh/zeroex/orderwatch.(*Watcher).setupEventWatcher.func1(0xc000358820, 0xc000421320)
    /go/src/github.com/0xProject/0x-mesh/zeroex/orderwatch/order_watcher.go:305 +0x133 fp=0xc0013e5fd0 sp=0xc0013e5220 pc=0xc1dbe3
runtime.goexit()
[...]

PR #279, opened 2019-07-18
orderwatch: Fix concurrent read/write issue in EventDecoder

fabioberger:
Fixes: #278 

Protect all read/writes to known contract addresses with a mutex so that they are concurrency-safe.

### DIFF
--- a/zeroex/orderwatch/event_decoder.go
+++ b/zeroex/orderwatch/event_decoder.go
@@ -6,6 +6,7 @@ import (
 	"math/big"
 	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


zeroex/orderwatch/event_decoder.go
```diff
@@ -194,46 +198,82 @@ func NewDecoder() (*Decoder, error) {
...
 // events for this token.
 func (d *Decoder) RemoveKnownERC20(address common.Address) {
+	d.knownERC20AddressesMu.Lock()
+	defer d.knownERC20AddressesMu.Unlock()
 	delete(d.knownERC20Addresses, address)
 }
 
+// isKnownERC20 checks if the supplied address is a known ERC20 contract
+func (d *Decoder) isKnownERC20(address common.Address) bool {
+	d.knownERC20AddressesMu.Lock()
```
albrow: `isKnownERC20` and other similar methods should use `RLock` and `RUnlock` since they are reading from the map but not writing to it.

albrow:
👍 LGTM

Issue #96, opened 2019-05-28
Make asynchronous APIs more consistent


albrow:
When we first started building 0x Mesh, we weren't certain about the best practices for writing asynchronous APIs in Go and had to consider a wide variety of unexpected use cases and constraints. At this point, enough of the pieces are in place that we should be able to go back and choose a general set of guidelines to apply to all of our asynchronous code. 

Here are some general rules I'm proposing:

1. Avoid exporting non-blocking functions/methods. Instead, mimic methods from the standard library like [`http.ListenAndServe`](https://golang.org/pkg/net/http/#ListenAndServe), which blocks until there is an error. Callers can wrap this in a goroutine if needed.
2. Don't expose channels in exported functions, methods, or fields.
3. Whenever we do continuous computation in a for loop (i.e. a for loop without a post statement) we should check for [`context.Done`](https://golang.org/pkg/context/#Context). This usually necessitates accepting a `context.Context` as an argument.
4. Don't use `context.Background` except in tests and __main.go__.
5. Don't use a `Close` method. Just cancel the context.
6. Don't pass in timeouts as a `time.Duration`. Just use a context and expect the caller to use `context.WithTimeout` if they need timeouts.

There are of course going to be some exceptions (e.g. `go-ethereum/rpc` requires us to have an exported method which returns a channel), but I think using a set of guidelines will make it a lot easier for us to maintain the code going forward.

PR #289, opened 2019-07-22
expirationwatch: Use consistent asynchronous API

albrow:
As mentioned in #96, we are moving towards making our asynchronous APIs more consistent across all packages. This PR applies the new conventions to the `expirationwatch` package only.

One consequence of making this change one package at a time is that there are some places where we need to awkwardly bridge the various old asynchronous coding styles with the new style. Some things will get uglier before they get prettier. For example, we are now creating contexts and storing cancel funcs for the sole purpose of passing them into `expirationwatch.Watcher.Watch`. But after all packages transition to the new conventions, we will be able to inherit from a parent context instead of creating a new one.

Since we are changing the API anyways, I also wanted to rename `Start` to `Watch` and `Receive` to `ExpiredItems`. This mimics methods from the standard library like [`http.ListenAndServe`](https://golang.org/pkg/net/http/#ListenAndServe) in that they are more descriptive than generic names like `Start` and `Stop`.

### DIFF
--- a/expirationwatch/expiration_watcher_test.go
+++ b/expirationwatch/expiration_watcher_test.go
@@ -1,7 +1,9 @@
 package expirationwatch
 
 import (
+	"context"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


expirationwatch/expiration_watcher_test.go
```diff
@@ -123,38 +125,44 @@ func TestRemoveItemWhichSharesExpirationTimeWithOtherItems(t *testing.T) {
 	assert.Equal(t, expiryEntryOne, pruned[0])
 }
 
-func TestStartsAndStopsPoller(t *testing.T) {
+func TestStartAndStopPoller(t *testing.T) {
 	watcher := New(0)
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
fabioberger: Since we are explicitly calling cancel below, do we need to defer calling it as well?

albrow: Yes, because the function could return (e.g. if any `assert.Require` fails) before we reach the explicit call to `cancel`, which would result in a leaked goroutine.

PR #300, opened 2019-07-23
Use consistent asynchronous API for core package

### LINKED ISSUE #299: Fail fast after calling app.Close
I've seen the following error in our logs:

```
    "error": "leveldb: closed",
    "level": "error",
    "msg": "could not check if order was already stored",
```

This happens because a previous error caused us to call `app.Close` (which closes the database) but we didn't fully shutdown the process. Some goroutines were still running and trying to use the database.

The situation can likely be fixed by #96 and #161 but we might also want to call `os.Exit` or `log.Fatal` in these situations as a short term fix (effectively a non-graceful shutdown).

albrow:
Related to #96.

Since I was already changing some related code, I went ahead and fixed #299 in this PR too.

### DIFF
--- a/cmd/mesh/main.go
+++ b/cmd/mesh/main.go
@@ -6,6 +6,8 @@
 package main
 
 import (
+	"context"
+
 	"github.com/0xProject/0x-mesh/core"
 	"github.com/plaid/go-envvar/envvar"
 	log "github.com/sirupsen/logrus"
@@ -39,7 +41,9 @@ func main() {
 	if err != nil {
 		log.WithField("error", err.Error()).Fatal("could not initialize app")
 	}
-	if err := app.Start(); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	if err := app.Start(ctx); err != nil {
 		log.WithField("error", err.Error()).Fatal("fatal error while starting app")
 	}
 	defer app.Close()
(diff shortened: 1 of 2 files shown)


cmd/mesh/main.go
```diff
@@ -39,7 +41,9 @@ func main() {
 	if err != nil {
 		log.WithField("error", err.Error()).Fatal("could not initialize app")
 	}
-	if err := app.Start(); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
albrow: It's a little awkward that we have both `defer cancel()` and `defer app.Close` but that will go away once all the packages that `core` uses are also using the same consistent asynchronous API.

core/core.go
```diff
@@ -301,34 +301,38 @@ func (app *App) Start() error {
...
+			log.WithError(err).Fatal("snapshot expiration watcher exited with error")
 		}
 	}()
 
 	return nil
 }
 
-func (app *App) periodicallyCheckForNewAddrs(startingAddrs []ma.Multiaddr) {
+func (app *App) periodicallyCheckForNewAddrs(ctx context.Context, startingAddrs []ma.Multiaddr) {
+	// TODO: There might be a more efficient way to do this if we have access to
```
```suggestion
	// TODO(albrow): There might be a more efficient way to do this if we have access to
```

PR #301, opened 2019-07-23
core: implement network ID detection check on startup

### LINKED ISSUE #237: orderwatch: store networkId in DB
Currently Mesh does not store the `networkID` persistently corresponding to the block headers stored. If a Mesh operator is on Ganache (very low block numbers) and then switches to mainnet, Mesh will attempt to sync up to the latest block (potentially 8,000,000+ blocks) since it assumes we were previously on mainnet as well. 

By persistently storing the `networkID`, we can throw an error requiring the operator to remove the `orders` and `blocks` stored in the DB as they correspond to a different network.

hrharder:
_Functional but a work in progress, a few outstanding points (see notes section)._

## Overview

Closes #237 
- Add a `Metadata` collection to MeshDB to store the networkID we are starting up with
- Add a check during startup that checks if a networkID has been set, compares to the one loaded from the environment, and exits if they do not match.
- The detection check (function `initNetworkId`) also stores the network ID during the initial startup (no previous DB)

## Steps for testing:

1. Start mesh on network N (let it run, collect orders, etc)
```
P2P_LISTEN_PORT=<port> ETHEREUM_RPC_URL=<url> ETHEREUM_NETWORK_ID=3 ./mesh
```
2. Kill the process, and restart with a new network ID:
```
P2P_LISTEN_PORT=<port> ETHEREUM_RPC_URL=<url> ETHEREUM_NETWORK_ID=4 ./mesh
```

You should see:
```
{"level":"error","msg":"Mesh previously started on different Ethereum network; switch networks or remove DB","time":"2019-07-23T16:38:09-07:00"}
{"error_string":"expected networkID to be '4', got '3","level":"fatal","msg":"could not initialize app","time":"2019-07-23T16:38:09-07:00"}
```

## Notes
Still a WIP due to me having questions about the following:
- Is there a preferred key for the new collection other than `[]byte("metadata")`? (see [here](https://github.com/hrharder/0x-mesh/blob/development/meshdb/meshdb.go#L55))
- Are there any changes to method names, comments, etc you would like to see?
- Am I handling the mismatched ID case as you would expect? (see [here](https://github.com/hrharder/0x-mesh/blob/development/core/core.go#L278))


core/core.go
```diff
@@ -250,6 +257,42 @@ func initPrivateKey(path string) (p2pcrypto.PrivKey, error) {
 	return nil, err
 }
 
+func initNetworkId(networkID int, db *meshdb.MeshDB) error {
+	// see if there is already a stored networkID
+	snapshot, err := db.Metadata.GetSnapshot()
```
albrow: You don't need to use a snapshot here. The [doc comment for `Snapshot`](https://godoc.org/github.com/0xProject/0x-mesh/db#Snapshot) explains:

> Snapshot is a frozen, read-only snapshot of a DB state at a particular point in time.

Snapshots are only really needed when other goroutines can change the state of the database and you want a single consistent view which ignores those changes. For example, we use snapshots for the `GetOrders` JSON-RPC API endpoint, because we don't want the orders to change while you're in the middle of paginating through them.

In this case, no other goroutine is going to change the metadata, so we don't need to call `GetSnapshot`.

core/core.go
```diff
@@ -250,6 +257,42 @@ func initPrivateKey(path string) (p2pcrypto.PrivKey, error) {
...
+	if err != nil {
+		return err
+	}
+
+	filter := db.Metadata.EthereumNetworkIDIndex.All()
```
albrow: We don't need to use an index, filter, or query. Those features are for efficiently selecting a specific model or group of models from a collection. In this case, we know that the collection only holds exactly zero or one models. We can just get that model directly:

```go
metadata, err := db.Metadata.FindByID([]byte{})
if _, ok := err.(NotFoundError); ok {
    // No previous metadata was found. We need to save the metadata.
}  else if err != nil {
    // Some unexpected error occurred. We should return it.
}
// If we reached here, metadata was previously saved. We should check if the Ethereum
// network id matches what we expect and might also do other checks in the future.
```

Let me know if that makes sense or if you have any other questions.

meshdb/meshdb.go
```diff
@@ -45,11 +45,22 @@ func (o Order) ID() []byte {
...
 }
 
+// Metadata is the database representation of MeshDB instance metadata
+type Metadata struct {
+	EthereumNetworkID *big.Int
+}
+
+// ID returns the id used for the metadata collection (one per DB)
+func (m Metadata) ID() []byte {
+	return []byte("metadata")
```
albrow: For storing the metadata, we want to have a collection which contains exactly zero or one models. The `db` package wasn't really written with this in mind and up until now a collection always contained potentially many models.

Since there is always just one model in the collection, we don't really need a unique identifier for each model. To help save space, I suggest that we just use an empty ID here.

meshdb/meshdb.go
```diff
@@ -67,6 +78,12 @@ type OrdersCollection struct {
 	IsRemovedIndex                       *db.Index
 }
 
+// MetadataCollection represents a DB collection used to store instance metadata
+type MetadataCollection struct {
+	*db.Collection
+	EthereumNetworkIDIndex *db.Index
```
albrow: As mentioned before, we don't need to use an Index. We can simply remove this part of the code.

core/core.go
```diff
@@ -250,6 +257,42 @@ func initPrivateKey(path string) (p2pcrypto.PrivKey, error) {
...
+	if err != nil {
+		return err
+	}
+
+	switch len(metadataCol) {
+	case 1:
+		// on subsequent startups, verify we are on the same network
+		loadedNetworkID := metadataCol[0].EthereumNetworkID.Int64()
+		if loadedNetworkID != int64(networkID) {
+			log.Error("Mesh previously started on different Ethereum network; switch networks or remove DB")
```
albrow: Let's change this to:

```go
err := fmt.Errorf("expected networkID to be %d but got %d", loadedNetworkID, networkID)
log.WithError(err).Error("Mesh previously started on different Ethereum network; switch networks or remove DB")
return err
```

albrow:
For testing, we can just call `initNetworkID` directly and make sure that it behaves as expected. For example, calling `initNetworkID` twice with two different network IDs should result in an error.

core/core.go
hrharder: Makes sense, thanks for the tip. I was using some other methods as a guide, and assumed a snapshot should be used for all reads. will make this fix!

core/core.go
hrharder: Sweet -- this is much better. Are you passing an empty byte slice to `FindByID` just as an example, or is that exactly how it should be done? 

The key on the collection is `[]byte("metadata")` so should that be used?

meshdb/meshdb.go
hrharder: Gotcha - sounds good. I was unsure if that was possible. (This nullifies my previous question)

meshdb/meshdb.go
hrharder: Great -- seemed like overkill as I was doing it, but was closely modeling the new collection after existing ones.

hrharder:
@albrow I believe I addressed all your comments, please advise if I missed something. Will plan on pushing a test soon.

core/core.go
```diff
@@ -250,6 +257,32 @@ func initPrivateKey(path string) (p2pcrypto.PrivKey, error) {
 	return nil, err
 }
 
+func initNetworkId(networkID int, meshDB *meshdb.MeshDB) error {
+	var metadata meshdb.Metadata
+	err := meshDB.Metadata.FindByID([]byte{0}, &metadata)
```
albrow: After looking at this again, I think it is a leaky abstraction to expose the fact that we are using `[]byte{0}` for the ID in the `core` package. We had similar concerns about the use of the special separator `"|"` in the `Orders` collection, and to address those concerns we added some helper methods to the `meshdb` package (e.g. `FindOrdersByMakerAddressAndMaxSalt` and `FindOrdersLastUpdatedBefore`). These methods hide the details about the structure of keys in the database and make the `meshdb` package easier to use.

Can we take a similar approach here and create two helper function in the `meshdb` pacakge?

- `GetMetadata` returns the metadata (or a `db.NotFoundError` if no metadata has been found).
- `SaveMetadata` inserts the metadata into the database.

core/core.go
```diff
@@ -119,6 +120,12 @@ func New(config Config) (*App, error) {
 		return nil, err
 	}
 
+	// Check if the DB has been previously intialized with a different networkId
+	err = initNetworkId(config.EthereumNetworkID, meshDB)
```
albrow: For cases like this, when a function returns an `error` as it's only return value, we like to use the following shorthand:

```go
if err := initNetworkId(config.EthereumNetworkID, meshDB); err != nil {
    return nil, err
}
```

It has the advantage of limiting the scope of `err`.

core/core.go
```diff
@@ -250,6 +257,32 @@ func initPrivateKey(path string) (p2pcrypto.PrivKey, error) {
 	return nil, err
 }
 
+func initNetworkId(networkID int, meshDB *meshdb.MeshDB) error {
```
albrow: Let's follow our naming conventions in other parts of the code (and in `libp2p` and _most_ of the standard library) and call this function `initNetworkID` with a capital `D`.

(As an aside, my personal opinion is that the version you have here is correct since "id" is an abbreviation and not an acronym. But nonetheless the convention of using `ID` in Go is pretty well established at this point and some linters will even complain about it.)

core/core_test.go
```diff
@@ -0,0 +1,28 @@
+package core
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
```
albrow: Let's remove the unnecessary space here.

core/core.go
hrharder: Agreed -- is a bit smelly to see in `core`.

core/core.go
hrharder: Will do. I personally agree with you, readability is lost IMO in a case like `NetworkIDDetected`, but of course we should stick to the established convention.

core/core.go
hrharder: I'm going to push a commit with the `SaveMetadata` with the following signature, LMK if you would rather accept `metadata` as the argument. I figure either implementation is relatively easy to extend with future metadata. 

```go
func (m *MeshDB) SaveMetadata(networkID int) error
```

hrharder:
The commit I just pushed should address all the comments. My only remaining questions are:

- Should `SaveMetadata` accept the `networkID` or a `meshdb.Metadata` object?
- Right now, `GetMetadata` returns any error coming from `FindByID`, is that OK? or should I explicitly return a `ErrNotFound`?

core/core.go
```diff
@@ -250,6 +255,28 @@ func initPrivateKey(path string) (p2pcrypto.PrivKey, error) {
...
+	metadata, err := meshDB.GetMetadata()
+	if _, ok := err.(db.NotFoundError); ok {
+		// No stored metadata found (first startup)
+		if err := meshDB.SaveMetadata(networkID); err != nil {
+			return err
+		}
+		return nil
+	} else if err != nil {
+		return err
+	}
```
fabioberger: Slight preference for checking if `err != nil` first, and then nesting the err type checking beneath it. We should really only do the error type checking if we have an error.

```
if err != nil {
  	if _, ok := err.(db.NotFoundError); ok {
		// No stored metadata found (first startup)
		if err := meshDB.SaveMetadata(networkID); err != nil {
			return err
		}
		return nil
	}
        return err
}
```

meshdb/meshdb.go
```diff
@@ -45,11 +45,22 @@ func (o Order) ID() []byte {
 	return o.Hash.Bytes()
 }
 
+// Metadata is the database representation of MeshDB instance metadata
+type Metadata struct {
+	EthereumNetworkID *big.Int
```
fabioberger: `ETHEREUM_NETWORK_ID` is supplied as an `int` and is unlikely to become so large as to need representation as a `*big.Int`. My vote would be we represent it as an `int` in the DB as well.

hrharder: Totally agree. Meant to make that change in my last commit. I initially used `big.Int` when under the impression the value should be indexed and that type provides a simple way to access a byte slice representation.

Admittedly doesn't make a lot of sense looking back on it 😂

hrharder:
@fabioberger do you have any preference on the signature of the `SaveMetadata` method? IE should it accept a `Metadata` value of the network ID? I'm unsure which would be more extensible in the future with more possible metadata values.

meshdb/meshdb.go
```diff
@@ -84,10 +100,16 @@ func NewMeshDB(path string) (*MeshDB, error) {
...
+	metadata, err := setupMetadata(database)
+	if err != nil {
+		return nil, err
+	}
+
 	return &MeshDB{
 		database:    database,
 		MiniHeaders: miniHeaders,
 		Orders:      orders,
+		Metadata:    metadata,
```
albrow: Let's make the `Metadata` field unexported at this point, since users of the `meshdb` package should only need to two methods: `GetMetadata` and `SaveMetadata`.

PR #302, opened 2019-07-24
p2p: Use consistent asynchronous API

albrow:
Related to #96.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -29,6 +29,7 @@ import (
 	ethrpc "github.com/ethereum/go-ethereum/rpc"
 	"github.com/google/uuid"
 	p2pcrypto "github.com/libp2p/go-libp2p-crypto"
+	peer "github.com/libp2p/go-libp2p-peer"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
 	log "github.com/sirupsen/logrus"
@@ -86,6 +87,8 @@ type snapshotInfo struct {
 
 type App struct {
 	config                    Config
+	peerID                    peer.ID
+	privKey                   p2pcrypto.PrivKey
 	db                        *meshdb.MeshDB
 	node                      *p2p.Node
 	networkID                 int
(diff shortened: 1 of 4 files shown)


core/core.go
```diff
@@ -108,6 +111,18 @@ func New(config Config) (*App, error) {
 	log.SetLevel(log.Level(config.Verbosity))
 	log.AddHook(loghooks.NewKeySuffixHook())
 
+	// Load private key and add peer ID hook.
+	privKeyPath := filepath.Join(config.DataDir, "keys", "privkey")
```
albrow: As a side effect of not being able to call `p2p.New` without a `context.Context`, we need to load the private key first so we can find our peer ID and add the `PeerIDHook`.

p2p/node.go
```diff
@@ -99,25 +96,27 @@ type Config struct {
...
 }
 
-// New creates a new Node with the given config.
-func New(config Config) (*Node, error) {
-
-	nodeCtx, cancel := context.WithCancel(context.Background())
-
+// New creates a new Node with the given context and config. The Node will stop
+// all background operations if the context is canceled.
+func New(ctx context.Context, config Config) (*Node, error) {
```
albrow: I tried to write the `p2p` package such that `New` does not require a context and doesn't do any processing in the background. Unfortunately it is not practical to do so because of how `libp2p.Host` works. In order to create a `Host`, with `libp2p.New`, you need to pass in a context. So `p2p.New` expects a context and `p2p.Start` re-uses the context provided to `New`.

core/core.go
```diff
@@ -269,6 +262,22 @@ func initNetworkID(networkID int, meshDB *meshdb.MeshDB) error {
 }
 
 func (app *App) Start(ctx context.Context) error {
+	// Initialize the p2p node.
+	nodeConfig := p2p.Config{
```
albrow: Because `p2p.New` requires a context, we needed to move it into `core.App.Start`.

p2p/node_test.go
```diff
@@ -87,43 +87,117 @@ func newTestNode(t *testing.T) *Node {
...
-		node1.host.Network().Notify(notifee)
+		node.host.Network().Notify(notifee)
 	}
-	connectCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-	defer cancel()
+
+	return node
+}
+
+func connectTestNodes(t *testing.T, node0, node1 *Node) {
```
albrow: I had to change the tests around quite a bit because we need more fine-grained control around when and how `New`, `Start`, and `Connect` are called.

core/core.go
```diff
@@ -535,9 +542,6 @@ func (app *App) SubscribeToOrderEvents(sink chan<- []*zeroex.OrderEvent) event.S
 
 // Close closes the app
 func (app *App) Close() {
-	if err := app.node.Close(); err != nil {
```
albrow: We're one step closer to removing `core.App.Close`.

core/core.go
```diff
@@ -269,6 +262,22 @@ func initNetworkID(networkID int, meshDB *meshdb.MeshDB) error {
 }
 
 func (app *App) Start(ctx context.Context) error {
```
albrow: @fabioberger In a previous PR I mentioned that we want to avoid generic names like `Start` and `Stop`. I couldn't think of a name in this case that isn't overly verbose. Open to suggestions if you have any.

Issue #161, opened 2019-06-13
Shutdown gracefully


opaolini:
# About

Mesh should attempt to gracefully shutdown upon receiving a `SIGTERM`, basically allowing it to drain active connections and shutdown services. In the future it should also respond as `unhealthy`/`unready` when probed for health or readiness.

Having a graceful shutdown would also benefit operators running mesh on Kubernetes (and probably other orchestration tools), as when pods are to be updated/moved/evicted `kubelet` sends a `SIGTERM` to the main process and waits for up to a grace period (by default 30 seconds) until it sends a `SIGKILL`.

Below is a rough sketch implementation of that, however I should deliver a working PR in the coming days if interested!

## Rough implementation

```go
done := make(chan os.Signal, 1)
signal.Notify(done, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)

// Start core.App.
app, err := core.New(coreConfig)
if err != nil {
  log.WithField("error", err.Error()).Fatal("could not initialize app")
}
if err := app.Start(); err != nil {
  log.WithField("error", err.Error()).Fatal("fatal error while starting app")
}
defer app.Close()

// Start RPC server.
go func() {
  err := listenRPC(app, config)
  if err != nil {
    app.Close()
    log.WithField("error", err.Error()).Fatal("RPC server returned error")
  }
}()

// mesh is running at this point
<-done

ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer func() {
  // additional handling here
  cancel()
}()

// handle stopping services, and draining connections
if err := app.Shutdown(ctx); err != nil {
  log.Fatalf("mesh shutdown failed with err:%+v", err)
}
log.Info("mesh stopped")

```

albrow:
Definitely want to do this at some point. We might need to address #96 first.

albrow:
#27 is also potentially related.

fabioberger:
Another part of this will be revisiting all uses of `context.Background()` and making sure we actually pass down the top-level context for the entire application, so all internal ongoing requests are aborted if it gets cancelled.

PR #307, opened 2019-07-25
orderwatch: Use consistent asynchronous API 

albrow:
related to #96.

### DIFF
--- a/cmd/mesh/rpc_handler.go
+++ b/cmd/mesh/rpc_handler.go
@@ -112,9 +112,6 @@ func SetupOrderStream(ctx context.Context, app *core.App) (*ethRpc.Subscription,
 
 		for {
 			select {
-			case <-ctx.Done():
-				orderWatcherSub.Unsubscribe()
-				return
 			case orderEvents := <-orderEventsChan:
 				err := notifier.Notify(rpcSub.ID, orderEvents)
 				if err != nil {
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -19,10 +19,10 @@ import (
 	logger "github.com/sirupsen/logrus"
 )
 
-// minCleanupInterval specified the minimum amount of time between orderbook cleanup intervals. These
-// cleanups are meant to catch any stale orders that somehow were not caught by the event watcher
-// process.
-var minCleanupInterval = 1 * time.Hour
+// cleanupInterval specified the amount of time between orderbook cleanups.
+// These cleanups are meant to catch any stale orders that somehow were not
+// caught by the event watcher process.
+var cleanupInterval = 1 * time.Hour
 
 // lastUpdatedBuffer specifies how long it must have been since an order was last updated in order to
 // be re-validated by the cleanup worker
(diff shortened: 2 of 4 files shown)


cmd/mesh/rpc_handler.go
```diff
@@ -112,9 +112,6 @@ func SetupOrderStream(ctx context.Context, app *core.App) (*ethRpc.Subscription,
 
 		for {
 			select {
-			case <-ctx.Done():
```
albrow: This fixes a bug in #304 that was previously unnoticed. `ctx` here is not something we control and it seems to be canceled even though a client is still actively subscribed. I can't find any documentation about the expected behavior of `ctx` at https://godoc.org/github.com/ethereum/go-ethereum/rpc. For now, we have no real choice but to ignore the given context.

zeroex/orderwatch/order_watcher.go
```diff
@@ -99,50 +93,321 @@ func New(meshDB *meshdb.MeshDB, blockWatcher *blockwatch.Watcher, orderValidator
...
+	for {
+		select {
+		case <-ctx.Done():
+			cleanupTicker.Stop()
+			w.blockSubscription.Unsubscribe()
+			close(blockEvents)
+			return nil
+		case expiredOrders := <-w.expirationWatcher.ExpiredItems():
+			w.handleExpiration(expiredOrders)
+		case <-cleanupTicker.C:
```
albrow: @fabioberger one consequence of this change is that the cleanup worker runs _synchronously_ with all other handlers in `mainLoop`. In other words, while the cleanup worker is running, we will stop processing block events and expirations.

This new version has some nice benefits:

1. It allows us to handle errors synchronously and will make it much easier to implement graceful shutdowns in the future (see #161).
2. It helps prevent race conditions because no one else will be validating/updating/removing orders while the cleanup worker is running.
3. It may make it easier for us to implement rate-limiting for Etheruem RPC requests, but it's hard to say for sure until we actually implement it.

However, this does come with an obvious downside in terms of performance. You have a better sense of how long the cleanup worker will take to run and how it compares to the other tasks within the `orderwatch` package. Do you think this trade-off is worth it or do you want to go back to having the cleanup worker run asynchronously?

fabioberger: The cleanup working needs to re-validate _every_ order stored in the DB. This will involve multiple Ethereum JSON RPC requests and so can take a considerable amount of time. I worry that this could take several seconds to minutes depending on the number of orders stored. With the number of orders in Mesh today, it probably takes a few seconds (which is not a deal-breaker), but I worry this approach won't work once the network has more orders (it would only take one stress-tester to make happen).

I feel like running it synchronously with all other handlers is not tenable. Let's discuss this some more offline.

zeroex/orderwatch/order_watcher.go
```diff
@@ -99,50 +93,382 @@ func New(meshDB *meshdb.MeshDB, blockWatcher *blockwatch.Watcher, orderValidator
...
+func (w *Watcher) handleExpiration(expiredOrders []expirationwatch.ExpiredItem) {
+	orderEvents := []*zeroex.OrderEvent{}
+	for _, expiredOrder := range expiredOrders {
+		order := &meshdb.Order{}
+		err := w.meshDB.Orders.FindByID(common.HexToHash(expiredOrder.ID).Bytes(), order)
+		if err != nil {
+			logger.WithFields(logger.Fields{
+				"error":     err.Error(),
+				"orderHash": expiredOrder.ID,
+			}).Warning("Order expired that was no longer in DB")
```
fabioberger: Should we keep this as a `Warning` or lower this to a `Trace`? It is expected to happen from time to time.

albrow: Sure. Good idea.

PR #312, opened 2019-07-27
Final changes to make all code follow our async conventions

### LINKED ISSUE #96: Make asynchronous APIs more consistent

albrow:
Fixes #96.

Depends on #307. After that PR is merged I'll change the base branch for this one to `development`.

Summary of changes:
- Make `core.App.Start` blocking.
- Remove `core.App.Stop`.
- Avoid `log.Panic` and `log.Fatal` whenever possible (there are some lingering cases where e.g., we pass a function into another function and don't have a good way to return errors).
- Remove the `blockwatch.Watcher.Errors` channel. For non-critical errors, we log them in `blockwatch.Watcher` directly. For critical errors, we return them in `blockwatch.Watcher.Watch`. This makes `blockwatch` more aligned with the rest of our async code.
- Don't start any goroutines without a way to stop them (usually by waiting for `ctx.Done()` but sometimes with short timeouts or other signals).

### DIFF
--- a/ethereum/blockwatch/block_watcher.go
+++ b/ethereum/blockwatch/block_watcher.go
@@ -54,7 +54,6 @@ type Config struct {
 // handling block re-orgs and network disruptions gracefully. It can be started from any arbitrary
 // block height, and will emit both block added and removed events.
 type Watcher struct {
-	Errors              chan error
 	blockRetentionLimit int
 	startBlockDepth     rpc.BlockNumber
 	stack               *Stack
@@ -73,10 +72,7 @@ type Watcher struct {
 func New(config Config) *Watcher {
 	stack := NewStack(config.MeshDB, config.BlockRetentionLimit)
 
-	// Buffer the first 5 errors, if no channel consumer processing the errors, any additional errors are dropped
-	errorsChan := make(chan error, 5)
 	bs := &Watcher{
-		Errors:              errorsChan,
 		pollingInterval:     config.PollingInterval,
 		blockRetentionLimit: config.BlockRetentionLimit,
 		startBlockDepth:     config.StartBlockDepth,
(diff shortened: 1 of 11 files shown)


ethereum/blockwatch/block_watcher.go
```diff
@@ -426,6 +420,12 @@ func (w *Watcher) getLogsInBlockRange(from, to int) ([]types.Log, int) {
 			go func(index int, b *blockRange) {
 				defer wg.Done()
 
+				select {
+				case <-ctx.Done():
+					return
```
fabioberger: We need to add an entry to `indexToLogResult` before returning otherwise the code below will panic. It expects there to be an entry for every index.

Let's set it to:
```
indexToLogResult[index] = logRequestResult{
					From: b.FromBlock,
					To:   b.ToBlock,
					Err:  errors.New("Parent context was canceled"),
					Logs: []types.Log{},
				}
```

albrow: Just added this.

PR #314, opened 2019-07-27
Reduce verbosity of "fetching block logs" message

albrow:
@fabioberger I think the verbosity of this log should change from `Info` to `Trace`. Whenever I run Mesh locally for quick tests, this log message is taking up a large percentage of the total number of logs. I don't think it is helpful to see hundreds of logs for what is a pretty common occurrence. What do you think?


fabioberger:
Does it still print out a one-liner, explaining why it's not starting the node immediately? I'm fine with removing this so long as there is still _a_ log explaining what the node is doing.

Issue #199, opened 2019-07-02
validation: reconsider OrderAlreadyAdded rejection


dekz:
In using Mesh and integrating with Launch Kit Backend I got bit by the `OrderAlreadyAdded` rejection error. 

Here was the thought process around restarting Launch Kit Backend and ensuring all orders are in Mesh.

1. Reboot Backend and Mesh
2. Query DB for all orders
3. Post all orders to Mesh because backend didn't receive updates during downtime
4. Remove all rejected orders from the database

Since `OrderAlreadyAdded` orders and a persistent Mesh db, most orders were removed completely from the database due to (4). Whilst it's not difficult to filter for `OrderAlreadyAdded` it might be worth re-considering if this is a requirement or if it can perform in a more idempotent way.

When orders which are already added are valid (they just already exist in mesh), it would be preferred if they were accepted again and all rejected orders are actually non-fillable orders, not a potential combination of fillable and non-fillable.

This may also allow for a quick, efficient, final check before an on-chain operation.

I'm open to whether this order gets fully validated again versus just returned the current `fillableTakerAssetAmount`.

fabioberger:
Thanks for the suggestion @dekz. After thinking carefully about this, I don't think using the `AddOrders` endpoint to re-validate the orders in your DB is the correct long-term approach. Instead, you should use `GetOrders` (once this PR merges: https://github.com/0xProject/0x-mesh/pull/198). 

The difference here is that `AddOrders` will _only_ check if the orders already in your DB are still valid. It will _not_ let you know of any orders _added_ to Mesh during the clients downtime. `GetOrders` will get all the orders currently in the Mesh DB, and these can be cross-referenced with those in your local DB. 

- Orders in your DB that are no longer in Meshes were removed so delete them
- Orders changed in Mesh should be updated in your DB
- Orders found in Mesh that aren't in your DB should be added

Now users who are only using Mesh as an OrderWatcher replacement probably don't care about orders found by Mesh that they didn't add themselves, making this difference a moot point. I also like your reasoning about only invalid orders being rejected from Mesh. So perhaps even though `GetOrders` is the more complete solution, we should re-consider the `OrderAlreadyAdded` rejection in favor of simply accepting the order again and returning it's updated state. 

Thoughts on this @albrow?

PR #316, opened 2019-07-27
core: Handle stored orders differently if flagged for removal

fabioberger:
Fixes: #199


albrow:
@fabioberger this change messes up some of the other logic in the `core` package. For example, [we assume that every order in `ValidationResults.Accepted` is a new unique valid order and log a message indicating that](https://github.com/0xProject/0x-mesh/blob/85a6b90f8fd6e72005c307f4fe4b7985c7ef7151/core/message_handler.go#L145). It's important that we don't log this message for orders which we've already seen. It might also have implications for our peer ranking system but at this point it's too early to tell.

I think we still need a way to indicate that an order is valid but already stored. What do you think?

Issue #253, opened 2019-07-14
Automate release of new versions


albrow:
Right now releasing new versions is a manual process, but it seems like something that could be automated.

Note: we might want to use "Automated releases" in DockerHub: https://cloud.docker.com/u/0xorg/repository/docker/0xorg/mesh/builds

fabioberger:
Additional things that would be nice to automate during a release:

- Publishing the TS client to NPM
- Ensuring the Markdown docs for TS client have been updated before publishing

fabioberger:
Update: the remaining things that are not automated are:

- Posting the release notes to Github
- Publishing the TS client to NPM
- Creating the Github tag, and matching branch (the branch is needed by GitBook)
- Adding the new version to GitBook version dropdown
- Updating the Mesh bootstrap nodes

PR #317, opened 2019-07-28
docs: Support GitBook

fabioberger:
I spent some time looking into [GitBook](https://www.gitbook.com/) as a potential way to render the Mesh docs. It's come a long way since the last time I checked it out. There are several benefits we gain from using GitBook:

1. It syncs our Github repo markdown files with the rendered docs so they are always up-to-date and the authoritative copy lives in our repo.
2. We can have a version picker render on the GitBook docs, and clicking on a version will show you the docs for that particular version. It does this using Github branches instead of tags, but it's a small price to pay.
3. Amazing search that "Just Works TM". (Note: Search won't activate until we upgrade to a paid plan)

Take a look at what it would look like: https://0x-mesh.gitbook.io/docs/

This PR:

- Re-organizes our documentation markdown files so that they render beautifully on our [GitBook docs](https://0x-mesh.gitbook.io/docs/) page
- Moves all doc markdown files into a 'docs' directory
- Adds TypeDoc commands to the TS client and generates MD docs for it
- Removes emojis from README since they don't render on GitBook
- Removes the badges from the top of doc files since they are no longer necessary
- Adds a `make cut-release` command that:
  -  updates all hard-coded versions across the repo
  -  generates the TS client docs


cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/plaid/go-envvar/envvar"
+)
+
+type envVars struct {
+	// Version is the new release version to use
+	Version string `envvar:"VERSION" default:"development"`
```
albrow: I think we should not have a default here. `development` is only what you want some of the time and is not necessarily a safe default.

cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	updateHardCodedVersions(env.Version)
+
+	generateTypescriptClientDocs()
+}
+
+func generateTypescriptClientDocs() {
+	// Run `yarn install` to make sure `TypeDoc` dep is installed
+	cmd := exec.Command("yarn", "install")
```
albrow: I think we should use the `--frozen-lockfile` here. We aren't adding any new dependencies and we don't want to make any changes to the lockfile in a commit that should solely be responsible for changing the version numbers and updating the documentation.

cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+	cmd.Dir = "rpc/clients/typescript"
+	if err := cmd.Run(); err != nil {
+		log.Fatal(err)
```
albrow: `err` here doesn't include the output from the command itself. I think we should also log [`cmd.CombinedOutput`](https://golang.org/pkg/os/exec/#Cmd.CombinedOutput).

cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	// Modify the paths to be prefixed with `json_rpc_clients/typescript`
+	modifiedDat := strings.Replace(string(dat), "](", "](json_rpc_clients/typescript/", -1)
+	modifiedDat = strings.Replace(modifiedDat, "](json_rpc_clients/typescript/)", "]()", -1)
+	finalTsClientSummary := strings.Replace(modifiedDat, "* [", "  * [", -1)
+
+	// Replace the summary content nested under `Typescript client`
+	mainSummaryPath := "docs/SUMMARY.md"
+	dat, err = ioutil.ReadFile(mainSummaryPath)
```
albrow: We could optionally use a [`buffio.Scanner`](https://golang.org/pkg/bufio/#Scanner) to cut down on memory usage. The current version loads the entire file into memory at once, but `buffio.Scanner` can load it into memory in pieces. Might not necessarily need to do it now, but It matters more if __SUMMARY.md__ is large.

```go
    summaryFile, err := os.Open("docs/SUMMARY.md")
    if err != nil {
        log.Fatal(err)
    }
    defer file.Close()

    scanner := bufio.NewScanner(summaryFile)
    for scanner.Scan() {
        // scanner.Text() tells you the current line.
    }
    if err := scanner.Err(); err != nil {
        log.Fatal(err)
    }
```

Of course, we would also need to figure out a way to replace/cut certain lines without loading the entire file into memory.

cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	finalSummaryLines := []string{}
+	lines := strings.Split(string(dat), "\n")
+	isCutting := false
+	for _, l := range lines {
```
albrow: I think this could use a comment explaining the significance of the `"## Contributing"` and `"[Typescript client]"` lines.

cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Update the version string in all files that must be updated for a new release
+func updateHardCodedVersions(version string) {
+
+	// Update `rpc/clients/typescript/package.json`
+	tsClientPackageJSONPath := "rpc/clients/typescript/package.json"
+	updateVersionKey(tsClientPackageJSONPath, version)
+
+	// Update `core.ts`
```
albrow:
```suggestion
	// Update `core.go`
```

cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+	readmePath := "README.md"
+	updateBadge(readmePath, version)
+}
+
+func updateVersionKey(filePath string, version string) {
+	dat, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		log.Fatal(err)
+	}
+	newVersionString := fmt.Sprintf(`"version": "%s",`, version)
```
albrow: If we follow the same flow that we have been using, `"version"` will be the previously released version (e.g. `"1.0.6-beta"`), not `"development"`. Since we don't know what the previous version was, we might need to use a regex to find/replace what we're looking for instead of a hardcoded string.

cmd/cut-release/main.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func updateBadge(filePath string, version string) {
+	dat, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		log.Fatal(err)
+	}
+	doubleDashVersion := strings.Replace(version, "-", "--", -1)
+	newSvgName := fmt.Sprintf("version-%s-orange.svg", doubleDashVersion)
+	modifiedDat := []byte(strings.Replace(string(dat), "version-development-orange.svg", newSvgName, 1))
```
albrow: Same here regarding the hard-coded `"development"` string.

docs/DEPLOYMENT.md
```diff
@@ -1,7 +1,3 @@
-[![Version](https://img.shields.io/badge/version-development-orange.svg)](https://github.com/0xProject/0x-mesh/releases)
```
albrow: Might some people still look at the docs on GitHub instead of on GitBook? Maybe it is still useful to include the version badge here?

docs/SUMMARY.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Getting started
+
+* [Deployment guide](DEPLOYMENT.md)
+* [Usage guide](USAGE.md)
+
+## JSON-RPC clients
+
+* [Golang client](https://godoc.org/github.com/0xProject/0x-mesh/rpc)
+* [Typescript client](json_rpc_clients/typescript/README.md)
```
albrow: nit: can we call it "TypeScript" with a capital "S" to match the spelling on the [official website](https://www.typescriptlang.org/)?

albrow:
I think this also fixes #253.

cmd/cut-release/main.go
albrow: Edit: I just realized `CombinedOutput` actually runs the command so what we really want to do is replace `Run` with `CombinedOutput`.

cmd/cut-release/main.go
fabioberger: This doesn't feel worthwhile for an utility command such as this. I also don't expect `SUMMARY.md` to ever be larger then twice it's current size (5kb).

docs/DEPLOYMENT.md
fabioberger: Agreed. It looks _so so_ on GitBook but agree it's probably worth it for people reading the docs on Github. Added it back!

Issue #201, opened 2019-07-05
rpc: add configuration / status


dmvt:
I'd like to be able to request the configuration and current status from an rpc endpoint. This would include:

- ethereum network id
- current block number / last block number received by mesh

fabioberger:
Thanks for the suggestion @dmvt. I really like this idea. I've moved out your second suggestion into it's own PR: https://github.com/0xProject/0x-mesh/issues/202.

fabioberger:
Thoughts on adding this @albrow?

albrow:
@dmvt the [Ethereum network id is an environment variable](https://github.com/0xProject/0x-mesh/blob/master/DOCUMENTATION.md#possible-environment-variables) which is required during initialization. It's set by you and will never change unless you change it, so I think there is limited value in having an RPC endpoint for it.

An endpoint for getting the block number makes a lot more sense to me. Can you share more about your use case so we can determine how to prioritize this?

dmvt:
Hey @albrow. I'd like to have the ability to confirm that Mesh and the regular ETH node connection are in sync programmatically. The network id being returned by Mesh would allow me to verify on startup that the application has not been directed to connect to a Mesh node on a different network than the RPC endpoint. Likewise, knowing the Mesh node's current block will allow the software to compare it against the RPC endpoint for the ETH node and verify that neither is behind the other (within a tolerance zone). In both cases, error states can trigger proactive notification to devops.

albrow:
Makes sense. We could also have this endpoint return other information like how many peers you are currently connected to.

dmvt:
That would be great. Any idea on ETA for this feature?

albrow:
@dmvt we're focusing on tech debt and bug fixes at the moment. It would be next week at the earliest.

fabioberger:
@dmvt working on this now.

PR #322, opened 2019-07-30
rpc: Add `mesh_getStats` JSON-RPC method

fabioberger:
Fixes: #201


core/core.go
```diff
@@ -44,6 +43,7 @@ const (
 	peerConnectTimeout         = 60 * time.Second
 	checkNewAddrInterval       = 20 * time.Second
 	expirationPollingInterval  = 50 * time.Millisecond
+	version                    = "development"
```
fabioberger: Note: We'll need to update the publish script once it merges.

albrow: We also have to keep in mind that right now the string is "development" but after the next release, it will change to e.g., "v2.0.0-beta". We might have to update our script to work for both cases.

fabioberger: Changed the script and it uses regexes now :)

PR #331, opened 2019-08-02
core/rpc: Dedup orders received via `mesh_addOrders` endpoint

fabioberger:
I noticed we weren't de-duping orders sent over JSON-RPC to the Mesh node, so I fixed this.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -539,7 +539,7 @@ func (app *App) AddOrders(signedOrdersRaw []*json.RawMessage) (*zeroex.Validatio
 		Accepted: []*zeroex.AcceptedOrderInfo{},
 		Rejected: []*zeroex.RejectedOrderInfo{},
 	}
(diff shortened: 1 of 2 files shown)


core/core.go
```diff
@@ -538,6 +539,7 @@ func (app *App) AddOrders(signedOrdersRaw []*json.RawMessage) (*zeroex.Validatio
 		Accepted: []*zeroex.AcceptedOrderInfo{},
 		Rejected: []*zeroex.RejectedOrderInfo{},
 	}
+	orderHashesSeen := map[common.Hash]interface{}{}
```
```suggestion
	orderHashesSeen := map[common.Hash]struct{}{}
```

Usually, you want to use an empty struct (`struct{}`) for the value type if you don't intend to use it. It is one of few data structures (maybe the only one?) that doesn't use any bytes of memory. The empty interface (`interface{}`) *does* use up some memory.

PR #333, opened 2019-08-02
zeroex: Treat partially fillable orders as invalid

fabioberger:
0x has this quirk where an order can be _partially fillable_. What this means is that the order maker has a sufficient balance/allowance to fill some portion of the outstanding order but _not the total remaining amount_. We used to consider any order that can be filled for a non-zero amount as valid. This PR introduces a breaking change to instead consider any order where the `fillableAmount` != `remainingAmount` to be considered invalid. 

After doing an empirical analysis of existing 0x orderbooks, the only orders that we've found in this state were mistakes. Additionally, it makes batch fill methods less effective, since the 0x smart contracts do not validate balance/allowance amounts before fills but rely on the`token transfer` reverting to detect insufficient funds/allowance. This can happed for partially filled orders and so we've decided to simply reject these orders from the onset. This change also reduces the cognitive load on devs building on 0x, since they can now rely on the `filled` mapping on the smart contract to know how much of an order is still fillable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Upcoming release
 
+### Breaking changes 🛠 
+
+- Modified Mesh's validation logic to reject and consider invalid any _partially fillable_ orders (#333)
+
 ### Bug fixes 🐞 
 
 - Added `"declaration": true,` to TS client's `tsconfig.json` so that downstream projects can use it's TS typings. (#325)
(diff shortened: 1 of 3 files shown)


core/core.go
```diff
@@ -229,11 +229,11 @@ func unquoteConfig(config Config) Config {
 }
 
 func getPubSubTopic(networkID int) string {
-	return fmt.Sprintf("/0x-orders/network/%d/version/1", networkID)
+	return fmt.Sprintf("/0x-orders/network/%d/version/2", networkID)
```
albrow: @fabioberger I wonder if we really need to bump the version number for the rendezvous string and pubsub topic? The format and types of messages we send to peers has not changed, just one detail about the semantics of which orders are considered valid.

If we don't change the version number, the worst thing that could happen is that some peers will send messages to others that they consider invalid due to the change in semantics. If you receive too many such messages, you will eventually disconnect from that peer. This will should be rare in practice and already can happen anyways due to differing order expiration buffers.

fabioberger: Thinking about it again, I agree. Also, there have only been ~10 orders that fit this edge-case on Radar right now. Let's stay on the same version and urge as many people to upgrade their nodes.

PR #335, opened 2019-08-02
meshdb: Add browser support

### LINKED ISSUE #297: Add full WebAssembly support to go-leveldb
[Partial Wasm support](https://github.com/syndtr/goleveldb/pull/277) (works in Node but not in browsers) has already been implemented. We likely will need to use [BrowserFS](https://github.com/jvilk/BrowserFS) bindings in order to achieve browser support.

albrow:
This PR adds full browser support for the `meshdb` package (and the `db` package, which it depends on). Fixes #297.

I've also added a new test which uses QUnit bindings. The test is compiled to WebAssembly and run in a headless browser as part of `go test`. We might be able to clean up and simplify the process in the future (e.g. by using a tool which automatically builds the required __index.html__ file instead of checking it into git), but it should be good enough for now.

I think our testing strategy should be to run all unit tests in Node.js because it's easier and allows us to use the same test files we already have (any file ending in `_test.go`). We only need to run high-level integration tests in the browser

This depends on two external PRs. I modified our __Gopkg.toml__ file to pull from our forks while we are waiting for these PRs to be merged.

- https://github.com/syndtr/goleveldb/pull/288 adds full support for WebAssembly in browsers and Node.js to the `goleveldb` pacakge.
- https://github.com/norunners/vert/pull/9 adds support for array types to the `norunners/vert` package, which we are using to help convert between Go and JavaScript types.

### DIFF
--- a/meshdb/browser-tests/index.html
+++ b/meshdb/browser-tests/index.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="utf-8">
+	<title>Browser/WebAssembly Tests for meshdb</title>
+	<link rel="stylesheet" href="resources/qunit-2.9.2.css">
+	<script src="resources/qunit-2.9.2.js"></script>
+	<script src="resources/browserfs-1.4.3.js"></script>
+	<script>
+		BrowserFS.configure({
+			fs: "LocalStorage",
+		}, function(e) {
+			if (e) {
+				throw e;
+			}
+		});
+		// In browsers, goleveldb expects the global variable "browserFS" to be set.
+		window.browserFS = BrowserFS.BFSRequire('fs');
+	</script>
+	<script src="resources/wasm_exec.js"></script>
+	<script>
+		QUnit.config.autostart = false;
+		const go = new Go();
+		WebAssembly.instantiateStreaming(fetch("main.wasm"), go.importObject).then((result) => {
+			go.run(result.instance);
+		});
+	</script>
+</head>
+<body>
+	<div id="qunit"></div>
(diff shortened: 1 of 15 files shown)


albrow:
Looks like the browser tests are failing in CI because it can't run `qunit-puppeteer`. I'll look further into it soon.

albrow:
Got CI to work by apt-get installing some additional dependencies.

meshdb/browser-tests/index.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
```
fabioberger: It looks like we're not super consistent in our directory naming between using hyphens and underscores. Do we want to have a mix of both?

package.json
```diff
@@ -13,5 +13,7 @@
     "isomorphic-fetch": "^2.2.1",
     "wrtc": "0.4.1"
   },
-  "dependencies": {}
+  "dependencies": {
+    "qunit-puppeteer": "^1.0.1"
```
fabioberger: Shouldn't this be a `devDependency`?

meshdb/browser-tests/index.html
albrow: Yeah we have been inconsistent here. Part of the problem is that there is an unofficial Go convention of using hyphens, but we also have things like __node_modules/__ which is using underscores.

How about we use hyphens for directory names whenever possible?

package.json
albrow: Good call. I moved it.

meshdb/browser-tests/index.html
fabioberger: Yep, I'm down.

Issue #336, opened 2019-08-05
Add support for custom topics


fabioberger:
Not all Mesh peers care about all 0x orders in existence. We want Mesh nodes to be able to specify specific topics they wish to subscribe to. A topic can represent a trading pair, an asset type, or any other sub-type of 0x orders. 

~~One idea is to implement two more JSON-RPC endpoints and modify a third:~~

~~- `mesh_joinTopic` -- Instruct the Mesh node to join a specific topic~~
~~- `mesh_leaveTopic` -- Instruct the Mesh node to leave a specific topic~~
~~- `mesh_getStats` -- Modify it to also return the topics the Mesh node is currently subscribed to~~

fabioberger:
In the end, we decided to start with a simpler implementation that allows a node to specify a single custom topic at instantiation. The node will still publish to both this custom topic and the default topic, but only receive messages from the custom topic (if specified).

PR #349, opened 2019-08-12
core/p2p/rpc: Add support for a custom topic

fabioberger:
Fixes: #336 

Add support for a single custom topic supplied via ENV VAR that is published to alongside the default topic, but that is exclusively received from.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -81,9 +81,13 @@ type Node struct {
 
 // Config contains configuration options for a Node.
 type Config struct {
-	// Topic is a unique string representing the pubsub topic. Only Nodes which
-	// have the same topic will share messages with one another.
-	Topic string
+	// DefaultTopic is the topic upon which all orders are broadcast. This is the
+	// default topic nodes share on if a custom topic is not specified at startup
+	DefaultTopic string
+	// CustomTopic is a custom topic to which the node operator wishes to subscribe to.
+	// The node will still share all orders on both the custom topic and the default topic
+	// but will only receive from the custom topic.
+	CustomTopic string
 	// TCPPort is the port on which to listen for incoming TCP connections.
 	TCPPort int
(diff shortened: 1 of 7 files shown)


p2p/node.go
```diff
@@ -71,13 +71,18 @@ type Node struct {
...
+	topics           []string
 }
 
 // Config contains configuration options for a Node.
 type Config struct {
-	// Topic is a unique string representing the pubsub topic. Only Nodes which
-	// have the same topic will share messages with one another.
-	Topic string
+	// DefaultTopic is the topic upon which all orders are broadcast. This is the
+	// default topic nodes share on if a custom topic is not specified at startup
```
albrow: I think it makes more sense in the `p2p` package to separate these into `PublishTopics` and `SubscribeTopics`. We will publish all messages to each topic in `PublishTopics` and subscribe to each topic in `SubscribeTopics`. IMO it's a better separation of concerns (the p2p package doesn't need to know about the difference between default topics and user-defined topics) and also offers us more flexibility in the future (for example if we wanted to support multiple custom topics we wouldn't need to make any changes to the `p2p` package).

rpc/clients/typescript/src/types.ts
```diff
@@ -182,7 +182,7 @@ export interface LatestBlock {
 
 export interface GetStatsResponse {
     Version: string;
-    PubSubTopic: string;
+    PubSubTopics: string[];
```
albrow: Maybe we should also separate the `GetStatsResponse` into `PublishTopics` and `SubscribeTopics` in order to provide a clearer picture of what is really going on under the hood.

p2p/node.go
fabioberger: I like this idea. Changed it to `PublishTopics` and `SubscribeTopic` (singular) since we only support a single subscribe topic for now.

albrow:
Superseded by #563.

PR #352, opened 2019-08-12
zeroex: Add order expiration buffer to zeroex.OrderValdiator

### LINKED ISSUE #341: validation: Previously Expired Orders Re-Added
### Current Behaviour
Order is added with expiration of `1,565,160,239`. It is removed from Mesh with the 10 second buffer. Another node shares the same order or it is re-posted via some mechanism. 0x Mesh emits an `ADDED` then immediately follows with an `EXPIRED` milliseconds after.

```
occuredAt    eventName    hash    uuid
1565160074981    ADDED    0xccd0cfef2afee9bbfaca688607c6e52441448a84e4d2178fe97ddc72b22b7401    3077d6a5-8252-4a8e-bc1a-293ed1035031
1565160229037    EXPIRED    0xccd0cfef2afee9bbfaca688607c6e52441448a84e4d2178fe97ddc72b22b7401    306b6ff3-80bd-49a4-8929-49ecd777f143
1565160229332    ADDED    0xccd0cfef2afee9bbfaca688607c6e52441448a84e4d2178fe97ddc72b22b7401    4736c0be-d553-4ce0-8ce5-f31e88d5728a
1565160229337    EXPIRED    0xccd0cfef2afee9bbfaca688607c6e52441448a84e4d2178fe97ddc72b22b7401    7232d6c5-3839-4982-95c7-09a72df9ac17
```

### Expected Behaviour
Once an order has been marked as `EXPIRED` an `ADDED` should not re-occur with the same Mesh node settings. An `ADDED` should not occur if it fails the Expired + Buffer offset validation. 

### Other info
The two events occur within milliseconds of each other which makes it hard in an asynchronous environment to ensure they are processed and finalised in the correct order, without introducing a queue mechanism. Whilst this change doesn't remove such an edgecase it reduces the likelihood of it occurring.

albrow:
Fixes #341.

The problem was occurring because we did not respect the expiration buffer when validating incoming messages. This PR fixes the issue and also uses the same function `IsExpired` in both the expiration watcher and the order validator to avoid these kinds of issues in the future.

### DIFF
--- a/zeroex/order_validator.go
+++ b/zeroex/order_validator.go
@@ -167,6 +167,10 @@ var (
 		Code:    "OrderHasInvalidSignature",
 		Message: "order signature must be valid",
 	}
+	ROMaxExpirationExceeded = RejectedOrderStatus{
+		Code:    "OrderMaxExpirationExceeded",
+		Message: "order expiration too far in the future",
+	}
 )
 
 // ConvertRejectOrderCodeToOrderEventKind converts an RejectOrderCode to an OrderEventKind type
@@ -193,6 +197,7 @@ type RejectedOrderKind string
 const (
 	ZeroExValidation = RejectedOrderKind("ZEROEX_VALIDATION")
 	MeshError        = RejectedOrderKind("MESH_ERROR")
+	MeshValidation   = RejectedOrderKind("MESH_VALIDATION")
 	CoordinatorError = RejectedOrderKind("COORDINATOR_ERROR")
 )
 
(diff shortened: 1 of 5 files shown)


zeroex/order_validator.go
```diff
@@ -167,6 +167,10 @@ var (
 		Code:    "OrderHasInvalidSignature",
 		Message: "order signature must be valid",
 	}
+	ROMaxExpirationExceeded = RejectedOrderStatus{
```
albrow: @fabioberger It doesn't exactly feel right to move `ROMaxExpirationExceeded` and `MeshValidation` into this package but I couldn't figure out another way to do it without introducing cyclical imports. Any ideas?

fabioberger: Agreed. We could also create a dedicated package but this works for now.

Issue #343, opened 2019-08-09
blockwatch: Remove dependency on meshdb


fabioberger:
Currently one cannot use `blockwatch` without pulling in `meshdb`. Others might want to use `blockwatch` with their own persistence solution. We should make that possible by slightly modifying the `blockwatch` package.

PR #355, opened 2019-08-13
blockwatch: Separate blockwatch dependency on meshdb 

fabioberger:
Fixes: #343 

Initial attempt at making the blockwatch package independent of the meshdb package so that any persistence layer could be used with it. This is a per-requisite to external teams being able to import the `blockwatch` package into their own projects.

TODO
- [x] Add unit tests for `blockstack.BlockStack`


ethereum/blockstack/block_stack.go
```diff
@@ -1,13 +1,15 @@
...
 
+	"github.com/0xProject/0x-mesh/ethereum/miniheader"
 	"github.com/0xProject/0x-mesh/meshdb"
 )
 
-// Stack allows performing basic stack operations on a stack of meshdb.MiniHeaders.
-type Stack struct {
+// BlockStack allows performing basic stack operations on a stack of meshdb.MiniHeaders stored
+// in the DB backing our meshdb package.
+type BlockStack struct {
```
albrow: Optionally, maybe we could call this `DBStack` to more clearly indicate how this implementation is different from `SimpleStack`?

ethereum/miniheader/miniheader.go
```diff
@@ -0,0 +1,21 @@
...
+// MiniHeader is a representation of a succinct Ethereum block headers
+type MiniHeader struct {
+	Hash   common.Hash
+	Parent common.Hash
+	Number *big.Int
+	Logs   []types.Log
+}
+
+// ID returns the MiniHeader's ID
+func (m *MiniHeader) ID() []byte {
```
albrow: Hmm.. `ID` doesn't have anything to do with Ethereum. It's something we made up for compatibility with the `db` package.

Instead of defining this here, could we include create a wrapper type with the `ID` method in the `meshdb` package? Something like:

```go
type MiniHeader ethereum.MiniHeader

func (m *MiniHeader) ID() []byte {
    return m.Hash.Bytes()
}
```

Then whenever we save a mini header in the database, we would just need to cast it to `meshdb.MiniHeader`.

ethereum/blockwatch/block_watcher.go
```diff
@@ -37,50 +37,54 @@ const (
...
+	BlockHeader *miniheader.MiniHeader
+}
+
+// Stack defines the interface a stack must implement in order to be used by
+// OrderWatcher for block header storage
+type Stack interface {
+	Pop() (*miniheader.MiniHeader, error)
+	Push(*miniheader.MiniHeader) error
+	Peek() (*miniheader.MiniHeader, error)
+	Inspect() ([]*miniheader.MiniHeader, error)
```
albrow: IIRC `Inspect` is only used in tests and debugging. I wonder if we should leave the `Inspect` method out of the interface and just upcast if we need to use it in tests?

```go
// In the tests we always know that we're using SimpleStack so we can call the Inspect method.
w.stack.(*SimpleStack).Inspect()
```

This would reduce the amount of code needed for external developers to write a `Stack` implementation, and in particular would make it easier/more efficient to implement `Stack` using a linked list.

albrow:
@fabioberger overall looks really good. I just left some suggestions about naming/organization. I could be convinced to merge as is; just want to talk about it first.

ethereum/miniheader/miniheader.go
fabioberger: I first tried to do that but it required 4 casts (not only when saving) and was generally very cumbersome. I've instead put a `// HACK` comment above it and if you feel strongly about it, I'm happy to discuss this further.

ethereum/blockwatch/block_watcher.go
fabioberger: It's actually also used in `BlockWatcher` on line 330. What I've done instead is renamed it to `PeekAll`. The Stack implementation doesn't need to be that performant since it only stored up to 20 blocks. It won't ever make sense to store much more then that.

PR #359, opened 2019-08-14
p2p: Fix public IP hack for browser environments

albrow:
https://ifconfig.me responds with different results depending on the value of the `User-Agent` header. For browser environments, it responds with HTML and for non-browser environments it responds with your IP address in plain text. Adding `/ip` to the URL results in the same behavior regardless of user agent.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -484,7 +484,7 @@ func newAddrsFactory(advertiseAddrs []ma.Multiaddr) func([]ma.Multiaddr) []ma.Mu
 }
 
 func getPublicIP() (string, error) {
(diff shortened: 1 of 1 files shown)


fabioberger:
@albrow does this return a sensible, usable response? If not, why not just skip this step if in the browser?

albrow:
> does this return a sensible, usable response? If not, why not just skip this step if in the browser?

Good point; probably not. It might be useful for WebRTC connections but I think the ICE process is already designed to do this.

Nonetheless we should still use the url with the `/ip` suffix since it's more reliable and doesn't depend on User-Agent (which could still change slightly for different platforms even for non-browser peers).

PR #360, opened 2019-08-15
cmd/mesh-bootstrap: Allow bootstrap nodes to advertise arbitrary multiaddrs

albrow:
This PR allows bootstrap nodes to advertise any multiaddress. Prior to this, they were only advertising TCP addresses of the form `/ip4/172.12.0.3/tcp/60559`. This is a problem because browser-based peers can't speak TCP and therefore had no way of connecting to the bootstrap nodes.

This PR consolidates the `P2P_LISTEN_PORT` and `PUBLIC_IP_ADDRS` environment variables into a single one called `P2P_ADDRS`. It's a comma-separated list of multiaddresses which the node will both bind to and advertise.

Importantly, this will allow us to advertise both different protocols and different addresses for each protocol. The full list of multiaddresses that will be advertised by bootstrap nodes in the near term is:

1. TCP with IPv4 address.
2. TCP with IPv6 address.
3. TCP with domain name.
4. WebSockets with IPv4 address.
5. WebSockets with IPv6 address.
6. WebSockets with domain name.

This allows maximum compatibility with peers who want to connect to the bootstrap nodes. In the future we can easily extend this list to include new protocols/transports like WebRTC and QUIC.

### DIFF
--- a/cmd/mesh-bootstrap/Dockerfile
+++ b/cmd/mesh-bootstrap/Dockerfile
@@ -26,7 +26,9 @@ WORKDIR /usr/mesh
 
 COPY --from=mesh-builder /go/src/github.com/0xProject/0x-mesh/mesh-bootstrap /usr/mesh/mesh-bootstrap
 
+# Default port for TCP multiaddr
 EXPOSE 60558
+# Default port for WebSockets multiaddr
 EXPOSE 60559
 
 RUN chmod +x ./mesh-bootstrap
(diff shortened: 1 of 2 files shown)


cmd/mesh-bootstrap/Dockerfile
```diff
@@ -26,8 +26,8 @@ WORKDIR /usr/mesh
 
 COPY --from=mesh-builder /go/src/github.com/0xProject/0x-mesh/mesh-bootstrap /usr/mesh/mesh-bootstrap
 
-ENV P2P_LISTEN_PORT=60558
 EXPOSE 60558
+EXPOSE 60559
```
fabioberger: Can we add a comment above each of these explaining what they are exposed for?

PR #365, opened 2019-08-16
zeroex: Cache order hash calculations

albrow:
Copied over from #167, this PR introduces a performance optimization by caching order hashes instead of computing them multiple times for the same order. 

This optimization has become increasingly relevant as we're testing Mesh in the browser. Performance profiling in Firefox and Chrome revealed that computing order hashes was taking a significant amount of CPU time. This PR reduces it from 2.5% total CPU time to 0.6%.


albrow:
The commit history on the `development` branch shows that this PR was actually merged, but GitHub seems to be confused. I'm going to close it here.

Screen Shot 2019-08-20 at 11 16 03 AM

PR #366, opened 2019-08-16
Update all configuration options in preparation for browser support

albrow:
This PR changes a variety of configuration options throughout the code in preparation for full browser support. It also introduces a functioning MVP example of Mesh running in the browser in the __examples/__ directory and enables seamless headless browser tests in CircleCI and as part of `make test-all`. This allows us to remove the more manual qunit-based tests from the `meshdb` package and just run all our tests normally.

The in-browser example currently comes with the following caveats:

1. The only data that is stored permanently is the private key. This allows browser peers to retain the same Peer ID between refreshes, but all other data (orders, block headers, and peers) will be lost.
2. Browser peers are undialable cannot connect directly to other browser peers. They can be connected to through a relay. However, they can dial any peer that supports the WebSockets transport.
3. There is no JavaScript API for interacting with Mesh and therefore no way to add orders or receive order updates. The only way we can see what Mesh is doing is by looking at the logs.
4. We still need to do some work to optimize performance and reduce bundle size. You might see your CPU usage spike while running the example.

To run the browser example yourself, you need a web server that supports the Wasm MIME type. An easy option is to install [`goexec`](https://github.com/shurcooL/goexec). Then run the following commands from the root of the repository:

```
> GOOS=js GOARCH=wasm go build -o ./examples/browser/main.wasm ./examples/browser
> goexec 'http.ListenAndServe(":8000", http.FileServer(http.Dir("./examples/browser")))'
```

Finally visit http://localhost:8000 in your browser and open up the developer console to observe Mesh running 😄


ethereum/blockwatch/block_watcher_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !browser
```
albrow: This build tag means the test won't run in browser environments. It wouldn't work because we try to load a JSON file.

PR #369, opened 2019-08-21
Add TypeScript bindings

### LINKED ISSUE #298: Implement TypeScript/JavaScript bindings for interacting with Mesh from the browser
After we have all the other components in place and Mesh is runnable in the browser, we will need to publish TypeScript/JavaScript bindings which can be used to interact with Mesh in browser environments. These bindings will be analogous to the JSON-RPC API and should contain basically all of the same methods/interfaces.

albrow:
Fixes #298.

This PR adds TypeScript bindings and an npm package for using Mesh in the browser. The Wasm bytecode is embedded in the package itself which makes it really easy to import and use. It also adds an example of how to import and use the TypeScript bindings.

Here's a breakdown of the directory structure and key files:

- __browser/__ contains TypeScript bindings for running Mesh in the browser. This directory is a publishable npm package called `@0x/mesh-browser`.
  - __go/__ The source code for the Wasm bytecode that will be embedded in the package. It imports `0x-mesh/core` and adds everything we need to the global scope.
  - __scripts/__ contains one-off scripts related to the TypeScript bindings.
    - __generate_wasm_buffer.go__ a script that generates a TypeScript file containing the Wasm bytecode as a base64-encoded `ArrayBuffer`.
  - __ts/__ the entry point for the npm package. This is a lightweight wrapper around the variables that we injected into global scope that adds some extra niceties.
    - __generated/__ contains the TypeScript file generated by __scripts/generate_wasm_buffer.go__.
  - __wasm/__ contains the Wasm bytecode itself (the result of compiling __go/main.go__).
  - __lib/__ contains the JavaScript code, source maps, and type definitions (the result of compiling __ts/__).
- __examples/browser__ is an example of how to import and use the `@0x/mesh-browser` package in a browser-based application. It uses TypeScript and webpack.
  - __src/__ is the TypeScript code for the example application.
  - __dist/__ is where webpack outputs the results. Everything is compiled into a single __bundle.js__ file.

I'm planning to clean this up and document it in a README, but for now here are the steps for running the example:

1. In the __browser/__ directory run `yarn install && yarn build`.
2. In the __examples/browser__ directory run `yarn install && yarn build`.
3. In the root of the repository, run `goexec 'http.ListenAndServe(":8000", http.FileServer(http.Dir("./examples/browser/dist")))'`
4. Visit http://localhost:8000 in your browser and look at the console output.


albrow:
Latest commit adds a `setOrderEventsHandler` method, which is analogous to the `mesh_subscribe("orders")` JSON RPC endpoint.

albrow:
@fabioberger I removed the WIP tag and this PR is ready for review. I still expect that there will be some discussion about naming, interface design, and project structure.

browser/package.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "@0x/mesh-browser",
```
albrow: @fabioberger what do you think of this name? I also considered just calling it `@0x/mesh` but I want to consider the possibility that we might have a separate, slightly modified version designed to work in Node.js in the future.

browser/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "version": "3.0.1-beta",
```
albrow: My preference would be to set this version to `development` in the development branch, but unfortunately that doesn't really work with Yarn/NPM. Okay with just setting it to the last released version?

browser/ts/index.ts
```diff
@@ -0,0 +1,370 @@
...
+    class Go {
+        run(instance: WebAssembly.Instance): void;
+        importObject: any;
+    }
+
+    // Defined in ../go/main.go
+    const zeroExMesh: ZeroExMesh;
+}
+
+/**
```
albrow: @fabioberger We should probably generate docs for the TypeScript bindings similar to what we did for the TypeScript RPC client. Can you point me in the right direction? Alternatively we could do that in a separate PR since this is already pretty big.

albrow:
I know there's not a lot of tests here. I'm planning to write integration tests in a separate PR. The way it will work is that we'll spin up a few nodes on a local network and confirm that they can connect and share orders with one another. It will involve coordinating between some native Go nodes and some that are running in a headless browser.

examples/browser/package.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "devDependencies": {
+        "rimraf": "^3.0.0",
+        "ts-loader": "^6.0.4",
+        "typescript": "^3.5.3",
+        "webpack": "^4.39.2",
+        "webpack-cli": "^3.3.7"
+    },
+    "dependencies": {
+        "@0x/mesh-browser": "./../../browser/"
```
albrow: I'll update this to use the actual version once we have published the TypeScript bindings.

browser/go/main.go
```diff
@@ -0,0 +1,267 @@
...
+		Verbosity:                   2,
+		DataDir:                     "0x-mesh",
+		P2PTCPPort:                  0,
+		P2PWebSocketsPort:           0,
+		UseBootstrapList:            true,
+		BlockPollingInterval:        5 * time.Second,
+		EthereumRPCMaxContentLength: 524288,
+	}
+
+	// Required conig options
```
fabioberger:
```suggestion
	// Required config options
```

browser/package.json
fabioberger: I like it. Let's stick with it.

browser/package.json
fabioberger: Yep, that's fine.

browser/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "description": "TypeScript and JavaScript bindings for running Mesh directly in the browser.",
+    "main": "./lib/index.js",
+    "license": "MIT",
```
fabioberger: Not Apache 2 like all our other code?

browser/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "private": true,
```
fabioberger: It's not private since we publish it to NPM.

browser/package.json
```diff
@@ -0,0 +1,22 @@
...
+        "build": "yarn build:go && yarn build:generate && yarn build:ts",
+        "build:ts": "node --max_old_space_size=4096 ./node_modules/.bin/tsc -b",
+        "build:generate": "go run ./scripts/generate_wasm_buffer.go",
+        "build:go": "GOOS=js GOARCH=wasm go build -o ./wasm/main.wasm ./go/main.go"
+    },
+    "dependencies": {
+        "@0x/order-utils": "^8.3.0",
+        "@0x/utils": "^4.5.0",
+        "@types/base64-arraybuffer": "^0.1.0",
+        "@types/node": "^12.7.2",
```
fabioberger: Should these types be devDeps? Also, we typically set `@types/node` to `*` since there is a high chance that the end user is using a different version.

browser/package.json
```diff
@@ -0,0 +1,22 @@
...
+        "base64-arraybuffer": "^0.2.0",
+        "typescript": "^3.5.3"
```
fabioberger: This is def. a `devDep`

browser/ts/index.ts
fabioberger: You'll need `"@0x/ts-doc-gen": "^0.0.16",` dep and a command:

```
"scripts": {
    "docs:md": "ts-doc-gen --sourceDir=./src --output=${npm_package_config_docsPath}"
},
"config": {
    "docsPath": "../../../docs/json_rpc_clients/typescript"
},
```

Just change the `docsPath` to where you want the MD generated to and it should work 👍

browser/ts/index.ts
```diff
@@ -0,0 +1,370 @@
...
+    blockPollingIntervalSeconds?: number;
+    // The maximum request Content-Length accepted by the backing Ethereum RPC
+    // endpoint used by Mesh. Geth & Infura both limit a request's content
+    // length to 1024 * 512 Bytes. Parity and Alchemy have much higher limits.
+    // When batch validating 0x orders, we will fit as many orders into a
+    // request without crossing the max content length. The default value is
+    // appropriate for operators using Geth or Infura. If using Alchemy or
+    // Parity, feel free to double the default max in order to reduce the number
+    // of RPC calls made by Mesh. Defaults to 524288 bytes.
+    ethereumRPCMaxContentLength?: number;
```
fabioberger: How will we remember to update these in both places? Can we perhaps add a comment/warning in both places mentioning the need to edit the other?

browser/ts/index.ts
```diff
@@ -0,0 +1,370 @@
...
+// analog isn't defined here. Instead we re-use the definition in
+// @0x/order-utils.
+interface MeshSignedOrder {
+    makerAddress: string;
+    makerAssetData: string;
+    makerAssetAmount: string;
+    makerFee: string;
+    takerAddress: string;
+    takerAssetData: string;
+    takerAssetAmount: string;
```
fabioberger: Why aren't amounts parsed into BigNumbers?

browser/ts/index.ts
```diff
@@ -0,0 +1,370 @@
...
+    isNew: boolean;
+}
+
+// The type for rejected orders exposed by MeshWrapper.
+interface MeshRejectedOrderInfo {
+    orderHash: string;
+    signedOrder: MeshSignedOrder;
+    kind: RejectedOrderKind;
+    status: RejectedOrderStatus;
+}
```
fabioberger: These are all duplicated from `@0x/mesh-rpc-client`. Any ideas on how to avoid that? I can't think of any without creating a TS monorepo within this repo...

browser/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "version": "3.0.1-beta",
+    "description": "TypeScript and JavaScript bindings for running Mesh directly in the browser.",
+    "main": "./lib/index.js",
+    "license": "MIT",
+    "private": true,
+    "scripts": {
+        "build": "yarn build:go && yarn build:generate && yarn build:ts",
+        "build:ts": "node --max_old_space_size=4096 ./node_modules/.bin/tsc -b",
+        "build:generate": "go run ./scripts/generate_wasm_buffer.go",
+        "build:go": "GOOS=js GOARCH=wasm go build -o ./wasm/main.wasm ./go/main.go"
```
fabioberger: Can we add `tslint`? See a bunch of things that the linter would have caught.

browser/ts/index.ts
```diff
@@ -0,0 +1,370 @@
...
+        const meshResults = await this._wrapper.addOrdersAsync(meshOrders);
+        return meshValidationResultsToValidationResults(meshResults);
+    }
+}
+
+function sleepAsync(ms: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+function meshSignedOrderToSignedOrder(meshSignedOrder: MeshSignedOrder): SignedOrder {
```
fabioberger: Why you're calling `MeshSignedOrder` we call `StringifiedSignedOrder` in `mesh-rpc-client`. Thoughts on standardizing?

examples/browser/src/index.ts
```diff
@@ -0,0 +1,56 @@
+import { Mesh, OrderEvent, SignedOrder, BigNumber } from '@0x/mesh-browser';
+
+(async () => {
+    // Confingure Mesh to use mainnet and Infura.
```
fabioberger:
```suggestion
    // Configure Mesh to use mainnet and Infura.
```

examples/browser/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    const mesh = new Mesh({
+        ethereumRPCURL: 'https://mainnet.infura.io/v3/af2e590be00f463fbfd0b546784065ad',
```
fabioberger: You sure you want to leave this in here?

browser/go/main.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/ethereum/go-ethereum/event"
+)
+
+const (
+	// loadEventName is the name of a global event that will be fired after all
+	// WebAssembly is done loading.
+	loadEventName = "0xmeshload"
+	// orderEventsBufferSize is the buffer size for the orderEvents channel. If
+	// the buffer is full, any additional events won't be processed.
+	orderEventsBufferSize = 100
```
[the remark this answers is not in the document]
albrow: Increasing `orderEventsBufferSize` reduces the chance of order events being missed but increases memory usage. I think 100 is a decent starting point but we can adjust it if needed. The channel should never get that full if an `orderEventsHandler` is set.

browser/package.json
albrow: This was just the default. I changed it to `Apache-2.0`.

browser/package.json
albrow: Yeah I just set it to private so I didn't accidentally publish it early. I'll change it back.

browser/package.json
albrow: Realized we don't need `@types/node` anymore so I just removed it. Also moved `typescript` and `@types/base64-arraybuffer` to `devDependencies`.

browser/ts/index.ts
albrow: 👍@fabioberger just added this. Mind eyeballing the docs to make sure everything looks right?

browser/ts/index.ts
albrow: I'll add a comment for now. We can potentially generate code in both places from a canonical source (e.g. JSON) in the future.

browser/package.json
albrow: 👍just added `tslint`.

examples/browser/src/index.ts
albrow: 🤷‍♂It's probably fine. I'm not using that key for anything at the moment. It has the advantage of making the example easily runnable.

browser/ts/index.ts
albrow: I think I made this pretty clear in the comments for `MeshWrapper` and `MeshSignedOrder`:

```ts
// A direct translation of the MeshWrapper type in Go. Its API exposes only
// simple JavaScript types like number and string, some of which will be
// converted. For example, we will convert some strings to BigNumbers.
interface MeshWrapper
```

```ts
// The type for signed orders exposed by MeshWrapper. Unlike other types, the
// analog isn't defined here. Instead we re-use the definition in
// @0x/order-utils.
interface MeshSignedOrder
```

`MeshSignedOrder` describes the type of something that comes directly from Go/Wasm. We can't (easily) convert strings to `BigNumber` in Go/Wasm so instead we do it here. The types that are actually exposed by the `@0x/mesh-browser` package are `BigNumber`s where appropriate. Take a look at the `meshSignedOrderToSignedOrder` function for an example of where the conversion happens.

Do you feel there is something missing from the comments or is there anything I can do to help make this more clear?

browser/ts/index.ts
albrow: I can't think of anything else either. I'm not sure if it's worth the overhead of using a TS monorepo here. Let's copy it manually for now and discuss later?

browser/ts/index.ts
albrow: Hmm... `StringifiedSignedOrder` is an okay name in this case because all the types are strings. However, I want to pick a name that highlights the pattern of converting between Go types and TypeScript types. I'm thinking that there will be other type conversions in the future and not just `string`  to `BigNumber`. I can't think of a good name to cover the use cases in both `mesh-rpc-client` and `mesh-browser`. Maybe something like `BasicSignedOrder`?

browser/ts/index.ts
fabioberger: The comment above `MeshWrapper` is clear to me. Given the "it's API exposes only simple Javascript types..." part, it wasn't immediately obvious to me that this included `MeshSignedOrder`. But now it's clear. I think we can probably just leave it, I must have skimmed through it too quickly.

browser/ts/index.ts
fabioberger: My vote would perhaps be `StringifiedSignedOrder` for now, and if there are other types involved in the future, changing it then. Don't feel too strongly about this though, so we can also leave it for now.

albrow: I'm going to go with `WrapperSignedOrder`, `WrapperOrderEvent`, `WrapperValidationResults`, etc. to more closely match the fact that these are types exposed by `MeshWrapper` that might need further conversion. This is what I was originally trying to show with the `Mesh*` prefix but I feel that name was too confusing. Happy to re-evaluate later.

PR #376, opened 2019-08-28
Add integration tests

albrow:
Still needs to be cleaned up a bit. Opening PR now so I can make sure it works in CI.


albrow:
@fabioberger no longer WIP. This PR is ready for review.

Makefile
```diff
@@ -42,6 +42,11 @@ test-go:
 	go test ./... -race -timeout 30s
 
 
+.PHONY: test-integration
+test-integration:
+	go test ./integration-tests -timeout 185s --integration
```
fabioberger: 185s? How long do we expect them to take?

integration-tests/browser/src/index.ts
```diff
@@ -0,0 +1,66 @@
+import { Mesh, OrderEvent, SignedOrder, BigNumber } from '@0x/mesh-browser';
+
+(async () => {
+    // Configure Mesh to use our local Ganache instance and network ID 51
```
fabioberger:
```suggestion
    // Configure Mesh to use our local Ganache instance and network ID 50
```

integration-tests/browser/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    mesh.onOrderEvents((events: Array<OrderEvent>) => {
+        for (let event of events) {
+            // console.log('received order event: ' + JSON.stringify(event));
+        }
+    });
+
+    // Start Mesh *after* we set up the handlers.
+    await mesh.startAsync();
+
+    // This order is for demonstration purposes only and is invalid. It will be
```
fabioberger: Please add to this comment _how_ it is invalid. Just nice to know.

Makefile
fabioberger: Ah yep, ran into that with webpack before. Really surprising it's not more optimized.

PR #384, opened 2019-09-03
ethereum: Use canonical 0x Ganache snapshot

fabioberger:
This PR replaces our custom Mesh snapshot with the canonical 0x snapshot. Doing so also allows us to use the `DevUtils` smart contract instead of `EthBalanceCheck` and `OrderValidationUtils` (which will soon no longer be deployed stand-alone).

Our custom snapshot also set maker/taker allowance/balance for ZRX/WETH (assets involved in the test order), so this PR also adds logic to setup the test scenario in the test.


fabioberger:
Tests failing because the same scenario setup logic needs to be added to the integration tests. Will do this once #382 lands.

scenario/scenario.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/stretchr/testify/require"
+)
+
+var eighteenDecimalsInBaseUnits = new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
+var wethAmount = new(big.Int).Mul(big.NewInt(50), eighteenDecimalsInBaseUnits)
+var zrxAmount = new(big.Int).Mul(big.NewInt(100), eighteenDecimalsInBaseUnits)
+
+// SetupBalancesAndAllowances sets up the proper balance/allowance for the maker/taker of
+// ZRX/WETH respectively so that the created orders are fillable.
```
albrow: Can we make `wethAmount` and `zrxAmount` parameters for this function so that it is more generally useful? As written, it's also not clear which orders you're referring to and what it sets the balances and allowances to exactly.

Issue #190, opened 2019-06-25
orderwatch: Missing updates to order `lastUpdated` in DB


fabioberger:
Currently the `lastUpdated` timestamp is only updated when a new order is added, becomes unfillable or it's fillability increases up from 0. If should also be updated when it is partially filled or it's fillability increases up from > 0.

PR #386, opened 2019-09-03
orderwatch: Update lastUpdated and fillableTakerAssetAmount for order in DB

fabioberger:
Fixes: #190 

Currently, we aren't updating an orders `fillableTakerAssetAmount` nor it's `lastUpdated` timestamp in the DB when processing a contract event impacting it _unless_ it is being unwatched or re-watched. This PR makes it such that we always update an order's DB entry, no matter what the event so that the `lastUpdated` and `fillableTakerAssetAmount` in the DB is always accurate.

### DIFF
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -613,10 +613,10 @@ func (w *Watcher) generateOrderEventsIfChanged(hashToOrderWithTxHashes map[commo
 			orderEvents = append(orderEvents, orderEvent)
 		} else if oldFillableAmount.Cmp(newFillableAmount) == 0 {
 			// No important state-change happened, simply update lastUpdated timestamp in DB
-			w.updateOrderLastUpdatedAndFillableAmount(order)
+			w.updateOrderDBEntry(order)
 		} else if oldFillableAmount.Cmp(big.NewInt(0)) == 1 && oldAmountIsMoreThenNewAmount {
 			// Order was filled, emit  event and update order in DB
-			w.updateOrderLastUpdatedAndFillableAmount(order)
+			w.updateOrderDBEntry(order)
 			orderEvent := &zeroex.OrderEvent{
 				OrderHash:                acceptedOrderInfo.OrderHash,
 				SignedOrder:              order.SignedOrder,
@@ -628,7 +628,7 @@ func (w *Watcher) generateOrderEventsIfChanged(hashToOrderWithTxHashes map[commo
 		} else if oldFillableAmount.Cmp(big.NewInt(0)) == 1 && !oldAmountIsMoreThenNewAmount {
 			// The order is now fillable for more then it was before. E.g.: A fill txn reverted (block-reorg)
 			// Update order in DB and emit event
-			w.updateOrderLastUpdatedAndFillableAmount(order)
+			w.updateOrderDBEntry(order)
 			orderEvent := &zeroex.OrderEvent{
 				OrderHash:                acceptedOrderInfo.OrderHash,
 				SignedOrder:              order.SignedOrder,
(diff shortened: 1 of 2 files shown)


zeroex/orderwatch/order_watcher.go
```diff
@@ -607,10 +607,11 @@ func (w *Watcher) generateOrderEventsIfChanged(hashToOrderWithTxHashes map[commo
...
 			orderEvents = append(orderEvents, orderEvent)
 		} else if oldFillableAmount.Cmp(newFillableAmount) == 0 {
-			// No important state-change happened, ignore
-			// Noop
+			// No important state-change happened, simply update lastUpdated timestamp in DB
+			w.updateOrderDBEntry(order)
 		} else if oldFillableAmount.Cmp(big.NewInt(0)) == 1 && oldAmountIsMoreThenNewAmount {
-			// Order was filled, emit  event
+			// Order was filled, emit  event and update order in DB
+			w.updateOrderDBEntry(order)
```
albrow: @fabioberger I think we need to do this inside of a transaction, otherwise we run the risk of hitting a write-after-read data race. We are trying to update individual fields of the order, but if a separate goroutine is trying to update a field at the same time, one of the changes will be overwritten. So for example:

1. Goroutine-0 reads the order from the database.
2. Goroutine-1 reads the order from the database.
3. Goroutine-0 changes the `FillableTakerAssetAmount` field on its local view of the order.
4. Goroutine-1 changes the `LastUpdated` field on its local view of the order.
5. Goroutine-0 saves the order to the database, including the changes to `FillableTakerAssetAmount`.
6. Goroutine-1 saves the order to the database, overwriting the changes to `FillableTakerAssetAmount`.

This happens because each goroutine has its own local view of the order and is not aware of the changes potentially made by other goroutines.

Unfortunately the only way to avoid this right now is with a lock on the entire `Orders` collection, but hopefully we can take care to hold the lock for as little time as possible. It should be good enough for now. In the future, the `db` package could support (1) locks on individual models instead of entire columns or (2) updating specific fields instead of updating the entire model at once. Either one of those new features would offer a more efficient way to solve the problem. Another way is by using optimistic locking (this is how I [solved the problem in Zoom](https://github.com/albrow/zoom/tree/v0.18.0#concurrent-updates-and-optimistic-locking)).

fabioberger: Thanks for finding this issue @albrow. That makes a lot of sense. Looks like we'll need to put the entirety of `handleBlockEvents` within a Orders collection lock. How do I go about acquiring a lock on the Orders collection specifically? I only see the more generic `colLock` and the `globalWriteLock`.

albrow: Use [`Collection.OpenTransaction`](https://godoc.org/github.com/0xProject/0x-mesh/db#Collection.OpenTransaction).

fabioberger: Added the transaction.

zeroex/orderwatch/order_watcher.go
```diff
@@ -565,6 +565,7 @@ func (w *Watcher) findOrders(makerAddress, tokenAddress common.Address, tokenID
 }
 
 func (w *Watcher) generateOrderEventsIfChanged(hashToOrderWithTxHashes map[common.Hash]*OrderWithTxHashes) error {
+	ordersColTxn := w.meshDB.Orders.OpenTransaction()
```
albrow: As I mentioned in Slack, it's very important that we read from the database *after* opening the transaction in order to avoid the data race I described above. Otherwise, the order can get modified between when it was read from the database and when the transaction was opened. I think we can address this problem by not reading the order from the database until we are inside the `generateOrderEventsIfChanged` method.

albrow: It's also really important to make sure that the transaction is always either committed or discarded. Right now there are some cases where we return an error before committing the transaction, and that means the orders collection would be locked forever and no writes could be made. You can fix this by adding a `defer` statement:

```go
defer func() {
    _ = ordersColTxn.Discard()
}()
```

fabioberger: Good catch. Moved the transaction higher up so that it now encompasses all the reads, as well as the writes. I also added the defer.

albrow:
@fabioberger I think you meant to pass the transaction through to `generateOrderEventsIfChanged`?

PR #389, opened 2019-09-04
p2p: Add ability to ban IP addresses

### LINKED ISSUE #387: Add ability to ban/blacklist certain IP addresses.
Related to #119. In order to implement rate-limiting, we need a way to ban the IP addresses of peers that are sending too much data as they are detected.

albrow:
Fixes #387.

I originally thought we should create our own `ConnectionManager` to implement this feature, but after digging more into the libp2p code, I don't think that will work. `ConnectionManager` is only responsible for pruning existing connections and doesn't really have the ability to prevent connections from being established in the first place. There is also an incentive to do this at the lowest abstraction level possible to minimize the resources that a peer with a banned IP address can use. Luckily, what we need is basically already implemented in the `swarm` package (see [`Swarm.Filters`](https://godoc.org/github.com/libp2p/go-libp2p-swarm#Swarm)).

Unfortunately `libp2p.New` does not currently expose any way to configure or access the underlying swarm. To workaround this issue, I just copied the code from `libp2p.New` and modified to expose the swarm.

This PR also automatically protects the bootstrap nodes from being banned.

### DIFF
--- a/integration-tests/browser/package.json
+++ b/integration-tests/browser/package.json
@@ -5,7 +5,7 @@
     "license": "Apache-2.0",
     "private": true,
     "scripts": {
-        "build": "node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js",
+        "build": "node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --mode=development",
         "postinstall": "yarn rimraf ./node_modules/@0x/mesh-browser/go ./node_modules/@0x/mesh-browser/scripts"
     },
     "devDependencies": {
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -3,6 +3,11 @@
 package p2p
 
 import (
+	"net"
+	"sync"
+
+	"github.com/albrow/stringset"
+
 	"context"
 	"crypto/rand"
 	"errors"
@@ -22,6 +27,7 @@ import (
 	dht "github.com/libp2p/go-libp2p-kad-dht"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	swarm "github.com/libp2p/go-libp2p-swarm"
+	filter "github.com/libp2p/go-maddr-filter"
 	ma "github.com/multiformats/go-multiaddr"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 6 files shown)


integration-tests/browser/package.json
```diff
@@ -5,7 +5,7 @@
     "license": "Apache-2.0",
     "private": true,
     "scripts": {
-        "build": "node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js",
+        "build": "node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --mode=development",
```
albrow: I was seeing the integration tests occasionally timeout in CI and this seems to help get them passing.

p2p/node.go
```diff
@@ -287,6 +306,63 @@ func (n *Node) Connect(peerInfo peer.AddrInfo, timeout time.Duration) error {
...
+	return nil
+}
+
+// BanIP adds the IP address of the given Multiaddr to the blacklist. The
+// node will no longer dial or accept connections from this IP address. However,
+// if the IP address is protected, calling BanIP is a no-op. BanIP does not
+// automatically disconnect from the given multiaddress if there is currently an
+// open connection.
+func (n *Node) BanIP(maddr ma.Multiaddr) error {
+	ipNet, err := ipNetFromMaddr(maddr)
```
fabioberger: So we return an error if the `multiAddr` doesn't include an IP address? I guess the caller would then check for that specific error and then noop for now?

p2p/node.go
```diff
@@ -287,6 +306,63 @@ func (n *Node) Connect(peerInfo peer.AddrInfo, timeout time.Duration) error {
...
+		return err
+	}
+	n.unbanIPNet(ipNet)
+	return nil
+}
+
+func (n *Node) unbanIPNet(ipNet net.IPNet) {
+	// There is no guarantee in the public API of the filters package that would
+	// prevent multiple filters being added for the same IPNet (though it
+	// shoudln't happen in practice). We use a for loop here to make sure we
```
fabioberger:
```suggestion
	// shouldn't happen in practice). We use a for loop here to make sure we
```

p2p/node.go
```diff
@@ -454,3 +530,57 @@ func (n *Node) receive(ctx context.Context) (*Message, error) {
...
+	ma.ForEach(maddr, func(c ma.Component) bool {
+		switch c.Protocol().Code {
+		case ma.P_IP6ZONE:
+			return true
+		case ma.P_IP6, ma.P_IP4:
+			found = true
+			ip = net.IP(c.RawValue())
+			return false
+		default:
+			return false
```
fabioberger: What does returning `true` vs. `false` do here?

p2p/node.go
albrow: That's pretty much right. But there are at least some cases where the caller should not no-op. It's potentially a big problem if you try to ban a misbehaving IP address and it doesn't work. In any case it feels dangerous to silently fail here.

p2p/node.go
albrow: Returning false causes `ma.ForEach` to stop iterating through components of the multiaddress. Returning true will cause it to keep iterating.

PR #408, opened 2019-09-12
WIP: Remove oldest orders when database is full

### LINKED ISSUE #373: core: Delete random orders when DB is full
If the DB is full, the simplest thing we can do is remove random orders to make space for incoming orders. We will start there and then eventually replace this with a more robust solution.

albrow:
Fixes #373.

Still WIP due to missing tests and some outstanding issues to discuss.


albrow:
Closing this since we are going to go with a different approach.

PR #415, opened 2019-09-14
Refactor integration tests using `scenario`

fabioberger:
Currently `scenario` exposes the following interface:

```golang
func SetupBalancesAndAllowances(t *testing.T, makerAddress, takerAddress common.Address, wethAmount *big.Int, zrxAmount *big.Int) {
```

It assumes the caller has already created a suitable `SignedOrder` and would call this method to setup the requisite balances/allowances to make it fillable. This loose coupling of the `SignedOrder` and it's state is error-prone and while writing the `OrderWatcher` tests, I've found myself needing an order where the `makerAsset` is `WETH` and `takerAddress` `ZRX` (the reverse of the assumptions in the above method). This forced me to re-evaluate this interface. 

The reason we did not expose something like:

```golang
func CreateZRXForWETHSignedTestOrder(t *testing.T, makerAddress, takerAddress common.Address, wethAmount *big.Int, zrxAmount *big.Int) *zeroex.SignedOrder {
```

from the get-go is because referencing `zeroex` introduced a circular dependency when importing `scenario` into the `OrderValidator` tests (these were in the `zeroex` pkg). This PR moves `OrderValidator` and it's tests out into it's own package `ordervalidate` resolving the circular dep, and updates `scenario`'s interface to the above. This removes the loose coupling and we can then add further methods that set up additional orders (e.g., `CreateWETHForZRXSignedTestOrder`).

Lastly, this PR also updates the integration tests so that we don't use the ZRX coinbase address (with 1 billion ZRX balance) as the order maker.

### DIFF
--- a/cmd/mesh/rpc_handler.go
+++ b/cmd/mesh/rpc_handler.go
@@ -14,6 +14,7 @@ import (
 	"github.com/0xProject/0x-mesh/core"
 	"github.com/0xProject/0x-mesh/rpc"
 	"github.com/0xProject/0x-mesh/zeroex"
+	"github.com/0xProject/0x-mesh/zeroex/ordervalidate"
 	ethRpc "github.com/ethereum/go-ethereum/rpc"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	log "github.com/sirupsen/logrus"
@@ -70,7 +71,7 @@ func (handler *rpcHandler) GetOrders(page, perPage int, snapshotID string) (*rpc
 }
 
 // AddOrders is called when an RPC client calls AddOrders.
-func (handler *rpcHandler) AddOrders(signedOrdersRaw []*json.RawMessage) (*zeroex.ValidationResults, error) {
+func (handler *rpcHandler) AddOrders(signedOrdersRaw []*json.RawMessage) (*ordervalidate.ValidationResults, error) {
 	log.WithField("count", len(signedOrdersRaw)).Debug("received AddOrders request via RPC")
 	validationResults, err := handler.app.AddOrders(signedOrdersRaw)
 	if err != nil {
--- a/cmd/mesh/rpc_handler.go
+++ b/cmd/mesh/rpc_handler.go
@@ -14,7 +14,7 @@ import (
 	"github.com/0xProject/0x-mesh/core"
 	"github.com/0xProject/0x-mesh/rpc"
 	"github.com/0xProject/0x-mesh/zeroex"
-	"github.com/0xProject/0x-mesh/zeroex/ordervalidate"
+	"github.com/0xProject/0x-mesh/zeroex/ordervalidator"
 	ethRpc "github.com/ethereum/go-ethereum/rpc"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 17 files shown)


zeroex/ordervalidate/order_validator_test.go
```diff
@@ -136,24 +128,48 @@ func TestBatchValidateOffChainCases(t *testing.T) {
...
+		signedOrders := []*zeroex.SignedOrder{
 			&testCase.SignedOrder,
 		}
 
 		orderValidator, err := NewOrderValidator(ethClient, constants.TestNetworkID, constants.TestMaxContentLength, 0)
 		require.NoError(t, err)
 
-		validationResults := orderValidator.BatchValidate(signedOrders, areNewOrders)
-		isValid := len(validationResults.Accepted) == 1
+		offchainValidOrders, rejectedOrderInfos := orderValidator.BatchOffchainValidation(signedOrders)
```
fabioberger: Refactored these tests to use the `BatchOffChainValidation` method directly, since that's what it's actually supposed to be testing.

cmd/mesh/rpc_handler.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/0xProject/0x-mesh/core"
 	"github.com/0xProject/0x-mesh/rpc"
 	"github.com/0xProject/0x-mesh/zeroex"
+	"github.com/0xProject/0x-mesh/zeroex/ordervalidate"
```
albrow: Package names are almost always nouns, not verbs. Can we rename this to `ordervalidator`?

core/core.go
```diff
@@ -179,7 +180,7 @@ func New(config Config) (*App, error) {
 	blockWatcher := blockwatch.New(blockWatcherConfig)
 
 	// Initialize the order validator
-	orderValidator, err := zeroex.NewOrderValidator(ethClient, config.EthereumNetworkID, config.EthereumRPCMaxContentLength, config.OrderExpirationBuffer)
+	orderValidator, err := ordervalidate.NewOrderValidator(ethClient, config.EthereumNetworkID, config.EthereumRPCMaxContentLength, config.OrderExpirationBuffer)
```
albrow: If we rename the package to `ordervalidator`, convention would be to call this method `ordervalidator.New`.

zeroex/ordervalidate/order_validator.go
```diff
@@ -808,33 +809,33 @@ func (o *OrderValidator) computeOptimalChunkSizes(signedOrders []*SignedOrder) [
 }
 
 func isSupportedSignature(signature []byte, orderHash common.Hash) bool {
-	signatureType := SignatureType(signature[len(signature)-1])
+	signatureType := zeroex.SignatureType(signature[len(signature)-1])
 
 	switch signatureType {
-	case IllegalSignature:
-	case InvalidSignature:
+	case zeroex.IllegalSignature:
```
albrow: Sorry I didn't notice this before. In Go, switch cases do not implicitly fall through. As it is currently written `case zeroex.IllegalSignature` doesn't have any associated statements which means "break the switch statement and do nothing". You probably want to write it this way:

```go
case zeroex.IllegalSignature, zeroex.InvalidSignature:
    return false
```

See https://tour.golang.org/flowcontrol/9.

We can address this in a separate PR if you prefer.

fabioberger: Thanks for catching. Since it did not impact the actual logic, I decided to fix it here.

albrow: This *does* impact the actual logic. In the old code, the `IllegalSignature` case would break the switch and cause the function to return true. Maybe you're saying that it ultimately didn't affect anything because the signature is redundantly validated by the smart contract?

fabioberger: Yes, that's what I meant. It didn't actually cause the validation to misbehave from an end-users perspective.

Issue #405, opened 2019-09-09
orderwatch: Add better test coverage


fabioberger:
Currently we are testing every individual sub-component of `OrderWatcher` but not the watcher itself. 

In order to test it in a sane way, we'd want to be able to pass it decoded logs that could be hand-written. The first step of completing this issue is to refactor it to accept decoded logs. After that, we just need to write tests to cover all edge-cases.

PR #416, opened 2019-09-14
orderwatch: Add missing test coverage

fabioberger:
Fixes: #405 

WARNING: Don't merge. Currently opened against #415. Will open against `development` once #415 lands.

This PR adds test coverage of the OrderWatcher's functionality of properly processing block events and updating the orders in the DB accordingly. In order to do this, we needed to trigger each of the events we listen for and then check to make sure it was picked up and processed properly.


integration-tests/integration_test.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/0xProject/0x-mesh/zeroex"
 	"github.com/chromedp/cdproto/runtime"
 	"github.com/chromedp/chromedp"
+	"github.com/ethereum/go-ethereum/ethclient"
+	ethRpc "github.com/ethereum/go-ethereum/rpc"
```
albrow: nit: we've been using the non-camel-cased `ethrpc` in other places.

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -0,0 +1,555 @@
...
+
+var makerAddress = constants.GanacheAccount1
+var takerAddress = constants.GanacheAccount2
+var eighteenDecimalsInBaseUnits = new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
+var wethAmount = new(big.Int).Mul(big.NewInt(50), eighteenDecimalsInBaseUnits)
+var zrxAmount = new(big.Int).Mul(big.NewInt(100), eighteenDecimalsInBaseUnits)
+var tokenID = big.NewInt(1)
+var rpcClient *ethrpc.Client
+
+func TestSetup(t *testing.T) {
```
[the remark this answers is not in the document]
fabioberger: Went with the `init` approach.

Makefile
```diff
@@ -39,7 +39,7 @@ test-all: test-go test-wasm-node test-wasm-browser
 
 .PHONY: test-go
 test-go:
-	go test ./... -race -timeout 30s
+	go test ./... -race -timeout 30s -p=1
```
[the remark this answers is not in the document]
fabioberger: Decided to split the tests so that those that can run in parallel still do.

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -0,0 +1,666 @@
...
+	txn, err := zrx.Transfer(opts, constants.GanacheAccount4, zrxAmount)
+	require.NoError(t, err)
+	ctx, cancelFn = context.WithTimeout(context.Background(), 4*time.Second)
+	receipt, err := bind.WaitMined(ctx, ethClient, txn)
+	cancelFn()
+	require.NoError(t, err)
+	assert.Equal(t, receipt.Status, uint64(1))
+
+	var orderEvents []*zeroex.OrderEvent
+	select {
```
albrow: After seeing how many times this is used in the tests, maybe it's worth refactoring into a function? Something like:

```go
func waitForOrderEvents(t *testing.T, orderEventsChan chan<- []*zeroex.OrderEvent, timeout time.Duration) []*zeroex.OrderEvent {
	select {
	case orderEvents = <-orderEventsChan:
		return orderEvents
	case <-time.After(timeout):
		t.Fatal("timed out waiting for order events")
	}
} 
```

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -0,0 +1,666 @@
...
+		t.Fatal("timed out waiting for order event")
+	}
+	require.Len(t, orderEvents, 1)
+	orderEvent := orderEvents[0]
+	assert.Equal(t, zeroex.EKOrderBecameUnfunded, orderEvent.Kind)
+
+	var orders []*meshdb.Order
+	err = meshDB.Orders.FindAll(&orders)
+	require.NoError(t, err)
+	assert.Len(t, orders, 1)
```
albrow: This should probably be `require` because if the length is zero it will lead to a panic with an "index out of range" error. Alternatively, you could add an if statement so that the other assertions are not run if the length is 0, but that would require more code.

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -0,0 +1,666 @@
...
+}
+
+var rpcClient *ethrpc.Client
+var ethClient *ethclient.Client
+var zrx *wrappers.ZRXToken
+var dummyERC721Token *wrappers.DummyERC721Token
+var exchange *wrappers.Exchange
+var weth *wrappers.WETH9
+
+func init() {
```
albrow: Just a thought: I'm pretty comfortable with this `init` function so far because it purely does set up and should never fail under normal circumstances. However, we should keep an eye on it. We want to avoid testing conditions/assumptions/behaviors within `init` because the error messages can be harder to interpret. Anything that could reasonably fail should be included in a test function where we have access to `testing.T`.

scenario/scenario.go
```diff
@@ -49,21 +51,20 @@ func CreateZRXForWETHSignedTestOrder(t *testing.T, makerAddress, takerAddress co
...
 	opts := &bind.TransactOpts{
 		From:   takerAddress,
 		Value:  wethAmount,
-		Signer: getTestSignerFn(takerAddress),
+		Signer: GetTestSignerFn(takerAddress),
 	}
 	txn, err := weth9.Deposit(opts)
 	require.NoError(t, err)
-	receipt, err := bind.WaitMined(context.Background(), ethClient, txn)
+	ctx, cancelFn := context.WithTimeout(context.Background(), 4*time.Second)
```
albrow: This pattern for calling `WaitMined` also shows up a lot and could potentially be refactored to a function.

fabioberger: Great idea! Added `waitTxnSuccessfullyMined`

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -688,3 +635,18 @@ func setupSubTest(t *testing.T) func(t *testing.T) {
...
+		fmt.Println("orderEvents!", orderEvents)
+		return orderEvents
+	case <-time.After(timeout):
+		elapsed := time.Since(start)
+		fmt.Println("elapsed", elapsed)
+		fmt.Println("timeout!")
+		t.Fatal("timed out waiting for order events")
+	}
+	return []*zeroex.OrderEvent{}
+}
```
albrow: nit: There's no new line at the end of this file. Do you have your editor configured to automatically add them?

scenario/scenario.go
```diff
@@ -313,3 +265,11 @@ func GetTestSignerFn(signerAddress common.Address) func(signer types.Signer, add
 		return tx.WithSignature(signer, signature)
 	}
 }
+
+func waitTxnSuccessfullyMined(t *testing.T, ethClient *ethclient.Client, txn *types.Transaction) {
+	ctx, cancelFn := context.WithTimeout(context.Background(), 4*time.Second)
+	receipt, err := bind.WaitMined(ctx, ethClient, txn)
+	require.NoError(t, err)
+	cancelFn()
```
albrow: nit: I think `defer cancelFn()` immediately after creating the context is cleaner and less brittle. As written, the context might not be canceled if we ever changed the code in the future or a panic caused the function to return early.

It's especially important outside of the context of tests. Consider cases where:

1. An unexpected panic causes a function to return early. The code never reaches `cancelFn` so the context is not canceled.
2. We recover from that panic and resume program execution.
3. The context is never canceled, potentially creating a goroutine leak.

To address issues like this, it is generally better to call `defer cancelFn` immediately. (It is less important here because we are not recovering from panics and the test binary will always exit).

zeroex/orderwatch/order_watcher_test.go
fabioberger: I have `    "files.insertFinalNewline": true,`

Issue #428, opened 2019-09-27
Reduce dependency bloat in TypeScript RPC client


albrow:
David from GUDecks.com is reporting that the TypeScript RPC client is pulling in 200MB of dependencies. We should see if we can trim this down.

PR #433, opened 2019-09-30
Move some deps to devDependencies for TS RPC client

albrow:
Related to #428. We can take even more steps if needed; this is just the low-hanging fruit.

### DIFF
--- a/rpc/clients/typescript/package.json
+++ b/rpc/clients/typescript/package.json
@@ -5,14 +5,7 @@
         "node": ">=6.12"
     },
     "description": "A Javascript library for interacting with the Mesh JSON RPC API over WebSockets",
-    "keywords": [
-        "p2p",
-        "mesh",
-        "0xproject",
-        "ethereum",
-        "tokens",
-        "exchange"
-    ],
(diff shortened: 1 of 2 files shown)


albrow:
CI failed twice in a row with `error Incorrect integrity when fetching from the cache`. I'm looking into it...

albrow:
Hard-coding the Yarn version in _.circleci/config.yml_ to `1.17` fixes the issue.

rpc/clients/typescript/package.json
```diff
@@ -37,8 +30,6 @@
     },
     "homepage": "https://github.com/0xProject/0x-mesh/rpc/clients/typescript/README.md",
     "dependencies": {
-        "@0x/assert": "^2.1.2",
```
fabioberger: The `@0x/assert` package is used in `ws_client.ts`, and is therefore not a `devDep`.

rpc/clients/typescript/package.json
```diff
@@ -37,8 +30,6 @@
...
-        "@0x/dev-utils": "^2.2.6",
```
fabioberger: `@0x/dev-utils` on the other hand is only used in tests and can therefore be moved. ✅

PR #444, opened 2019-10-07
Add RPCAddr option for the JSON-RPC API

nfeignon:
Allows choosing the JSON-RPC API listening interface.
Set it to localhost as default for security considerations.
It currently listens on any interface (0.0.0.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This changelog is a work in progress and may contain notes for versions which ha
 ### Features ✅ 
 
 - Implemented a new strategy for limiting the amount of database storage used by Mesh and removing orders when the database is full. This strategy involves a dynamically adjusting maximum expiration time. When the database is full, Mesh will enforce a maximum expiration time for all incoming orders and remove any existing orders with an expiration time too far in the future. If conditions change and there is enough space in the database again, the max expiration time will slowly increase. This is a short term solution which solves the immediate issue of finite storage capacities and does a decent job of protecting against spam. We expect to improve and possibly replace it in the future. See [#450](https://github.com/0xProject/0x-mesh/pull/450) for more details.
-- Added support for a new feature called "order pinning" ([#474](https://github.com/0xProject/0x-mesh/pull/474)). Pinned orders will not be affected by any DDoS prevention or incentive mechanisms (including the new dynamic max expiration time feature) and will always stay in storage until they are no longer fillable.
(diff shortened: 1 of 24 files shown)


albrow:
Thanks @nfeignon! Sorry for the late reply, we were busy at Ethereum Devcon last week.

This is a good change, but if we're going to allow users to specify a custom listen address, I would rather combine `RPC_ADDR` and `RPC_PORT` into a single environment variable for convenience. Can we change the `RPC_ADDR` env var to expect an IP address as well as a port? Then we can deprecate `RPC_PORT` and eventually remove it.

fabioberger:
@nfeignon we unfortunately had to force-push to `development` under exceptional circumstances. Could you please rebase your PR off of development?

The steps assuming you've updated your forks `development` to be identical to upstream would be:

```
git checkout feature/add-rpc-addr-option
git checkout -b feature/add-rpc-addr-option-old
git checkout development
git branch -D feature/add-rpc-addr-option
git checkout -b feature/add-rpc-addr-option
git cherry-pick c11591e14238808a7cfd34ce849375546b45c0b1
git push origin feature/add-rpc-addr-option -f
```

nfeignon:
@fabioberger No problem, I've rebased my PR.

@albrow I chose to separate RPC_ADDR and RPC_PORT because that's what Parity (--jsonrpc-port, --jsonrpc-interface) and Geth (--rpcaddr, --rpcport) do but I don't mind editing my PR :)
Should I remove RPC_PORT completely or ignore it if RPC_ADDR is defined? What about the default interface (any or localhost) if only RPC_PORT is defined?

albrow:
> Should I remove RPC_PORT completely or ignore it if RPC_ADDR is defined? What about the default interface (any or localhost) if only RPC_PORT is defined?

Here's what I'm thinking:

- Let's `RPC_PORT` for now, in order to maintain backwards compatibility.
- If `RPC_ADDR` is defined, we should ignore `RPC_PORT`.
- `RPC_ADDR` should be expected to always include the address *and* port.
- If `RPC_PORT` is defined and `RPC_ADDR` is not defined, we should use the current default bind address: `0.0.0.0` to maintain backwards compatibility.

Then, in a future release, we can remove the `RPC_PORT` environment variable.

albrow:
@nfeignon talked it over with @fabioberger and since we are nearing the `6.0.0` release at this point, we would like to go ahead and break backwards compatibility here. (It's better to do it now than later since we are still in beta at the moment).

Instead of what I suggested in my previous comment, let's simply remove `RPC_PORT`. `RPC_ADDR` should be expected to always include the address and port.

nfeignon:
> Instead of what I suggested in my previous comment, let's simply remove `RPC_PORT`. `RPC_ADDR` should be expected to always include the address and port.

Great, I updated my commit to remove RPC_PORT.

albrow:
@nfeignon in the interest of time, I'm moving this over to https://github.com/0xProject/0x-mesh/pull/487 so I can fix the merge conflicts for you. You will still get credited for your commit on GitHub.

Thanks again for your contribution and thank you for bearing with us!

PR #445, opened 2019-10-15
Add support for custom network ID/custom contract addresses

albrow:
This PR adds support for declaring a custom network ID and custom contract addresses. This is primarily intended for testing purposes and allows Mesh to be used with local Ethereum networks other than Ganache.

### DIFF
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -66,36 +66,30 @@ export interface Config {
     // Parity, feel free to double the default max in order to reduce the number
     // of RPC calls made by Mesh. Defaults to 524288 bytes.
     ethereumRPCMaxContentLength?: number;
-    // customContractAddresses is a mapping of
-    // network ID to contract addresses for that network ID. The contract
-    // addresses for most common networks are already included by default, so this
-    // is typically only needed for testing on custom networks. The given
-    // addresses are added to the default list of addresses for known networks and
-    // overriding any known contract addresses is not allowed. The addresses for
-    // exchange, devUtils, erc20Proxy, and erc721Proxy are required for each
-    // network. For example:
+    // customContractAddresses is set of custom addresses to use for the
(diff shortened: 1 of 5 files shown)


browser/ts/index.ts
```diff
@@ -66,6 +66,40 @@ export interface Config {
...
+    // network. For example:
+    //
+    //    {
+    //        "999": {
+    //            "exchange":"0x48bacb9266a570d521063ef5dd96e61686dbe788",
+    //            "devUtils": "0x38ef19fdf8e8415f18c307ed71967e19aac28ba1",
+    //            "erc20Proxy": "0x1dc4c1cefef38a777b15aa20260a54e584b16c48",
+    //            "erc721Proxy": "0x1d7022f5b17d2f8b695918fb48fa1089c9f85401"
+    //         }
+    //    }
```
fabioberger: Since there is never a reason to supply more then a single set of contract addresses, why don't we remove the networkID nesting and assume the supplied addresses are for the specified networkID (an already existing configuration option)?

integration-tests/browser/src/index.ts
```diff
@@ -45,6 +45,21 @@ provider.start();
         ethereumRPCURL,
         ethereumNetworkID: 50,
         bootstrapList: ['/ip4/127.0.0.1/tcp/60500/ws/ipfs/16Uiu2HAmGd949LwaV4KNvK2WDSiMVy7xEmW983VH75CMmefmMpP7'],
+        // Note(albrow): customContractAddresses is not actually required for
+        // the integration tests. Including them it here to test parsing and
```
fabioberger:
```suggestion
        // the integration tests. Including them here to test parsing and
```

browser/ts/index.ts
albrow: Good point. I think this should work since I'll also have access to the Ethereum network ID from the config object. I'll update this PR.

PR #446, opened 2019-10-15
ethereum: Update kovan DevUtils.sol address

fabioberger:
Apparently the fixed `DevUtils.sol` contract has not been re-deployed to Kovan until now. This PR updates it's address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
+## v5.0.1-beta
+
+### Bug fixes 🐞 
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
+## v5.0.1-beta
```
albrow: Have you thought about what to call the next version? I was assuming we would use something like `5.0.1-beta-0xv3` to show that it is designed for v3 of the protocol but still have a way to indicate breaking changes. It looks like on Docker we just tagged the image as `0xV3` which I don't like as much since there is no way to indicate breaking changes. I think on npm we used the same version but a special dist tag which is fine by me.

fabioberger: Yeah, this is a good point. I guess we should just suffix with `-0xv3`. I'll do that in the CHANGELOG and when we publish to Docker hub / NPM. The NPM tag will still be `protocolV3` to remain consistent with the rest of the 0x packages.

PR #450, opened 2019-10-16
core, orderwatch, meshdb: Implement a dynamically decreasing max expiration time for orders

### LINKED ISSUE #431: Solve the finite database storage problem
Mesh currently has no mechanisms to prevent the database from becoming full and eating up all available storage space. Internally, there have been a variety of proposals for mechanisms to remove certain orders when the database is becoming full. We need to reach consensus on a mechanism (we're getting close) and then implement it.

albrow:
Fixes #431.

This PR is still a WIP but I'm opening it now so we can discuss the progress so far. This is an implementation of the idea we've been discussing on Slack and a solution to the problem of finite database storage. To quickly re-summarize the idea:

When storage space is plentiful, we start without any kind of limit on expiration time. As the database fills up, we enforce a limit on the maximum expiration time. Any orders that expire after that time will be removed. If more orders keep coming in we can lower the expiration time and remove more orders.

So far this PR adds an index on `ExpirationTime` and implements a new method called `TrimOrdersByExpirationTime` which contains the bulk of the core logic for this feature. There are still some challenges around how to use this function and when to call it, but the approach is looking very feasible as of now. We also still need to determine when and how quickly to increase the max expiration time after storage frees up.

### DIFF
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -85,7 +85,7 @@ export interface Config {
     // The maximum number of orders that Mesh will keep in storage. As the
     // number of orders in storage grows, Mesh will begin enforcing a limit on
     // maximum expiration time for incoming orders and remove any orders with an
-    // expiration time too far in the future.
+    // expiration time too far in the future. Defaults to 100,000.
     maxOrdersInStorage?: number;
 }
 
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -66,14 +66,13 @@ export interface Config {
     // Parity, feel free to double the default max in order to reduce the number
     // of RPC calls made by Mesh. Defaults to 524288 bytes.
     ethereumRPCMaxContentLength?: number;
-    // customContractAddresses is set of custom addresses to use for the
-    // configured network ID. The contract addresses for most common networks
-    // are already included by default, so this is typically only needed for
-    // testing on custom networks. The given addresses are added to the default
-    // list of addresses for known networks and overriding any contract
-    // addresses for known networks is not allowed. The addresses for exchange,
-    // devUtils, erc20Proxy, and erc721Proxy are required for each network. For
-    // example:
+    // A set of custom addresses to use for the configured network ID. The
+    // contract addresses for most common networks are already included by
+    // default, so this is typically only needed for testing on custom networks.
+    // The given addresses are added to the default list of addresses for known
+    // networks and overriding any contract addresses for known networks is not
+    // allowed. The addresses for exchange, devUtils, erc20Proxy, and
+    // erc721Proxy are required for each network. For example:
     //
     //    {
     //        exchange: "0x48bacb9266a570d521063ef5dd96e61686dbe788",
@@ -83,6 +82,11 @@ export interface Config {
     //    }
     //
     customContractAddresses?: ContractAddresses;
+    // The maximum number of orders that Mesh will keep in storage. As the
+    // number of orders in storage grows, Mesh will begin enforcing a limit on
+    // maximum expiration time for incoming orders and remove any orders with an
+    // expiration time too far in the future.
+    maxOrdersInStorage?: number;
 }
 
 export interface ContractAddresses {
@@ -133,6 +137,7 @@ interface WrapperConfig {
     blockPollingIntervalSeconds?: number;
     ethereumRPCMaxContentLength?: number;
     customContractAddresses?: string; // json-encoded instead of Object.
+    maxOrdersInStorage?: number;
 }
 
 // The type for signed orders exposed by MeshWrapper. Unlike other types, the
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -399,6 +399,7 @@ export enum OrderEventEndState {
     Expired = 'EXPIRED',
     Unfunded = 'UNFUNDED',
     FillabilityIncreased = 'FILLABILITY_INCREASED',
+    StoppedWatching = 'STOPPED_WATCHING',
 }
 
 interface WrapperOrderEvent {
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -2,6 +2,7 @@ package constants
 
 import (
 	"errors"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 
@@ -60,3 +61,11 @@ var ErrInternal = errors.New("internal error")
 
 // TestMaxContentLength is the max Ethereum RPC Content-Length used in tests
 var TestMaxContentLength = 1024 * 512
+
+// UnlimitedExpirationTime is the maximum value for uint256, which mens there is
+// effectively no limit on the maximum expiration time for orders.
+var UnlimitedExpirationTime *big.Int
+
+func init() {
+	UnlimitedExpirationTime, _ = big.NewInt(2).SetString("115792089237316195423570985008687907853269984665640564039457584007913129639935", 10)
+}
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -62,10 +62,11 @@ var ErrInternal = errors.New("internal error")
 // TestMaxContentLength is the max Ethereum RPC Content-Length used in tests
 var TestMaxContentLength = 1024 * 512
 
-// UnlimitedExpirationTime is the maximum value for uint256, which means there is
-// effectively no limit on the maximum expiration time for orders.
+// UnlimitedExpirationTime is the maximum value for uint256 (2^256-1), which
+// means there is effectively no limit on the maximum expiration time for
+// orders.
 var UnlimitedExpirationTime *big.Int
 
 func init() {
-	UnlimitedExpirationTime, _ = big.NewInt(2).SetString("115792089237316195423570985008687907853269984665640564039457584007913129639935", 10)
+	UnlimitedExpirationTime, _ = big.NewInt(0).SetString("115792089237316195423570985008687907853269984665640564039457584007913129639935", 10)
 }
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -62,7 +62,7 @@ var ErrInternal = errors.New("internal error")
 // TestMaxContentLength is the max Ethereum RPC Content-Length used in tests
 var TestMaxContentLength = 1024 * 512
 
-// UnlimitedExpirationTime is the maximum value for uint256, which mens there is
+// UnlimitedExpirationTime is the maximum value for uint256, which means there is
 // effectively no limit on the maximum expiration time for orders.
 var UnlimitedExpirationTime *big.Int
 
--- a/core/core.go
+++ b/core/core.go
@@ -218,6 +218,7 @@ func New(config Config) (*App, error) {
 	}
 
 	// Initialize order watcher (but don't start it yet).
+	// TODO(albrow): Load previous max expiration time from the database.
 	orderWatcher, err := orderwatch.New(orderwatch.Config{
 		MeshDB:           meshDB,
 		BlockWatcher:     blockWatcher,
--- a/core/core.go
+++ b/core/core.go
@@ -118,7 +118,7 @@ type Config struct {
 	// storage. As the number of orders in storage grows, Mesh will begin
 	// enforcing a limit on maximum expiration time for incoming orders and remove
 	// any orders with an expiration time too far in the future.
-	MaxOrdersInStorage int `envvar:"MAX_ORDERS_IN_STORAGE" default:"10000"`
+	MaxOrdersInStorage int `envvar:"MAX_ORDERS_IN_STORAGE" default:"100000"`
 }
 
 type snapshotInfo struct {
--- a/core/core.go
+++ b/core/core.go
@@ -748,6 +748,7 @@ func (app *App) GetStats() (*rpc.GetStatsResponse, error) {
 		NumOrders:                 numOrders,
 		NumPeers:                  app.node.GetNumPeers(),
 		NumOrdersIncludingRemoved: numOrdersIncludingRemoved,
+		MaxExpirationTime:         app.orderWatcher.MaxExpirationTime().String(),
 	}
 	return response, nil
 }
@@ -776,6 +777,7 @@ func (app *App) periodicallyLogStats(ctx context.Context) {
 			"numOrders":                 stats.NumOrders,
 			"numOrdersIncludingRemoved": stats.NumOrdersIncludingRemoved,
 			"numPeers":                  stats.NumPeers,
+			"maxExpirationTime":         stats.MaxExpirationTime,
 		}).Info("current stats")
 	}
 }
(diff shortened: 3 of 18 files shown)


constants/constants.go
```diff
@@ -57,3 +58,11 @@ var ErrInternal = errors.New("internal error")
 
 // TestMaxContentLength is the max Ethereum RPC Content-Length used in tests
 var TestMaxContentLength = 1024 * 512
+
+// UnlimitedExpirationTime is the maximum value for uint256, which mens there is
```
fabioberger:
```suggestion
// UnlimitedExpirationTime is the maximum value for uint256, which means there is
```

constants/constants.go
```diff
@@ -57,3 +58,11 @@ var ErrInternal = errors.New("internal error")
...
+// effectively no limit on the maximum expiration time for orders.
+var UnlimitedExpirationTime *big.Int
+
+func init() {
+	UnlimitedExpirationTime, _ = big.NewInt(2).SetString("115792089237316195423570985008687907853269984665640564039457584007913129639935", 10)
```
fabioberger: Can you add a comment to what this number is? e.g., `2^64-1`?

core/core.go
```diff
@@ -114,6 +114,11 @@ type Config struct {
 	//    }
 	//
 	CustomContractAddresses string `envvar:"CUSTOM_CONTRACT_ADDRESSES" default:""`
+	// MaxOrdersInStorage is the maximum number of orders that Mesh will keep in
+	// storage. As the number of orders in storage grows, Mesh will begin
+	// enforcing a limit on maximum expiration time for incoming orders and remove
+	// any orders with an expiration time too far in the future.
+	MaxOrdersInStorage int `envvar:"MAX_ORDERS_IN_STORAGE" default:"10000"`
```
fabioberger: How many orders does GUDecks have? Think we might need to default this to a higher number. Thoughts?

zeroex/order.go
```diff
@@ -179,6 +179,12 @@ const (
 	// Fillability for an order can increase if a previously processed fill event
 	// gets reverted, or if a maker tops up their balance/allowance backing an order
 	ESOrderFillabilityIncreased = OrderEventEndState("FILLABILITY_INCREASED")
+	// Order is potentially still valid but was removed for a different reason
+	// (e.g. the database is full or the peer that sent the order was
+	// misbehaving). The order will no longer be watched and no further events for
+	// this order will be emitted. In some cases, the order may be re-added in the
+	// future.
+	ESOrderRemoved = OrderEventEndState("REMOVED")
```
fabioberger: It's super important for us to add pinning of orders added via JSON-RPC. Otherwise, this behavior is a deal-breaker for MMers and GUDecks.

zeroex/orderwatch/order_watcher.go
```diff
@@ -40,6 +41,14 @@ var permanentlyDeleteAfter = 4 * time.Minute
 // orders
 var expirationPollingInterval = 50 * time.Millisecond
 
+// maxOrdersTrimRatio affects how many orders are trimmed whenver we reach the
```
fabioberger:
```suggestion
// maxOrdersTrimRatio affects how many orders are trimmed whenever we reach the
```

zeroex/orderwatch/order_watcher.go
```diff
@@ -56,30 +65,46 @@ type Watcher struct {
...
 	assetDataDecoder := zeroex.NewAssetDataDecoder()
-	contractAddresses, err := ethereum.GetContractAddressesForNetworkID(networkID)
+	contractAddresses, err := ethereum.GetContractAddressesForNetworkID(config.NetworkID)
 	if err != nil {
 		return nil, err
 	}
 
+	if config.MaxOrders == 0 {
+		config.MaxOrders = defaultMaxOrders
+	}
```
fabioberger: This feels like the wrong place to do this. A default should be  set after parsing the configs no? It should be set before any part of the code could potentially use it.

zeroex/orderwatch/order_watcher.go
```diff
@@ -539,6 +587,48 @@ func (w *Watcher) Add(orderInfo *ordervalidator.AcceptedOrderInfo) error {
...
+
+	return nil
+}
+
+// MaxExpirationTime returns the current maximum expiration time for incoming
+// orders.
+func (w *Watcher) MaxExpirationTime() *big.Int {
+	return w.maxExpirationTime
+}
+
```
fabioberger: I'm fine with having this logic here but we should _also_ be checking if the order's expiration is greater than the current `maxExpirationTime` _after_ we validate that the message conforms to the order schema but _before_ we validate the order with an ETH RPC call. This should be done at the top of `core.go`'s `AddOrders()` and `message_handler.go`'s `HandleMessages()`. This will avoid nodes wasting RPC calls on orders that we already know won't get stored.

zeroex/orderwatch/order_watcher.go
```diff
@@ -539,6 +587,48 @@ func (w *Watcher) Add(orderInfo *ordervalidator.AcceptedOrderInfo) error {
...
+	for _, removedOrder := range removedOrders {
+		// Fire a "REMOVED" event for each order that was removed.
+		orderEvent := &zeroex.OrderEvent{
+			OrderHash:                removedOrder.Hash,
+			SignedOrder:              removedOrder.SignedOrder,
+			FillableTakerAssetAmount: removedOrder.FillableTakerAssetAmount,
+			EndState:                 zeroex.ESOrderRemoved,
+		}
+		w.orderFeed.Send([]*zeroex.OrderEvent{orderEvent})
+		// TODO(albrow): remove order from in-memory state.
```
fabioberger: Removing in-memory state involves:

```
expirationTimestamp := time.Unix(order.SignedOrder.ExpirationTimeSeconds.Int64(), 0)
w.expirationWatcher.Remove(expirationTimestamp, order.Hash.Hex())

err = w.removeAssetDataAddressFromEventDecoder(order.SignedOrder.MakerAssetData)
if err != nil {
	// This should never happen since the same error would have happened when adding
	// the assetData to the EventDecoder.
	logger.WithFields(logger.Fields{
		"error":       err.Error(),
		"signedOrder": order.SignedOrder,
	}).Error("Unexpected error when trying to remove an assetData from decoder")
	return err
}
```

zeroex/order.go
albrow: Definitely. But pinning is fairly independent so I'm planning to implement it in a subsequent PR.

zeroex/orderwatch/order_watcher.go
albrow: I only added it here to make it more convenient to write tests. Since tests don't go through the initialization process in the `core` package, they wouldn't have the defaults applied. That said, I'd be okay with removing it and changing our tests to always specify `MaxOrders`. What do you think?

zeroex/orderwatch/order_watcher.go
albrow: @fabioberger we already do that in __core/validation.go__. See the `validateOrders` function.

core/core.go
albrow: Just changed it from 10k to 100k. This keeps us [just under 1GB](https://www.wolframalpha.com/input/?i=8192+bytes+*+100000) (with some room for indexes) in the worst case scenario where every order is the max size.

GUDecks told us they have about 5k orders now so this leaves room for them to grow in the future. Also any NFT marketplaces with a larger number of orders can always configure their nodes to have a higher limit.

constants/constants.go
albrow: I already included a comment at the declaration of `UnlimitedExpirationTime` just above (the correct place for godoc comments). But I'll add that this is equal to 2^256-1.

browser/ts/index.ts
```diff
@@ -701,14 +705,14 @@ function wrapperContractEventsToContractEvents(wrapperContractEvents: WrapperCon
                 break;
             case ContractEventKind.ERC1155TransferBatchEvent:
                 const erc1155TransferBatchEvent = rawParameters as WrapperERC1155TransferBatchEvent;
-                const ids: BigNumber[] = []
```
albrow: @fabioberger This formatting change was made automatically by TSLint/Prettier. Can you double check that both of those are running automatically in your editor setup? I'll also add the corresponding checks to CI in a separate PR.

albrow: FYI opened the corresponding issue here: https://github.com/0xProject/0x-mesh/issues/460

albrow:
@fabioberger Removed WIP tag. This PR is ready for final review.

I do eventually want to optimize our `SlowCounter` implementation to reduce memory allocations and `math/big` operations. Our current strategy of using a polling loop inside of `OrderWatcher` to check if the max expiration time can be increased is also inefficient, but should be good enough for now.

albrow:
Just reminding myself to update the CHANGELOG after we include everything there for the upcoming 5.1.0-beta launch.

zeroex/orderwatch/slowcounter/slow_counter.go
```diff
@@ -0,0 +1,114 @@
...
+	// Rate controls how fast the offset increases after each increment.
+	Rate float64
+	// Interval is the amount of time to wait before each time the counter is
+	// incremented.
+	Interval time.Duration
+	// MaxCount is the maximum value for the counter. After reaching MaxCount, the
+	// counter will stop incrementing until reset.
+	MaxCount *big.Int
+
+	// maxCountFloat is CaxCount converted to a big.Float in order to make the
```
fabioberger:
```suggestion
	// maxCountFloat is MaxCount converted to a big.Float in order to make the
```

zeroex/orderwatch/order_watcher.go
```diff
@@ -16,29 +17,54 @@ import (
...
+	// increase the max expiration time.
+	maxExpirationTimeCheckInterval = 30 * time.Second
+
+	// configuration options for the SlowCounter used for increasing max
+	// expiration time. Effectively, we will increase every 5 minutes as long as
+	// there is enough space in the database for orders. The first increase will
+	// be 5 seconds and the amount doubles from there (second increase will be 10
+	// seconds, then 20 seconds, then 40, etc.)
+	slowCounterOffset   = 5 // seconds
+	slowCounterRate     = 2.0
```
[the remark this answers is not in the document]
albrow: I'm not convinced that "multiplier" is more descriptive/clear. "Rate" is a commonly used term in economics, biology, etc. as part of the [exponential growth formula](https://en.wikipedia.org/wiki/Exponential_growth). `SlowCounter` uses the exponential growth formula with only a slight variation (the use of an `offset` to allow finer tuning of the magnitude of increase after each increment).

zeroex/orderwatch/order_watcher.go
```diff
@@ -56,30 +82,75 @@ type Watcher struct {
...
+
+	// Apply defaults to config.
+	if config.MaxOrders == 0 {
+		config.MaxOrders = defaultMaxOrders
+	}
+	if config.MaxExpirationTime == nil {
+		config.MaxExpirationTime = constants.UnlimitedExpirationTime
+	} else if big.NewInt(time.Now().Unix()).Cmp(config.MaxExpirationTime) == 1 {
+		// MaxExpirationTime should never be in the past.
+		config.MaxExpirationTime = big.NewInt(time.Now().Unix())
```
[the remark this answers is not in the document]
albrow: `Unix` already converts the time to UTC. Since it's unclear how we are going to treat expiration buffers moving forward I would rather leave this as is for now.

zeroex/orderwatch/order_watcher.go
```diff
@@ -574,14 +671,33 @@ func (w *Watcher) cleanup(ctx context.Context) error {
...
+	// because of the way OrderWatcher works, the distinction shouldn't matter.
+	txn := w.meshDB.Orders.OpenTransaction()
+	defer func() {
+		_ = txn.Discard()
+	}()
+
+	// Final expiration time check before inserting the order. We might have just
+	// changed max expiration time above.
+	if orderInfo.SignedOrder.ExpirationTimeSeconds.Cmp(w.maxExpirationTime) == 1 {
+		return nil
```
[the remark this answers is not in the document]
albrow: Yes but it also means they never received the `ADDED` event, which seems consistent to me. What do you think is the right behavior here?

zeroex/orderwatch/order_watcher.go
albrow: After thinking about it some more I'd prefer to keep it the way it is. It is even more strange to rely on a completely separate package (`core`) and separate code path to apply the correct defaults for `OrderWatcher`. There's a huge potential footgun here where if `MaxOrders` is set to 0 a lot of things will break.

meshdb/meshdb_test.go
```diff
@@ -136,3 +136,119 @@ func TestParseContractAddressesAndTokenIdsFromAssetData(t *testing.T) {
 		assert.Equal(t, expectedSingleAssetData.TokenID, singleAssetData.TokenID)
 	}
 }
+
+func TestTrimOrdersByExpirationTime(t *testing.T) {
+	meshDB, err := New("/tmp/meshdb_testing/" + uuid.New().String())
+	require.NoError(t, err)
+	defer meshDB.Close()
+
+	// TODO(albrow): Move these to top of file.
```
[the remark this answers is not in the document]
albrow: Opened https://github.com/0xProject/0x-mesh/issues/472 to address this. There are a lot of places in the codebase where we should do this so I didn't want to include it in this PR.

zeroex/orderwatch/order_watcher.go
fabioberger: Let's throw on 0.

zeroex/orderwatch/order_watcher.go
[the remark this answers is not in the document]
fabioberger: I was talking about the existing buffer config but realize that it doesn't make sense to add it here.

zeroex/orderwatch/order_watcher.go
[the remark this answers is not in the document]
fabioberger: Since we `ACK`d the incoming JSON-RPC request, we should return `STOPPED_WATCHING`.

zeroex/orderwatch/order_watcher.go
```diff
@@ -574,14 +671,59 @@ func (w *Watcher) cleanup(ctx context.Context) error {
...
+	// Final expiration time check before inserting the order. We might have just
+	// changed max expiration time above.
+	if orderInfo.SignedOrder.ExpirationTimeSeconds.Cmp(w.maxExpirationTime) == 1 {
+		// HACK(albrow): This is technically not the ideal way to respond to this
+		// situation, but it is a lot easier to implement for the time being. In the
+		// future, we should return an error and then react to that error
+		// differently depending on whether the order was received via RPC or from a
+		// peer. In the former case, we should return an RPC error response
+		// indicating that the order was not in fact added. In the latter case, we
+		// should effectively no-op, nether penalizing the peer or emitting any
```
fabioberger:
```suggestion
		// should effectively no-op, neither penalizing the peer or emitting any
```

PR #453, opened 2019-10-17
Add support for ERC1155 Assets

fabioberger:
Fixes: #449 

Adds support for validating, storing, watching and sharing orders involving ERC1155 assets.


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
       - image: circleci/golang:1.12.9-browsers
       - image: 0xorg/ganache-cli:latest
         environment:
-            VERSION: 4.3.0
+            VERSION: 4.3.3
```
fabioberger: The V2 & V3 Ganache snapshots did not have ERC1155 Mintable tokens deployed on them so I had to add it.

zeroex/orderwatch/order_watcher.go
```diff
@@ -367,6 +367,74 @@ func (w *Watcher) handleBlockEvents(events []*blockwatch.Event) error {
...
+			case "ERC1155TransferSingleEvent":
+				var transferEvent decoder.ERC1155TransferSingleEvent
+				err = w.eventDecoder.Decode(log, &transferEvent)
+				if err != nil {
+					if isNonCritical := w.checkDecodeErr(err, eventType); isNonCritical {
+						continue
+					}
+					return err
+				}
+				// HACK(fabio): Currently we simply revalidate all orders involving assets in this
```
albrow: Are we sure this stuff regarding Augur is going to be true anymore?

fabioberger: Yes, since Augur will be using a DummyERC1155 token contract and the actual settlement could involve a slew of "complete set" magic, we cannot rely on the "ID" as we could with a regular ERC1155 token.

PR #454, opened 2019-10-17
core/message-handler: round robin order sharing

jalextowle:
## Description
The `GetOrdersToShare` function gets a set of orders from the mesh database to share with other peers. The current strategy for selecting orders to share is simply to choose a random segment of the valid orders in the database's snapshot of length `max`. This strategy works well---albeit imperfectly---when `(number of valid orders)` is reasonably close to `max`, but as the number of valid orders climbs, the risk of some orders not being propagated to peers increases. 

A simple strategy to mitigate this issue is to use a `round robin strategy` for selecting orders. This strategy will select orders by starting at the "beginning" of the database's collection of orders and working all the way up to the last orders in the database. Once the top order has been shared, the strategy loops around and starts sharing orders closer to the "beginning" of the order collection again. This strategy has the benefit that it will exhaustively go through a node's orders, so it will necessarily share all of the valid orders.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -137,6 +137,11 @@ type App struct {
 	snapshotExpirationWatcher *expirationwatch.Watcher
 	muIdToSnapshotInfo        sync.Mutex
 	idToSnapshotInfo          map[string]snapshotInfo
+	messageHandler            *MessageHandler
+}
+
+type MessageHandler struct {
+	nextOffset int
 }
 
 func New(config Config) (*App, error) {
@@ -234,6 +239,9 @@ func New(config Config) (*App, error) {
 	if err != nil {
 		return nil, err
 	}
+	messageHandler := &MessageHandler{
+		nextOffset: 0,
+	}
 
 	app := &App{
 		config:                    config,
@@ -249,6 +257,7 @@ func New(config Config) (*App, error) {
 		meshMessageJSONSchema:     meshMessageJSONSchema,
 		snapshotExpirationWatcher: snapshotExpirationWatcher,
 		idToSnapshotInfo:          map[string]snapshotInfo{},
+		messageHandler:            messageHandler,
 	}
 
 	log.WithFields(map[string]interface{}{
--- a/core/core.go
+++ b/core/core.go
@@ -140,10 +140,6 @@ type App struct {
 	messageHandler            *MessageHandler
 }
 
-type MessageHandler struct {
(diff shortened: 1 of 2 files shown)


core/message_handler.go
```diff
@@ -32,13 +46,17 @@ func (app *App) GetMessagesToShare(max int) ([][]byte, error) {
...
 	}
-	// If count is less than max, we use an offset of 0 to simply return all
-	// orders we have stored.
-	offset := 0
-	if count > max {
-		// If count is greater than max, choose an offset such that we always return
-		// at least max orders.
-		offset = rand.Intn(count - max)
+	// The offset will be the value closest to `nextOffset` such that the
+	// maximum number of orders that can be shared are selected.
```
albrow: Hmm, I think there is a small inefficiency in this implementation. Suppose `count = 25` orders and `max = 10`:

1. On the first iteration, `nextOffset = 0` and `offset = 0`. We select orders 0 through 9. Then we set `nextOffset = 10`.
2. On the second iteration, `nextOffset = 10` and `offset = 10`. We select orders 10 through 19. Then we set `nextOffset = 20`.
3. On the third iteration, `nextOffset = 20`. Here the condition `count - offset < max` is true, so we would set `offset = max(count - max, 0) = 15`. We would select orders 15 through 24.

Step 3 is inefficient since we already shared orders 15 through 19 in the previous iteration.

Instead of this, maybe we could either:

1. Leave `offset` unchanged and select orders 20 through 24 in step 3. It isn't hugely important to always select `max` orders now that we don't have to worry about the probability of some orders never being selected.
2. Perform a second query and select orders 20 through 24 *and* orders 0 through 4. Then set `nextOffset = 5`. This is the most optimal version but adds a little bit of complexity with the second query.

jalextowle: I liked the second strategy, so that's what I ended up going with. It does increase the complexity, but I don't think that it's too bad.

core/core.go
```diff
@@ -137,6 +137,11 @@ type App struct {
 	snapshotExpirationWatcher *expirationwatch.Watcher
 	muIdToSnapshotInfo        sync.Mutex
 	idToSnapshotInfo          map[string]snapshotInfo
+	messageHandler            *MessageHandler
+}
+
+type MessageHandler struct {
```
albrow: Let's move this to the __message_handler.go__ file.

PR #455, opened 2019-10-17
p2p: Segregate V3 Mesh network from V2 network

fabioberger:
Update pubSubTopic, rendezvous string and DHT ID so that the V3 Mesh network is separate from the V2 network.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
-## v5.0.1-beta-0xv3.1
+## v5.0.2-beta-0xv3
 
 ### Bug fixes 🐞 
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,12 +2,13 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
-## v5.0.1-beta-0xv3
+## v5.0.1-beta-0xv3.1
```
albrow: @fabioberger What's the reasoning here? `0xv3` is meant to indicate compatibility with 0x protocol version 3. This change makes it look like we are targeting protocol version 3.1.

p2p/bootstrap.go
```diff
@@ -12,7 +12,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const dhtProtocolID = protocol.ID("/0x-mesh-dht/version/1")
+const dhtProtocolID = protocol.ID("/0x-mesh-dht/version/2")
```
albrow: We need to keep the DHT protocol ID the same. Otherwise peer discovery won't work due to the fact that our bootstrap nodes are using `version/1`. The alternative is to spin up new bootstrap nodes, but I don't think there's a strong reason to do that.

CHANGELOG.md
fabioberger: Ah shoot, you are right. I was treating `beta-0xv3` as the beta tag. `.1` being the next version of that beta.

Let's go ahead and version them separately.

p2p/bootstrap.go
fabioberger: Good catch.

PR #465, opened 2019-10-22
Add ENABLE_RELAY_HOST env var for bootstrap nodes

### LINKED ISSUE #447: Build relay-only binary
We need to create binary executable which only runs a relay and doesn't do any order sharing. #430 and #383 are blocked by this.

albrow:
Fixes #447.

The idea is that bootstrap-only nodes will run with `ENABLE_RELAY_HOST=false` and relay-only nodes will run with `ENABLE_RELAY_HOST=true`. The only other functional difference is that the bootstrap nodes are hard-coded in: https://github.com/0xProject/0x-mesh/blob/90c3d03941d328b5f7f92abcc5b0654c7d250bc3/p2p/bootstrap.go#L19

### DIFF
--- a/cmd/mesh-bootstrap/main.go
+++ b/cmd/mesh-bootstrap/main.go
@@ -64,6 +64,9 @@ type Config struct {
 	// "/ip4/3.214.190.67/tcp/60558/ipfs/16Uiu2HAmGx8Z6gdq5T5AQE54GMtqDhDFhizywTy1o28NJbAMMumF").
 	// If empty, the default bootstrap list will be used.
 	BootstrapList string `envvar:"BOOTSTRAP_LIST" default:""`
+	// EnableRelayHost is whether or not the node should serve as a relay host.
+	// Defaults to true.
+	EnableRelayHost bool `envvar:"ENABLE_RELAY_HOST" default:"true"`
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/mesh-bootstrap/main.go
```diff
@@ -141,11 +144,15 @@ func main() {
...
 		libp2p.ConnectionManager(connManager),
-		libp2p.EnableRelay(circuit.OptHop),
 		libp2p.EnableAutoRelay(),
 		libp2p.Routing(newDHT),
 		libp2p.AddrsFactory(newAddrsFactory(advertiseAddrs)),
 	}
+	if config.EnableRelayHost {
+		opts = append(opts, libp2p.EnableRelay(circuit.OptHop))
+	} else {
+		opts = append(opts, libp2p.EnableRelay())
```
fabioberger: Looks like either way, we're enabling relay? Or is it not actually enabled if no args are passed in?

albrow: The name of this configuration option in libp2p is confusing. Here's the documentation for it:

> // EnableRelay configures libp2p to enable the relay transport with
> // configuration options. By default, this option only configures libp2p to
> // accept inbound connections from relays and make outbound connections
> // _through_ relays when requested by the remote peer. (default: enabled)
> //
> // To _act_ as a relay, pass the circuit.OptHop option.

albrow: What we care about here is whether the bootstrap node can _act_ as a relay, and that is determined by the presence of the `circuit.OptHop` option.

fabioberger: Makes sense, thanks for clarifying.

Issue #459, opened 2019-10-21
orderwatch: Delete `removed` orders from DB more frequently


fabioberger:
Currently we re-evaluate whether to remove an order that has been flagged for removal from the DB during the cleanup job (runs every hour) or if an event triggers the orders revalidation. We feel the 1hr interval between cleanup jobs is too long and can cause the DB to be full of `removed` orders. We should delete `removed` orders after ~5min (the time it takes for a block re-org to become highly unlikely). We can do this, by checking `removed` orders if ~5min have passed in a separate loop every ~5min.

albrow:
Fixed by https://github.com/0xProject/0x-mesh/pull/471

PR #471, opened 2019-10-24
orderwatch: Check for stale `removed` orders more frequently

fabioberger:
Fixes: #459

This PR:
- Adds a process to OrderWatcher that checks for orders that have been flagged for removal and should now be permanently deleted every 5 minutes.
- Stops updating an order DB entry's `lastUpdatedTime` when the entry is in-fact not updated (i.e., when we revalidate the order but no change to fillability was found). This could cause an order from a particularly active maker to stick around indefinitely despite being expired/cancelled/fullyFilled, because the changes in the maker's balance would cause this order to constantly be re-validated and it's `lastUpdatedTime` to constantly get bumped up.
- Slightly increases permanentlyDeleteAfter to 25 block depth to account for us flagging an order for removal due to UTC expiry, but it still potentially getting filled/cancelled a few blocks after. This adjustment is out of extreme caution.

### DIFF
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -39,9 +39,9 @@ const (
 	lastUpdatedBuffer = 30 * time.Minute
 
 	// permanentlyDeleteAfter specifies how long after an order is marked as IsRemoved and not updated that
-	// it should be considered for permanent deletion. Blocks get mined on avg. every 12 sec, so 4 minutes
-	// corresponds to a block depth of ~20.
-	permanentlyDeleteAfter = 4 * time.Minute
+	// it should be considered for permanent deletion. Blocks get mined on avg. every 12 sec, so 5 minutes
+	// corresponds to a block depth of ~25.
+	permanentlyDeleteAfter = 5 * time.Minute
 
 	// expirationPollingInterval specifies the interval in which the order watcher should check for expired
 	// orders
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -959,8 +959,7 @@ func (w *Watcher) generateOrderEventsIfChanged(ordersColTxn *db.Transaction, ord
 			}
 			orderEvents = append(orderEvents, orderEvent)
 		} else if oldFillableAmount.Cmp(newFillableAmount) == 0 {
-			// No important state-change happened, simply update lastUpdated timestamp in DB
-			w.updateOrderDBEntry(ordersColTxn, order)
+			// No important state-change happened
 		} else if oldFillableAmount.Cmp(big.NewInt(0)) == 1 && oldAmountIsMoreThenNewAmount {
 			// Order was filled, emit  event and update order in DB
 			order.FillableTakerAssetAmount = newFillableAmount
@@ -997,8 +996,6 @@ func (w *Watcher) generateOrderEventsIfChanged(ordersColTxn *db.Transaction, ord
 			oldFillableAmount := order.FillableTakerAssetAmount
 			if oldFillableAmount.Cmp(big.NewInt(0)) == 0 {
 				// If the oldFillableAmount was already 0, this order is already flagged for removal.
-				// Update it's lastUpdated timestamp in DB
-				w.updateOrderDBEntry(ordersColTxn, order)
 			} else {
 				// If oldFillableAmount > 0, it got fullyFilled, cancelled, expired or unfunded, emit event
 				w.unwatchOrder(ordersColTxn, order, big.NewInt(0))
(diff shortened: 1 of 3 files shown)


zeroex/orderwatch/order_watcher.go
```diff
@@ -211,6 +226,27 @@ func (w *Watcher) cleanupLoop(ctx context.Context) error {
...
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+
+		start = time.Now()
+		if err := w.permanentlyDeleteStaleRemovedOrders(ctx); err != nil {
+			return err
+		}
```
fabioberger: We check for stale `removed` orders on the leading edge of the loop so that stale orders are removed on Mesh startup.

zeroex/orderwatch/order_watcher.go
```diff
@@ -826,7 +878,11 @@ func (w *Watcher) unwatchOrder(u updatableOrdersCol, order *meshdb.Order, newFil
 	w.expirationWatcher.Remove(expirationTimestamp, order.Hash.Hex())
 }
 
-func (w *Watcher) permanentlyDeleteOrder(ordersColTxn *db.Transaction, order *meshdb.Order) error {
+type deletableOrdersCol interface {
```
albrow: I think the idea behind using this `deletableOrdersCol` was to make it compatible with both transactions and the Orders collection? I don't think we ever actually pass in a transaction so maybe we can just remove this interface for now?

zeroex/orderwatch/order_watcher.go
```diff
@@ -702,8 +757,7 @@ func (w *Watcher) generateOrderEventsIfChanged(ordersColTxn *db.Transaction, ord
 			}
 			orderEvents = append(orderEvents, orderEvent)
 		} else if oldFillableAmount.Cmp(newFillableAmount) == 0 {
-			// No important state-change happened, simply update lastUpdated timestamp in DB
-			w.updateOrderDBEntry(ordersColTxn, order)
+			// No important state-change happened
 		} else if oldFillableAmount.Cmp(big.NewInt(0)) == 1 && oldAmountIsMoreThenNewAmount {
```
albrow: We don't need to necessarily address it right now, but we should go over all of our usage of the `math/big` package to see if there is room for optimization. For example, every occurrence of `big.NewInt(0)` allocates a non-trivial amount of memory and we could reduce the number of allocations by using a single variable instead.

zeroex/orderwatch/order_watcher.go
```diff
@@ -571,6 +607,25 @@ func (w *Watcher) cleanup(ctx context.Context) error {
 	return w.generateOrderEventsIfChanged(ordersColTxn, orderHashToDBOrder, orderHashToEvents, rpc.LatestBlockNumber)
 }
 
+func (w *Watcher) permanentlyDeleteStaleRemovedOrders(ctx context.Context) error {
+	var removedOrders []*meshdb.Order
+	isRemovedFilter := w.meshDB.Orders.IsRemovedIndex.ValueFilter([]byte{1})
```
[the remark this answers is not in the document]
fabioberger: Agreed.

zeroex/orderwatch/order_watcher.go
fabioberger: We are using a transaction when it's called in `generateOrderEventsIfChanged`. I think we need to keep it.

zeroex/orderwatch/order_watcher.go
fabioberger: Agreed. Can you open an issue for it?

zeroex/orderwatch/order_watcher.go
albrow: Ah I see. That doesn't show up in the diff for this PR.

In that case can we rename this interface to `orderDeleter` to conform with naming conventions for small interfaces in Go and also rename the argument name in `permanentlyDeleteOrder` to reflect this?

zeroex/orderwatch/order_watcher.go
albrow: Done: https://github.com/0xProject/0x-mesh/issues/473

zeroex/orderwatch/order_watcher.go
```diff
@@ -826,7 +875,11 @@ func (w *Watcher) unwatchOrder(u updatableOrdersCol, order *meshdb.Order, newFil
 	w.expirationWatcher.Remove(expirationTimestamp, order.Hash.Hex())
 }
 
-func (w *Watcher) permanentlyDeleteOrder(ordersColTxn *db.Transaction, order *meshdb.Order) error {
+type orderDeleter interface {
+	Delete(id []byte) error
+}
+
+func (w *Watcher) permanentlyDeleteOrder(ordersColTxn orderDeleter, order *meshdb.Order) error {
```
albrow:
```suggestion
func (w *Watcher) permanentlyDeleteOrder(deleter orderDeleter, order *meshdb.Order) error {
```

PR #474, opened 2019-10-25
Implement order pinning

### LINKED ISSUE #469: Add order pinning
We want to add a feature called "order pinning" so that users of Mesh can have confidence that _their_ orders won't stop being watched. This would be implemented by an additional field when submitting orders via JSON-RPC which allows the order to be "pinned". Pinned orders will never be removed due to their expiration time being too far in the future.

albrow:
Fixes #469. The new feature is an optional parameter for RPC requests and the `addOrdersAsync` method in the TypeScript bindings, so this PR maintains backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog is a work in progress and may contain notes for versions which ha
 ### Features ✅ 
 
 - Implemented a new strategy for limiting the amount of database storage used by Mesh and removing orders when the database is full. This strategy involves a dynamically adjusting maximum expiration time. When the database is full, Mesh will enforce a maximum expiration time for all incoming orders and remove any existing orders with an expiration time too far in the future. If conditions change and there is enough space in the database again, the max expiration time will slowly increase. This is a short term solution which solves the immediate issue of finite storage capacities and does a decent job of protecting against spam. We expect to improve and possibly replace it in the future. See [#450](https://github.com/0xProject/0x-mesh/pull/450) for more details.
+- Added support for a new feature called "order pinning" ([#474](https://github.com/0xProject/0x-mesh/pull/474)). Pinned orders will not be affected by any DDoS prevention or incentive mechanisms (including the new dynamic max expiration time feature) and will always stay in storage until they are no longer fillable.
 
 ## v5.1.0-beta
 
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This changelog is a work in progress and may contain notes for versions which ha
 ### Features ✅ 
 
 - Implemented a new strategy for limiting the amount of database storage used by Mesh and removing orders when the database is full. This strategy involves a dynamically adjusting maximum expiration time. When the database is full, Mesh will enforce a maximum expiration time for all incoming orders and remove any existing orders with an expiration time too far in the future. If conditions change and there is enough space in the database again, the max expiration time will slowly increase. This is a short term solution which solves the immediate issue of finite storage capacities and does a decent job of protecting against spam. We expect to improve and possibly replace it in the future. See [#450](https://github.com/0xProject/0x-mesh/pull/450) for more details.
+- Added support for a new feature called "order pinning" ([#474](https://github.com/0xProject/0x-mesh/pull/474)). Pinned orders will not be affected by any DDoS prevention or incentive mechanisms (including the new dynamic max expiration time feature) and will always stay in storage until they are no longer fillable.
```
fabioberger: Can we mention that orders submitted via the node's JSON-RPC interface will default to being pinned?

core/core.go
```diff
@@ -686,12 +689,22 @@ func (app *App) AddOrders(signedOrdersRaw []*json.RawMessage) (*ordervalidator.V
...
 
-	for _, acceptedOrderInfo := range allValidationResults.Accepted {
+	for i, acceptedOrderInfo := range allValidationResults.Accepted {
 		// Add the order to the OrderWatcher. This also saves the order in the
 		// database.
-		err = app.orderWatcher.Add(acceptedOrderInfo)
+		err = app.orderWatcher.Add(acceptedOrderInfo, pinned)
 		if err != nil {
-			return nil, err
+			if err == meshdb.ErrFilledWithPinnedOrders {
```
```suggestion
			if err == meshdb.ErrDBFilledWithPinnedOrders {
```

`filled` is used a lot when refering to order fills. Can we be more explicit?

CHANGELOG.md
albrow: 👍Good idea. Just added.

albrow:
> This is looking good! Can you also add support for the isPinned option to the TS client?

Nice catch! Just added.

PR #476, opened 2019-10-28
Unit Tests for `GetMessagesToShare`

jalextowle:
## Description
- Removed the `App` dependency from `GetMessagesToShare` by refactoring to use a `OrderSelector` object instead.
- Added an isolated test for `GetMessagesToShare`. In this context, isolated just means that `GetMessagesToShare` is being called on a fresh set of active orders in the mesh database.
- Added a test that serially calls `GetMessagesToShare` with a few different scenarios.


core/message_handler_test.go
```diff
@@ -0,0 +1,230 @@
...
+	orders := make([]*meshdb.Order, orderCount)
+
+	for i := 0; i < orderCount; i++ {
+		orders[i] = randomOrder(t)
+	}
+
+	return orders
+}
+
+func randomOrder(t *testing.T) *meshdb.Order {
```
albrow: Let's use the `scenario` package here instead of re-writing functions for generating orders. I don't really see any reason to randomize the order fields since the method we're testing doesn't look at the order fields or process them at all.

core/message_handler_test.go
```diff
@@ -0,0 +1,230 @@
...
+	orders = randomOrders(t, 12)
+	allOrders = append(allOrders, orders...)
+	insertOrders(t, selector, orders)
+	verifyRoundRobinSharing(t, selector, selector.nextOffset, 7)
+}
+
+// Verify that the correct messages are shared by `GetMessagesToShare` given `orders`, a `nextOffset`, and `max`
+func verifyRoundRobinSharing(t *testing.T, selector *OrderSelector, nextOffset int, max int) {
+	notRemovedFilter := selector.db.Orders.IsRemovedIndex.ValueFilter([]byte{0})
+	ordersSnapshot, err := selector.db.Orders.GetSnapshot()
```
albrow: I don't think we need a snapshot here. [The documentation](https://godoc.org/github.com/0xProject/0x-mesh/db#Snapshot) explains:

> Snapshot is a frozen, read-only snapshot of a DB state at a particular point in time.

It's only required when other goroutines might make modifications to the database and you want to run several queries without having the state change in between.

core/message_handler_test.go
```diff
@@ -0,0 +1,230 @@
...
+	// Get the number of orders in the database
+	count, err := ordersSnapshot.NewQuery(notRemovedFilter).Count()
+	require.NoError(t, err)
+
+	expectedOrdersLength := min(max, count)
+	expectedOrders := make([][]byte, expectedOrdersLength)
+
+	// Get all of the orders in the database.
+	var orderList []*meshdb.Order
+	err = ordersSnapshot.NewQuery(notRemovedFilter).Offset(0).Max(count).Run(&orderList)
```
albrow: `Offset` and `Max` are not needed here. `Offset` defaults to 0 and `Max` defaults to returning all models in the collection. We also need to handle potential errors.

```suggestion
	err = ordersSnapshot.NewQuery(notRemovedFilter).Run(&orderList)
	require.NoError(t, err)
```

core/message_handler_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// Update `nextOffset` to zero if it is larger than the number of orders that are stored
+	if nextOffset > count {
+		nextOffset = 0
+	}
+
+	// Calculate the orders that we expect to be shared
+	for i := 0; i < expectedOrdersLength; i++ {
+		encoding, err := encodeOrder(orderList[(nextOffset+i)%count].SignedOrder)
```
albrow: nit: How about renaming `encoding` to `encodedOrder`? `encoding` makes me think of something like JSON or hex encoding.

core/message_handler.go
```diff
@@ -12,8 +12,9 @@ import (
 // Ensure that App implements p2p.MessageHandler.
 var _ p2p.MessageHandler = &App{}
 
-type MessageHandler struct {
+type OrderSelector struct {
```
albrow: `OrderSelector` should probably be unexported. (`MessageHandler` should have been too; I just didn't notice it at the time).

core/core.go
```diff
@@ -570,7 +569,7 @@ func (app *App) GetOrders(page, perPage int, snapshotID string) (*rpc.GetOrdersR
 		app.muIdToSnapshotInfo.Unlock()
 	}
 
-	notRemovedFilter := app.db.Orders.IsRemovedIndex.ValueFilter([]byte{0})
+	notRemovedFilter := app.orderSelector.db.Orders.IsRemovedIndex.ValueFilter([]byte{0})
```
albrow: It doesn't seem right to have `app` constantly reach into the `orderSelector` field in order to access `db`. Maybe we should re-add the `db` field to `app` as well? `db` is just a pointer so it's not expensive to have multiple copies of it.

Alternatively, `orderSelector` could own all queries and `app` doesn't access `orderSelector.db` directly. This would involve a bit of work creating helper methods, so I'm more inclined to go with the first option.

core/validation.go
```diff
@@ -162,7 +162,7 @@ func (app *App) validateOrders(orders []*zeroex.SignedOrder) (*ordervalidator.Va
 
 		// Check if order is already stored in DB
 		var dbOrder meshdb.Order
-		err = app.db.Orders.FindByID(orderHash.Bytes(), &dbOrder)
+		err = app.orderSelector.db.Orders.FindByID(orderHash.Bytes(), &dbOrder)
```
albrow: This could also be changed if we re-add the `db` field to `app`.

```suggestion
		err = app.db.Orders.FindByID(orderHash.Bytes(), &dbOrder)
```

PR #481, opened 2019-10-29
Implement bandwidth limits for relay and bootstrap nodes

### LINKED ISSUE #390: Implement strict rate-limiting for relay hosts.
https://github.com/0xProject/0x-mesh/issues/387 introduced a new feature which makes it possible to ban certain IP addresses. The plan is to use this feature to ban the IP address of peers who are attempting to spam the network with too much data.

There is an issue with this approach. At a low abstraction level, all messages sent through relayed connections appear to be coming from the same IP address - the IP address of the relay host. But we actually don't want to punish relay hosts for merely relaying the messages; we want to ban the peer that was sending the messages in the first place.

The best way to do this is to make all relay hosts more strict about how many messages they are willing to relay on behalf of other peers. With this in place, we know that a well-behaving relay host should remain below a certain bandwidth cap. If any relay host exceeds this cap, we can ban them.

albrow:
Fixes #390.


albrow:
Removed WIP tag. This PR is ready for review.

PR #483, opened 2019-10-30
Implement rate-limiting for GossipSub messages

### LINKED ISSUE #432: Limit the rate at which new orders can be created
#119 describes rate-limiting on a low-level, which involves looking at the bandwidth used by each peer in terms of bytes/second. This is a necessary component, but won't work well as a DDoS prevention mechanism on its own. We should combine this with a higher-level limit on the number of new orders which can be created and shared. This limit can either apply to peer IDs/IP addresses, maker addresses, or both.

albrow:
Although it isn't exactly what we originally had in mind, this PR effectively fixes #432.

The features is implemented via a custom pubsub validator, which we register via [`RegisterTopicValidator`](https://godoc.org/github.com/libp2p/go-libp2p-pubsub#PubSub.RegisterTopicValidator). This custom validator only considers rate-limiting and looks at the number of messages sent globally as well as by each peer. Any messages which exceed the limit are considered "invalid". They are effectively dropped and are not processed or propagated to other peers, but the peers that sent those messages are not punished or banned. This will help cut down on bandwidth and other resource usage, especially if someone is spamming the network.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -10,7 +10,6 @@ import (
 	"io/ioutil"
 	mathrand "math/rand"
 	"path/filepath"
-	"sync"
 	"time"
 
 	"github.com/0xProject/0x-mesh/p2p/banner"
@@ -83,7 +82,6 @@ const (
 // messages.
 type Node struct {
 	ctx              context.Context
-	cancel           context.CancelFunc
 	config           Config
 	messageHandler   MessageHandler
 	host             host.Host
@@ -176,10 +174,6 @@ func New(ctx context.Context, config Config) (*Node, error) {
 		config.PerPeerPubSubMessageBurst = defaultPerPeerPubSubMessageBurst
 	}
 
-	// Create a child context so that we can preemptively cancel if there is an
-	// error.
-	ctx, cancel := context.WithCancel(ctx)
-
 	// We need to declare the newDHT function ahead of time so we can use it in
 	// the libp2p.Routing option.
 	var kadDHT *dht.IpfsDHT
@@ -196,7 +190,6 @@ func New(ctx context.Context, config Config) (*Node, error) {
 	// Get environment specific host options.
 	opts, err := getHostOptions(ctx, config)
 	if err != nil {
-		cancel()
 		return nil, err
 	}
 
@@ -222,7 +215,6 @@ func New(ctx context.Context, config Config) (*Node, error) {
 	// Initialize the host.
 	basicHost, err := libp2p.New(ctx, opts...)
 	if err != nil {
-		cancel()
 		return nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


p2p/node.go
```diff
@@ -124,6 +152,18 @@ func New(ctx context.Context, config Config) (*Node, error) {
 	} else if config.RendezvousString == "" {
 		return nil, errors.New("config.RendezvousString is required")
 	}
+	if config.GlobalPubSubMessageLimit == 0 {
```
albrow: @fabioberger This is similar to my other PR, but in this case the `core` package doesn't need to know about these config options at all and I'd like to keep it that way (they aren't user-configurable). The number of options spread around are codebase is going to continue to grow and I think we should be okay with creating defaults in certain sub-packages. What do you think?

p2p/ratevalidator/validator.go
```diff
@@ -0,0 +1,115 @@
...
+	peerLimiter, err := v.getOrCreateLimiterForPeer(peerID)
+	if err != nil {
+		log.WithError(err).Error("unexpected error in getOrCreateLimiterForPeer")
+		return false
+	}
+	if !peerLimiter.Allow() {
+		return false
+	}
+
+	return v.globalLimiter.Allow()
```
fabioberger: If we limit the number of peers a node can have, why do we need a global limiter at all? Doesn't having a per peer limiter and limiting the number of peers mean this global limiter will never get triggered?

p2p/node_test.go
```diff
@@ -510,3 +506,196 @@ func TestProtectIP(t *testing.T) {
...
+	// HACK(albrow): Wait for GossipSub messages to fully propagate.
+	time.Sleep(1 * time.Second)
+
+	require.NoError(t, node0.runOnce())
+	require.NoError(t, node1.runOnce())
+	require.NoError(t, node2.runOnce())
+
+	// node2 should only have config.PerPeerPubSubMessageBurst*2 messages.
+	// The others are expected to have been dropped.
+	expectedMessageCount := node0.config.PerPeerPubSubMessageBurst * 2
```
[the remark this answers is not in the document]
albrow: We expect `PerPeerPubSubMessageBurst * 2` here because two peers have sent `PerPeerPubSubMessageBurst` each.

p2p/ratevalidator/validator.go
albrow: > Doesn't having a per peer limiter and limiting the number of peers mean this global limiter will never get triggered?

Not necessarily. The per-peer limit is based on the `From` address of messages. Because of the way GossipSub works, the `From` address is not necessarily a peer that you are directly connected to. Because of this fact, the per-peer limit by itself does not limit the total amount of messages that you would process and propagate to your peers.

PR #485, opened 2019-10-31
Mesh 0x V2: Change mentions of `networkID` to `chainID`

fabioberger:
Replace mentions of `networkID` with `chainID` and start referencing Ganache by chainID `1337` not `networkID` 50.


zeroex/orderwatch/order_watcher.go
```diff
@@ -95,7 +95,7 @@ type Config struct {
 	MeshDB            *meshdb.MeshDB
 	BlockWatcher      *blockwatch.Watcher
 	OrderValidator    *ordervalidator.OrderValidator
-	NetworkID         int
+	ChainID         int
```
albrow: This change means that users upgrading from previous versions will see an error message about being on the wrong chain. The previous value was saved in the `NetworkID` field. If the `ChainID` field doesn't exist in the DB it will be loaded/interpreted as `0`.

Can we do one of the following?

1) Change this field name back to `NetworkID` since it's only an internal implementation detail and isn't exposed through any public APIs. We could also leave a comment explaining the situation.
2) Make a not in the CHANGELOG that wiping your database is required before upgrading.

fabioberger: I'll add a note about wiping the DB being required.

Issue #430, opened 2019-09-30
infra: Implement load balancing for bootstrap nodes and relay nodes.


albrow:
Right now we use one machine for each bootstrap node (and our bootstrap nodes are also the only relay nodes in the network). Ideally, we could separate bootstrap nodes and relay nodes as described in #383. Then, we should create a load-balancing cluster where each peer ID/multiaddress is being serviced by potentially more than one machine under the hood.

opaolini:
For load-balancing the nodes we can utilize either dedicated cloud services (for example AWS's Elastic Load Balancer [0]) or modern reverse-proxies / load-balancers such as Traefik [1]

Both AWS ELB [2] and Traefik [3] support Websockets and cookie-based sticky sessions, which would correctly route clients to servers during HTTP upgrade negotiation after the connection has been lost.

When a load balancer will first receive a client request it will generate a cookie that encodes an information about the selected target server for this particular client and will include that cookie in the response. The client then needs to include that cookie in the subsequent requests.

When target servers will become unhealthy both ELB and Traefik will properly select the new target and update the cookie. 

We would need to make sure that the JSON-RPC Websocket Clients pass the cookies.

[0] - https://aws.amazon.com/elasticloadbalancing/
[1] - https://traefik.io
[2] - https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#sticky-sessions
[3] - https://docs.traefik.io/routing/services/#sticky-sessions

albrow:
@opaolini I'm not sure if cookie based sessions will work. Mesh currently uses TCP or WebSockets (with a preference for TCP if possible) as the underlying transport for communication with peers (including bootstrap nodes and relay nodes). Cookies might work in the WebSockets transport but I don't think it works for TCP?

The JSON-RPC API is separate so we don't need to worry about it for now. Bootstrap nodes and relay nodes will not even enable that API. It will potentially come into play when we are building Mesh Gateway or SRA v3.

I do still need to make the relay-only binary. I'll make a separate issue for that.

opaolini:
@albrow AWS ELB [0] and Traefik since version 2.0 [1] offer TCP Load Balancing, so that shouldn't be an issue.

I will evaluate both ELB and Traefik during the load tests, we just have to settle on some scenarios.

[0] - https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html#network-load-balancer-overview
[...]

PR #486, opened 2019-10-31
Add SQL Datastore option to Bootstrap Node

opaolini:
About
===
This PR aims to solve https://github.com/0xProject/0x-mesh/issues/430

The idea is to allow the bootstrap nodes to be scaled around a shared state (`dhtStore` and `peerStore`)

As part of the PR:
- Added Bootstrap node configuration option to pick the underlying datastore for persistance
- Added https://github.com/opaolini/go-ds-sql as a datastore which uses PostgreSQL as the underlying datastore
- Added configurable persistent  `peerStore`


cmd/mesh-bootstrap/main.go
```diff
@@ -57,8 +57,27 @@ type Config struct {
...
-	// DataDir is the directory used for storing data.
+	// DataStoreType is the data store which will be used to store DHT data
+	// for the bootstrap node.
+	// DataStoreType can be either: leveldb or postgres
+	DataStoreType string `envvar:"DATA_STORE_TYPE" default:"leveldb"`
+	// DataDir is the directory used for storing data when using leveldb as data store type.
 	DataDir string `envvar:"DATA_DIR" default:"0x_mesh"`
+	// DataDBHost is the database host used to connect to the database when
+	// using postgres as data store type.
+	DataDBHost string `envvar:"DATA_DB_HOST" default:"localhost"`
```
albrow: Can we organize all of these into a [connection URL/connection string](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) to avoid having so many different env vars? Something like `postgresql://postgres@localhost:5432/datastore`.

cmd/mesh-bootstrap/main.go
```diff
@@ -115,14 +122,55 @@ func main() {
...
 	var kadDHT *dht.IpfsDHT
-	newDHT := func(h host.Host) (routing.PeerRouting, error) {
-		var err error
-		dhtDir := getDHTDir(config)
-		kadDHT, err = p2p.NewDHT(ctx, dhtDir, h)
-		if err != nil {
-			log.WithField("error", err).Fatal("could not create DHT")
+	var newDHT func(h host.Host) (routing.PeerRouting, error)
+
+	// TODO(oskar) - Figure out why returning an anonymous function from
```
albrow: @opaolini can you elaborate on what you are talking about here?

p2p/opts.go
```diff
@@ -115,3 +116,9 @@ func NewDHT(ctx context.Context, storageDir string, host host.Host) (*dht.IpfsDH
 	return dht.New(ctx, host, dhtopts.Datastore(store), dhtopts.Protocols(dhtProtocolID))
 
 }
+
+// NewDHTWithDatastore returns a new Kademlia DHT instance configured with store
+// as the persistant storage interface.
+func NewDHTWithDatastore(ctx context.Context, store datastore.Batching, host host.Host) (*dht.IpfsDHT, error) {
```
albrow: I don't think we necessarily need to put this here. The main point of __p2p/opts.go__ and __p2p/opts_js.go__ is to allow for different options when running Mesh in the browser and running in standalone mode. Since it looks like this is only used for bootstrap nodes we could move it to __cmd/mesh-bootstrap__.

cmd/mesh-bootstrap/storage.go
```diff
@@ -0,0 +1,137 @@
...
+	"path/filepath"
+
+	"github.com/0xProject/0x-mesh/keys"
+	"github.com/0xProject/0x-mesh/p2p"
+	p2pcrypto "github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/routing"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
+	peerstore "github.com/libp2p/go-libp2p-peerstore"
+
```
albrow: nit: Unnecessary space.

cmd/mesh-bootstrap/storage.go
```diff
@@ -0,0 +1,137 @@
...
+	case postgresDataStore:
+		sqlOpts := &sqlds.Options{
+			Host:     config.DataDBHost,
+			Port:     config.DataDBPort,
+			User:     config.DataDBUser,
+			Password: config.DataDBPassword,
+			Database: config.DataDBDatabaseName,
+			Table:    "peerStore",
+		}
+		store, err := sqlOpts.CreatePostgres()
```
albrow: It looks like we are using multiple instances of a datastore. For LevelDB we needed to do this in some cases in order to use a separate directory (LevelDB does not support separate databases). Can we share the same instance of `store` across the codebase? This could also simplify a lot of the code since once we have the `store`, a lot of things are the same.

cmd/mesh-bootstrap/storage.go
opaolini: removed.

cmd/mesh-bootstrap/storage.go
opaolini: The idea behind new datastore was that for every Key-Value store we use a different table to seperate the key space, which could make things easier and faster for querying data.

If for example we would like to grab provider data from the `dhtStore` (which has only a few entries) we wouldn't have to run the `SELECT` query against a table with hundreds of `peerStore` metadata rows present.

p2p/opts.go
opaolini: moved to **cmd/mesh-bootstrap**

cmd/mesh-bootstrap/main.go
opaolini: Maybe we could provide both options for Datastore creation in the `sql-datastore`? My idea to use seperate connection settings from a "deployment" and configuration management perspective is that:
- the password can be stored seperately (like in Kubernetes `secret` objects)
- the configuration is reusable / composable, for example if we have multiple bootstrap nodes the `DB_HOST`, `DB_PORT` will be common, however the `DB_USER` and `DB_PASSWORD` will be specific to each instance. And in the case of a db migration to a new host we wouldn't need to update different connection strings.

cmd/mesh-bootstrap/main.go
opaolini: So previously I had here:
```go
var kadDHT *dht.IpfsDHT
newDHT := getNewDHT(ctx, config, kadDHT)
```

However this caused panics with nil pointer dereference for `*dht.IpfsDHT` Method, I did not dive too deep into what exactly happens there.

cmd/mesh-bootstrap/main.go
albrow: If it were me managing deployments, I would much prefer having one environment variable instead of many. Maybe it's subjective though. Do whatever you think is best.

I don't like having both `DB_URL` and `DB_HOST`, `DB_PORT`, `DB_PASSWORD`, etc.

cmd/mesh-bootstrap/storage.go
albrow: I understand the reasoning for using a different table, but that doesn't require multiple connections to the same database. You can access many different tables with a single connection. If this is too much work or isn't supported by `github.com/opaolini/go-ds-sql` I'm okay with creating an issue for it and addressing it later.

cmd/mesh-bootstrap/main.go
albrow: It's hard for me to understand exactly what's happening without looking at the state of the codebase when the error occurred. I'm okay with just leaving it for now.

cmd/mesh-bootstrap/storage.go
opaolini: We can just use multiple datastores using the same `*sql.DB`, added an exported `NewQueriesForTable` in `sql-datastore` here: https://github.com/0xProject/sql-datastore/pull/1/commits/5301c2dd1bf697de5655b58a2f35fc8fdafe0780

Which allows us to:
```go
db, err := sql.Open("postgres", constr)
if err != nil {
  return nil, err
}

dhtStore, _ := sqlds.NewDatastore(db, sqlds.NewQueriesForTable("dhtkv"))
peerStore, _ := sqlds.NewDatastore(db, sqlds.NewQueriesForTable("peerkv"))
```

albrow: Great. @opaolini that looks good to me.

albrow:
Superseded by https://github.com/0xProject/0x-mesh/pull/493.

Issue #419, opened 2019-09-17
core: Remove `OrderExpirationBuffer` config


fabioberger:
Currently, we allow the operator to customize a buffer they'd like added to UTC time when validating an order's expiry. This buffer is applied when validating orders submitted via JSON-RPC as well as orders submitted from other peers. If peers have this set to different values, a peer with a lower buffer would consider the orders from a peer with a higher buffer invalid and penalize the peer for sharing those orders. We should either remove this configuration or alternatively, make it only apply to the JSON-RPC OrderEvents subscription, and _not_ the validation performed when sharing orders with peers.

Issue #418, opened 2019-09-17
rpc,orderwatch: Add blockTimestamp expiration events


fabioberger:
Currently we emit order expiration events when the order's expirationTimestamp is less than current UTC + buffer. Some users of Mesh, more notably MMers would prefer to receive an event when the order's expirationTimestamp is less than the latest blockTimestamp. We should add this as an additional event and let the subscriber decide which to listen for.

PR #490, opened 2019-11-04
Remove expiration buffer config and emit expiry events based on block timestamp

fabioberger:
Fixed: https://github.com/0xProject/0x-mesh/issues/419 and https://github.com/0xProject/0x-mesh/issues/418

This PR:

- Refactors Mesh to emit order expiration events when orders are expired according to the latest block timestamp. We no longer emit UTC-based expiration events. 
- An order can now become unexpired, if a block re-org causes the latest block timestamp to be earlier than it's predecessor. We therefore added the `UNEXPIRED` order event type.
- Removes the expiration buffer config as it is no longer needed.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -397,8 +397,8 @@ func (app *App) Start(ctx context.Context) error {
 			select {
 			case <-innerCtx.Done():
 				return
-			case <-ticker.C:
-				expiredSnapshots := app.snapshotExpirationWatcher.Prune(time.Now())
+			case now := <-ticker.C:
+				expiredSnapshots := app.snapshotExpirationWatcher.Prune(now)
 				for _, expiredSnapshot := range expiredSnapshots {
 					app.muIdToSnapshotInfo.Lock()
 					delete(app.idToSnapshotInfo, expiredSnapshot.ID)
--- a/core/core.go
+++ b/core/core.go
@@ -239,14 +239,24 @@ func New(config Config) (*App, error) {
 		return nil, err
 	}
 
+	latestHeader, err := meshDB.FindLatestMiniHeader()
+	if err != nil {
+		return nil, err
+	}
+	var latestBlockTimestamp *time.Time
+	if latestHeader != nil {
+		latestBlockTimestamp = &latestHeader.Timestamp
+	}
+
 	// Initialize order watcher (but don't start it yet).
 	orderWatcher, err := orderwatch.New(orderwatch.Config{
-		MeshDB:            meshDB,
-		BlockWatcher:      blockWatcher,
-		OrderValidator:    orderValidator,
-		ChainID:           config.EthereumChainID,
-		MaxOrders:         config.MaxOrdersInStorage,
-		MaxExpirationTime: metadata.MaxExpirationTime,
+		MeshDB:               meshDB,
+		BlockWatcher:         blockWatcher,
+		OrderValidator:       orderValidator,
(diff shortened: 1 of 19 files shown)


expirationwatch/expiration_watcher.go
```diff
@@ -84,49 +72,9 @@ func (w *Watcher) Remove(expirationTimestamp time.Time, id string) {
...
-// expired items. The channel will be closed if/when the watcher is done
-// watching.
-func (w *Watcher) ExpiredItems() <-chan []ExpiredItem {
-	return w.expiredItems
-}
-
-// prune checks for any expired items, removes them from the expiration watcher and returns them
-// to the caller
-func (w *Watcher) prune() []ExpiredItem {
+// Prune checks for any expired items given the latest blockTimestamp,
```
albrow: `expirationwatch.Watcher` is used for other things besides block timestamps right? Can we rename this parameter to reflect that?

core/core.go
```diff
@@ -237,27 +234,35 @@ func New(config Config) (*App, error) {
...
-		ExpirationBuffer:  config.OrderExpirationBuffer,
-		MaxOrders:         config.MaxOrdersInStorage,
-		MaxExpirationTime: metadata.MaxExpirationTime,
+		MeshDB:               meshDB,
+		BlockWatcher:         blockWatcher,
+		OrderValidator:       orderValidator,
+		ChainID:              config.EthereumChainID,
+		MaxOrders:            config.MaxOrdersInStorage,
+		MaxExpirationTime:    metadata.MaxExpirationTime,
+		LatestBlockTimestamp: latestBlockTimestamp,
```
albrow: Does `LatestBlockTimestamp` really need to be passed in as part of `Config` on startup? It seems like:

1. Unless there was a sudden unexpected shutdown, we will have already pruned any expired orders when the block watcher notified us of this block before Mesh was shutdown.
2. The block watcher is going to notify us of the next latest block in a matter of seconds (or less) anyways.

zeroex/orderwatch/order_watcher.go
```diff
@@ -89,16 +88,17 @@ type Watcher struct {
 	maxExpirationTime          *big.Int
 	maxExpirationCounter       *slowcounter.SlowCounter
 	maxOrders                  int
+	latestBlockTimestamp       *time.Time
```
albrow: I don't think we need a pointer to `time.Time` here.

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -667,6 +667,62 @@ func TestOrderWatcherERC20PartiallyFilled(t *testing.T) {
 	assert.Equal(t, false, orders[0].IsRemoved)
 	assert.Equal(t, halfAmount, orders[0].FillableTakerAssetAmount)
 }
+func TestOrderWatcherOrderExpiredThenUnexpires(t *testing.T) {
```
albrow:
```suggestion
func TestOrderWatcherOrderExpiredThenUnexpired(t *testing.T) {
```

core/core.go
fabioberger: We could potentially miss a block re-org caused `Unexpired` event emission since on restart the OrderWatcher would not know what the `latestBlockTimestamp` was before the shutdown. It's an edge-case for sure, I just wanted to make sure the implementation was bullet proof. If you don't want to handle this edge-case, I can get rid of it.

PR #493, opened 2019-11-05
Add SQL Datastore option to Bootstrap Node

albrow:
This PR is the same as https://github.com/0xProject/0x-mesh/pull/486 but rebased off of the `development` branch.

### DIFF
--- a/cmd/mesh-bootstrap/main.go
+++ b/cmd/mesh-bootstrap/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/0xProject/0x-mesh/loghooks"
 	"github.com/0xProject/0x-mesh/p2p"
 	"github.com/0xProject/0x-mesh/p2p/banner"
+	"github.com/ipfs/go-datastore"
 	libp2p "github.com/libp2p/go-libp2p"
 	autonat "github.com/libp2p/go-libp2p-autonat-svc"
 	circuit "github.com/libp2p/go-libp2p-circuit"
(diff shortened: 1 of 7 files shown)


cmd/mesh-bootstrap/main.go
```diff
@@ -66,8 +71,35 @@ type Config struct {
 	// P2PAdvertiseAddrs is a comma separated list of libp2p multiaddresses which the
 	// bootstrap node will advertise to peers.
 	P2PAdvertiseAddrs string `envvar:"P2P_ADVERTISE_ADDRS"`
-	// DataDir is the directory used for storing data.
+	// DataStoreType is the data store which will be used to store DHT data
```
albrow: Looking at these env vars again, I think it is confusing that they aren't separated in any way. Can we prefix the ones related to Postgres/SQL with `POSTGRES` or `SQL` and the ones related to LevelDB with `LEVEL
DB`?

For example `LEVELDB_DATA_DIR` and `SQL_DB_CONNECTION_STRING`.

opaolini: That makes sense, renamed.

albrow:
LGTM! Let's just make sure CI passes.

integration-tests/integration_test.go
```diff
@@ -308,7 +308,7 @@ func startStandaloneNode(t *testing.T, ctx context.Context, logMessages chan<- s
 	cmd.Env = append(
 		os.Environ(),
 		"VERBOSITY=5",
-		"DATA_DIR="+standaloneDataDir,
+		"LEVELDB_DATA_DIR="+standaloneDataDir,
```
albrow: For `mesh` (not `mesh-bootstrap`) the environment variable is still called `DATA_DIR`.

PR #494, opened 2019-11-05
decoder: Add support for Axie Infinity's ERC721 Transfer/Approve events

fabioberger:
Axie Infinity incorrectly implemented the ERC721 events `Transfer` and `Approve`, omitting the index on `_tokenId`. Since we've been seeing a lot of orders added to Mesh for this specific contract, we've decided to add support for decoding ERC721 events with this variant of indexes.


zeroex/orderwatch/decoder/event_decoder.go
```diff
@@ -33,6 +33,9 @@ const erc20EventsAbi = "[{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"na
...
 
+// Includes ERC721 `Transfer` and `Approval` as specified in Axie Infinity contract (without an index on TokenID)
+const incorrectERC721EventsAbi = "[{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"name\":\"_from\",\"type\":\"address\"},{\"indexed\":true,\"name\":\"_to\",\"type\":\"address\"},{\"indexed\":false,\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"Transfer\",\"type\":\"event\"},{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"name\":\"_owner\",\"type\":\"address\"},{\"indexed\":true,\"name\":\"_approved\",\"type\":\"address\"},{\"indexed\":false,\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"Approval\",\"type\":\"event\"}]"
```
albrow: Maybe we could rename this to distinguish it from other forms of incorrect implementations of the ERC721 spec?

```suggestion
const erc721EventsAbiWithoutTokenIDIndex = "[{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"name\":\"_from\",\"type\":\"address\"},{\"indexed\":true,\"name\":\"_to\",\"type\":\"address\"},{\"indexed\":false,\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"Transfer\",\"type\":\"event\"},{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"name\":\"_owner\",\"type\":\"address\"},{\"indexed\":true,\"name\":\"_approved\",\"type\":\"address\"},{\"indexed\":false,\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"Approval\",\"type\":\"event\"}]"
```

zeroex/orderwatch/decoder/event_decoder.go
```diff
@@ -574,7 +584,13 @@ func (d *Decoder) decodeERC721(log types.Log, decodedLog interface{}) error {
 
 	err := unpackLog(decodedLog, eventName, log, d.erc721ABI)
 	if err != nil {
-		return err
+		if _, ok := err.(UnsupportedEventError); ok {
+			// Try unpacking using the incorrect ERC721 event ABIs
+			err := unpackLog(decodedLog, eventName, log, d.incorrectERC721EventsABI)
+			if err != nil {
+				return err
```
albrow: It might be helpful to combine the two error messages here in order to get a clearer picture of what went wrong.

Something like:

```go
	erc721Err := unpackLog(decodedLog, eventName, log, d.erc721ABI)
	if erc721Err != nil {
		if _, ok := erc721Err.(UnsupportedEventError); ok {
			// Try unpacking using the incorrect ERC721 event ABIs
			fallbackErr := unpackLog(decodedLog, eventName, log, d.incorrectERC721EventsABI)
			if fallbackErr != nil {
				return fmt.Errorf("could not decode event with either correct ERC721 ABI or fallback ABI: %s %s", erc721Err.Error(), fallbackErr.Error())
			}
		}
	}
```

I'm not sure the best way to name these error variables or phrase the error message.

fabioberger: We need to throw the `UnsupportedEventError` type if it is an unsupported event since we want to log these to then add custom support for other incorrect standard implementations. Because of this, I don't think this makes a lot of sense. All the caller needs to know if that the we were unable to decode the event given the logic in place.

albrow: @fabioberger the problem with the way it is currently written is that we *don't* return the error that is associated with trying to unpack the event as a standard ERC721 event. We _only_ are returning the error associated with trying to unpack the event using `incorrectERC721EventsABI`. I'm worried that this could obscure some issues.

PR #496, opened 2019-11-05
Fix bug in Go RPC client when unmarshaling contract events

albrow:
This PR fixes a bug in the GO RPC client  that was most likely introduced in `5.0.0-beta`. The bug occurred when subscribing to order events and receiving one with one or more contract events:

```
json: cannot unmarshal object into Go struct field ContractEvent.Parameters of type zeroex.ContractEventParameters
```

This PR fixes the bug by adding an `UnmarshalJSON` event for each type of contract event.

### DIFF
--- a/zeroex/order.go
+++ b/zeroex/order.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/0xProject/0x-mesh/ethereum/signer"
 	"github.com/0xProject/0x-mesh/ethereum/wrappers"
+	"github.com/0xProject/0x-mesh/zeroex/orderwatch/decoder"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
 	gethsigner "github.com/ethereum/go-ethereum/signer/core"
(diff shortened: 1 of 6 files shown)


zeroex/order.go
```diff
@@ -88,7 +84,18 @@ type ContractEvent struct {
 	IsRemoved  bool
 	Address    common.Address
 	Kind       string
-	Parameters ContractEventParameters
+	Parameters interface{}
```
albrow: @fabioberger changed this back to `interface{}` because for some event types, we just use the default `json.Marshal` behavior (which is more efficient and saves us some lines of code).

zeroex/order_test.go
```diff
@@ -58,7 +59,22 @@ func TestMarshalUnmarshalOrderEvent(t *testing.T) {
 		SignedOrder:              signedOrder,
 		EndState:                 ESOrderAdded,
 		FillableTakerAssetAmount: big.NewInt(2000),
-		ContractEvents:           []*ContractEvent{},
+		ContractEvents: []*ContractEvent{
```
albrow: @fabioberger we didn't catch this bug because we never tried marshaling/unmarshaling an `OrderEvent` with a non-empty `ContractEvents` field.

PR #501, opened 2019-11-06
docs: Fix docker-compose.yml example

fabioberger:
Now that we use `RPC_ADDR` to specify the port, we need to set it to `mesh:60557` since the default is now `localhost:*` but we specifically port map `60557`.

### DIFF
--- a/docs/deployment_with_telemetry.md
+++ b/docs/deployment_with_telemetry.md
@@ -82,7 +82,24 @@ services:
 ```
 
 In most cases, the only change you need to make to the __docker-compose.yml__
-file is to set `ETHEREUM_RPC_URL` to your own Ethereum JSON RPC endpoint.
+file is to set `ETHEREUM_RPC_URL` to your own Ethereum JSON RPC endpoint. The 
+`RPC_ADDR` above will allow any Docker containers running in the same Docker 
+Compose file to access the Mesh RPC API via 
+[links](https://docs.docker.com/compose/networking/#links). To use this feature, 
+be sure to add the following line to any containers you wish to access the Mesh 
+RPC API from:
+
+```
+links:
+    - mesh
+```
+
+You can then use the URL `ws://mesh:60557` to access the RPC API.
+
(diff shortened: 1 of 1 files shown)


docs/deployment_with_telemetry.md
```diff
@@ -53,6 +53,7 @@ services:
         environment:
             - VERBOSITY=5
             - ETHEREUM_CHAIN_ID=1
+            - RPC_ADDR=0.0.0.0:60557
```
albrow:
```suggestion
            - RPC_ADDR=127.0.0.1:60557
```

Let's set this to `127.0.0.1` by default and add a comment about how to expose your API to the public or to your own servers if you want to.

docs/deployment_with_telemetry.md
```diff
@@ -82,10 +82,7 @@ services:
 ```
 
 In most cases, the only change you need to make to the __docker-compose.yml__
-file is to set `ETHEREUM_RPC_URL` to your own Ethereum JSON RPC endpoint. If you 
-wish to connect to your Mesh node over a subnet or over the public internet, you
- must set the `RPC_ADDR` to something other than `localhost` (e.g., `0.0.0.0` for 
- the public internet).
+file is to set `ETHEREUM_RPC_URL` to your own Ethereum JSON RPC endpoint.
```
albrow: I think we should still have a comment here. Something like:

> The `RPC_ADDR` above will allow any Docker containers running in the same Docker Compose file to access the Mesh RPC API via ["links"](https://docs.docker.com/compose/networking/#links). To use this feature, be sure to add the following line to any containers you wish to access the Mesh RPC API:
>
> ```
> links:
>     - mesh
> ```
>
> You can then use the URL `ws://mesh:60557` to access the RPC API.
>
> Alternatively, if you want to open up your Mesh RPC API to the public internet, you can set `RPC_ADDR=0.0.0.0:60557`. If you choose to go this route, we strongly recommend using an external firewall to restrict who can access your RPC API.
>

PR #503, opened 2019-11-06
Add a README for both browser and Go examples

albrow:
This will also help fix an issue with broken links on GitBook (seems that when linking to directories, GitBook assumes that a __README.md__ file exists and tries to link to it even if it's not there).

### DIFF
--- a/examples/browser/README.md
+++ b/examples/browser/README.md
@@ -1,0 +1,29 @@
+## 0x Mesh Browser Example
+
+This directory contains an example of how to run Mesh in the browser.
+
+### Running the Example
+
+To run the example, first change into the __examples/browser__ directory and
+then run:
+
+```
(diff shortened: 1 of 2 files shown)


examples/browser/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+This directory contains an example of how to run Mesh in the browser.
+
+### Running the Example
+
+To run the example, first change into the __examples/browser__ directory and
+then run:
+
+```
+yarn install --force && yarn build
```
fabioberger: Why `--force`?

albrow: `--force` is required because we use a local directory for `@0x/mesh-browser` in the __package.json__ file for the example. This makes it so that you are always using the version of Mesh that was intended for the example code.

PR #505, opened 2019-11-07
ethereum/ratelimit: Fix bug in for timezones behind UTC

albrow:
The following error was occurring at certain times of day for timezones that are _behind_ UTC:

```
{
  "error_string": "rate: Wait(n=1) would exceed context deadline",
  "level": "error",
  "msg": "core app exited with error",
  "myPeerID": "16Uiu2HAkws8tB3zR8m7PQuV5iDNDfYXfShnWBLUeVpdefuTs8mqh",
  "time": "2019-11-06T17:38:55-08:00"
}
```

To be more specific, the bug seems to occur whenever the current local time is _before_ midnight, but the time in UTC is the next day _after_ midnight. When Mesh enters this state, no Ethereum RPC requests are allowed and Mesh cannot really function effectively.

### DIFF
--- a/ethereum/ratelimit/rate_limiter.go
+++ b/ethereum/ratelimit/rate_limiter.go
@@ -218,5 +218,6 @@ func (r *rateLimiter) getGrantedInLast24hrsUTC() int {
 }
 
 func getUTCMidnightOfDate(date time.Time) time.Time {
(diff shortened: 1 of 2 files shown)


albrow:
@fabioberger I can manually confirm this fixes the bug, but ideally we should write a regression test to prevent this from popping up again. I can't figure out how to do that, and the existing tests always pass in my local timezone, regardless of time of day. Any ideas?

PR #508, opened 2019-11-07
core,docs: Enforce reasonable configs for BlockPollingInterval and EthereumRPCMaxRequestsPer24HrUTC

fabioberger:
I noticed that I had accidentally started my Kovan Mesh node with `BlockPollingInterval` set to `300ms` and `EthereumRPCMaxRequestsPer24HrUTC` defaulted to 100k requests. The math didn't add up and my node would definitively use up it's 24hr request limit and then stall. This PR does some math on Mesh startup to make sure these two configs won't eventually cause the Mesh node to malfunction due to using up all allowed requests for block polling alone.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -43,12 +43,12 @@ import (
 )
 
 const (
-	blockWatcherRetentionLimit           = 20
-	ethereumRPCRequestTimeout            = 30 * time.Second
-	peerConnectTimeout                   = 60 * time.Second
-	checkNewAddrInterval                 = 20 * time.Second
-	expirationPollingInterval            = 50 * time.Millisecond
-	rateLimiterCheckpointInterval        = 1 * time.Minute
+	blockWatcherRetentionLimit    = 20
+	ethereumRPCRequestTimeout     = 30 * time.Second
+	peerConnectTimeout            = 60 * time.Second
+	checkNewAddrInterval          = 20 * time.Second
+	expirationPollingInterval     = 50 * time.Millisecond
+	rateLimiterCheckpointInterval = 1 * time.Minute
 	// Computed with default blockPollingInterval (5s), and EthereumRPCMaxRequestsPer24HrUTC (100k)
 	defaultNonPollingEthRPCRequestBuffer = 82720
 	// logStatsInterval is how often to log stats for this node.
--- a/core/core.go
+++ b/core/core.go
@@ -189,10 +189,7 @@ func New(config Config) (*App, error) {
 	config = unquoteConfig(config)
 
 	// Ensure ETHEREUM_RPC_MAX_REQUESTS_PER_24_HR_UTC is reasonably set given BLOCK_POLLING_INTERVAL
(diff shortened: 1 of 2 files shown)


core/core.go
```diff
@@ -186,6 +188,27 @@ func New(config Config) (*App, error) {
 	}
 	config = unquoteConfig(config)
 
+	// Ensure ETHEREUM_RPC_MAX_REQUESTS_PER_24_HR_UTC is reasonably set given BLOCK_POLLING_INTERVAL
+	var per24HrPollingRequests int
+	if config.BlockPollingInterval > 1*time.Second {
```
albrow: We should remove this if statement. The math is the same regardless of whether `config.BlockPollingInterval` is greater than 1 second. The way it's currently written means that a duration of 1.5 seconds is truncated to 1 second which yields incorrect results. I think the math is also more complicated than it needs to be.

How about this instead?

```go
per24HrPollingRequests = int((24 * time.Hour) / config.BlockPollingInterval)
```

fabioberger: The issue with your proposal is that it overflows time.Duration (see: https://play.golang.org/p/VaAauCNj1i4. I've gone ahead and removed the conditional statement. We now always compute from perSecond which works.

albrow: My solution doesn't overflow `time.Duration`. You just needed to add parentheses in your example. See https://play.golang.org/p/nDSxy5hAt5O

fabioberger: Ah! Amazing. Ok, let's do this then.

docs/deployment.md
```diff
@@ -43,7 +43,8 @@ machine where all Mesh-related data will be stored.
 -   Ports 60557, 60558, and 60559 are the default ports used for the JSON RPC endpoint, communicating with peers over TCP, and communicating with peers over WebSockets, respectively.
 -   In order to disable P2P order discovery and sharing, set `USE_BOOTSTRAP_LIST` to `false`.
 -   Running a VPN may interfere with Mesh. If you are having difficulty connecting to peers, disable your VPN.
--   If you are running against a POA testnet (e.g., Kovan), you might want to shorten the `BLOCK_POLLING_INTERVAL` since blocks are mined more frequently then on mainnet.
+-   If you are running against a POA testnet (e.g., Kovan), you might want to shorten the `BLOCK_POLLING_INTERVAL` since blocks are mined more frequently then on mainnet. If you do this, your node will use more Ethereum RPC calls, so you will also need to adjust the `ETHEREUM_RPC_MAX_REQUESTS_PER_24_HR_UTC` to around `380000` (*warning:* this will not fit within Infura's
```
albrow: Since you didn't specify how much to reduce `BLOCK_POLLING_INTERVAL` by, including a specific number that you would need for `ETHEREUM_RPC_MAX_REQUESTS_PER_24_HR_UTC` doesn't make a lot of sense. Can we just warn about setting `ETHEREUM_RPC_MAX_REQUESTS_PER_24_HR_UTC` higher than 100k?

PR #527, opened 2019-11-14
rpc: Upgrade BigNumber.js version used by Mesh TS Client to ~9.0.0

fabioberger:
`@0x/mesh-rpc-client` is a dependency to `@0x/orderbook` and `@0x/asset-swapper` and we have been asked by Augur to upgrade our BigNumber.js dependency in our tooling (https://github.com/0xProject/0x-monorepo/pull/2331) -- and our version is quite old. 

In order to be able to refactor the monorepo to the new BigNumber version, it must first be updated in `@0x/mesh-rpc-client`. That is what this PR does. We will eventually switch back to using `@0x/utils` after the 0x packages have been published.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
+## v7.0.3-beta
```
albrow:
```suggestion
## v7.0.3-beta-0xv3
```

rpc/clients/typescript/src/types.ts
```diff
@@ -1,5 +1,5 @@
 import { SignedOrder } from '@0x/types';
-import { BigNumber } from '@0x/utils';
```
albrow: Is it okay to import `bignumber.js` directly here? IIRC in the past we were configuring `bignumber.js` a certain way in `@0x/utils` so we needed to always import from there.

fabioberger: That is a really good point. Thanks for catching this. I've gone ahead and temporarily added the same BigNumber configs that are in `@0x/utils` to our client.

PR #530, opened 2019-11-14
mesh-bootstrap: expose configuration options for the peer store cache

opaolini:
About
===

This PR allows us to configure `peerstore`'s cache and garbage collection options using environment variables.

This is mainly useful for disabling cache for nodes that are running around a shared state.

The default options are taken from https://github.com/libp2p/go-libp2p-peerstore/blob/master/pstoreds/peerstore.go#L41


albrow:
@opaolini the code changes look good to me. Is it okay if we wait a bit before merging? I want to make sure that this change is actually necessary after we have a better understanding for how the peerstore works and how it fits in with the DHT.

opaolini:
@albrow Definitely, I am closing this for now and will reopen as a draft PR.

Issue #532, opened 2019-11-15
Add full support for persistence in browsers


albrow:
As of now, Mesh stores its own peer ID in `localStorage` but doesn't persist any other data when running in the browser (orders, DHT, peerstore, etc).

We made a lot of progress toward browser persistence in https://github.com/0xProject/0x-mesh/issues/297 and https://github.com/0xProject/0x-mesh/pull/335 but ultimately had to put our efforts on hold when we hit a roadblock having to do with the limitations of `localStorage`.

PR #533, opened 2019-11-15
db, browser: Add support for persistence in the browser

albrow:
Fixes https://github.com/0xProject/0x-mesh/issues/532.

This PR adds support for persistence in browser environments. This is accomplished by using [BrowserFS](https://github.com/jvilk/BrowserFS) backed by [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API). The bulk of the work is actually in our fork of goleveldb, where we have [an open PR](https://github.com/syndtr/goleveldb/pull/288).

This solves the limitations of using localStorage as the backing for BrowserFS (limited to 10MB). This PR also includes code which will fallback to using an in-memory database if BrowserFS is not available. This is important because it allows us to continue to use `wasm-browser-test` for testing (`wasm-browser-test` does not offer any way to sideload JavaScript dependencies).

BrowserFS+IndexedDB support is included in the integration tests and I have also manually tested that it works with the browser example.

### DIFF
--- a/db/open_js.go
+++ b/db/open_js.go
@@ -53,7 +53,7 @@ func openBrowserFSDB(path string) (*DB, error) {
 	// variable. This is definitely a bit of a hack and wastes some CPU resources,
 	// but it is also extremely reliable. Given that we have a chicken and egg
 	// problem with both Wasm and JavaScript code loading and executing at the
-	// same time, it is difficult to match this level reliability with something
+	// same time, it is difficult to match this level of reliability with something
 	// like callback functions or events.
 	start := time.Now()
 	for {
--- a/db/open_js.go
+++ b/db/open_js.go
@@ -61,6 +61,7 @@ func openBrowserFSDB(path string) (*DB, error) {
 			return nil, errors.New("timed out waiting for BrowserFS to load")
 		}
 		if js.Global().Get("browserFS") != js.Undefined() && js.Global().Get("browserFS") != js.Null() {
+			log.Info("BrowserFS finished loading")
 			break
 		}
 		time.Sleep(browserFSLoadCheckInterval)
--- a/db/open_js.go
+++ b/db/open_js.go
@@ -3,14 +3,69 @@
 package db
 
 import (
+	"errors"
+	"syscall/js"
+	"time"
+
+	log "github.com/sirupsen/logrus"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/storage"
 )
 
-// Open creates a new database using in-memory storage. The given file path is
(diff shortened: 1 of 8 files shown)


albrow:
Looking into issues with the integration test. Hopefully it's an easy fix.

albrow:
CI issues resolved. Removing WIP tag.

db/open_js.go
```diff
@@ -3,13 +3,39 @@
...
 
-// Open creates a new database using in-memory storage. The given file path is
-// ignored.
+const (
+	// browserFSLoadCheckInterval is frequently to check whether browserFS is
+	// loaded.
+	browserFSLoadCheckInterval = 50 * time.Millisecond
+	// browserFSLoadTimeout is how long to wait for BrowserFS to finish loading
+	// before giving up.
+	browserFSLoadTimeout = 5 * time.Second
```
fabioberger: Could the length of time it takes to load `browserFS` be influenced by a users internet speed? If so, any harm in setting this wait time higher?

db/open_js.go
```diff
@@ -18,3 +44,33 @@ func Open(path string) (*DB, error) {
...
+
+func openBrowserFSDB(path string) (*DB, error) {
+	log.Info("BrowserFS deteceted. Using BrowserFS-backed databse.")
+	// Wait for browserFS to load.
+	//
+	// HACK(albrow): We do this by checking for the global browserFS
+	// variable. This is definitely a bit of a hack and wastes some CPU resources,
+	// but it is also extremely reliable. Given that we have a chicken and egg
+	// problem with both Wasm and JavaScript code loading and executing at the
+	// same time, it is difficult to match this level reliability with something
```
fabioberger:
```suggestion
	// same time, it is difficult to match this level of reliability with something
```

db/open_js.go
albrow: If you are using the `@0x/mesh-browser` package, then no, it will not be affected by your internet speed. The Wasm and JavaScript code will be bundled together and _downloaded_ at the same time. How long it takes to _parse and execute_ the Wasm and JavaScript code could potentially be influenced by your CPU speed. However, it is far more likely that the JavaScript code will execute first due to the fact that the Wasm bytecode is much larger and typically takes longer to parse.

I have considered the idea of offering an alternative way for users to run Mesh in the browser, which is to bundle the Wasm and JavaScript code separately. This would allow use of the [streaming API for WebAssembly](https://www.w3.org/TR/wasm-web-api-1/#streaming-modules) and could potentially improve performance. But for now it's not something I think we should worry about.

The downside to increasing the timeout is that it takes us longer to discover errors related to BrowserFS failing to load.

PR #534, opened 2019-11-17
[WIP] mesh-bootstrap: expose configuration options for the peer store cache

opaolini:
About
===

This PR allows us to configure `peerstore`'s cache and garbage collection options using environment variables.

This is mainly useful for disabling cache for nodes that are running around a shared state.

The default options are taken from https://github.com/libp2p/go-libp2p-peerstore/blob/master/pstoreds/peerstore.go#L41


opaolini:
As mentioned in https://github.com/0xProject/0x-mesh/pull/530, this is resubmitted as a draft until  we have a better understanding of peerstore and its relation to the DHT.

Issue #452, opened 2019-10-16
Investigate apparent memory leak


albrow:
I've been closely monitoring memory usage over the past week or so and it seems like Mesh has a memory leak.

![Screen Shot 2019-10-16 at 11 44 42 AM](https://user-images.githubusercontent.com/800857/66948947-7e406780-f00a-11e9-8151-2474102f26c0.png)

We can see by looking at the logs that the sharp drops in memory usage coincide with Mesh restarting. Most likely what is happening is that the OS is killing the Mesh process and then it automatically restarts due to `restart: always` in the __docker-compose.yml__ file.

albrow:
After 9 days of monitoring a Mesh node running uninterrupted, we can conclude that #539 did in fact fix the issue.

Here's a chart of memory usage over the past 7 days:

<img width="488" alt="Screen Shot 2019-11-27 at 4 24 41 PM" src="https://user-images.githubusercontent.com/800857/69767736-7956f380-1132-11ea-95a7-d5e71e39c085.png">

And a recent snapshot obtained via `go tool pprof`:

```
mesh
Type: inuse_space
Time: Nov 27, 2019 at 2:23am (PST)
Showing nodes accounting for 74.21MB, 100% of 74.21MB total
      flat  flat%   sum%        cum   cum%
   13.41MB 18.07% 18.07%    13.41MB 18.07%  github.com/0xProject/0x-mesh/vendor/github.com/syndtr/goleveldb/leveldb/util.(*BufferPool).Get
   12.81MB 17.26% 35.33%    13.87MB 18.70%  github.com/0xProject/0x-mesh/vendor/github.com/gogo/protobuf/io.(*varintWriter).WriteMsg
   12.67MB 17.07% 52.40%    12.67MB 17.07%  github.com/0xProject/0x-mesh/vendor/github.com/gogo/protobuf/io.(*varintReader).ReadMsg
      12MB 16.18% 68.58%       12MB 16.18%  github.com/0xProject/0x-mesh/vendor/github.com/syndtr/goleveldb/leveldb/memdb.New
   10.08MB 13.58% 82.16%    11.08MB 14.93%  github.com/0xProject/0x-mesh/vendor/github.com/whyrusleeping/timecache.(*TimeCache).Add
    1.58MB  2.13% 84.29%     1.58MB  2.13%  net/http.(*http2ClientConn).frameScratchBuffer
[...]

PR #539, opened 2019-11-18
Update go-libp2p-kad-dht to fix memory leak

albrow:
This is a potential solution to https://github.com/0xProject/0x-mesh/issues/452. Memory snapshots revealed that `go.opencensus.io/tag.(*Map).insert` was responsible for an unusually large amount of memory (up to 100MB after a few days). That package is used by `github.com/libp2p/go-libp2p-kad-dht`. https://github.com/libp2p/go-libp2p-kad-dht/issues/389 mentions this leak which was reported as fixed in version 0.2.1.

This PR updates `github.com/libp2p/go-libp2p-kad-dht` to the latest version. We will need to continue to monitor memory usage to see if this fixes the problem.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -319,13 +319,6 @@ func (n *Node) ID() peer.ID {
 // Start causes the Node to continuously send messages to and receive messages
 // from its peers. It blocks until an error is encountered or `Stop` is called.
 func (n *Node) Start() error {
-	// Note: dht.Bootstrap has a somewhat confusing name. It doesn't automatically
-	// connect to the bootstrap peers. It just starts the background process of
-	// searching for new peers.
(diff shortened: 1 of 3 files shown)


p2p/node.go
```diff
@@ -319,13 +319,6 @@ func (n *Node) ID() peer.ID {
 // Start causes the Node to continuously send messages to and receive messages
 // from its peers. It blocks until an error is encountered or `Stop` is called.
 func (n *Node) Start() error {
-	// Note: dht.Bootstrap has a somewhat confusing name. It doesn't automatically
```
albrow: Calling `Bootstrap` is no longer necessary in the latest version of `go-libp2p-kad-dht`. See https://github.com/libp2p/go-libp2p-kad-dht/releases/tag/v0.3.0.

PR #545, opened 2019-11-19
rpc: log errors when RPC method panics

fabioberger:
Until now, if an RPC request caused a panic, the panic was caught by the `go-ethereum` rpc library and a non-descript `method handler crashed` error message was returned to the caller. In order to log a helpful error message, including the stack trace of where the panic originated from, this PR adds panic recovery to all methods in RPCHandler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This changelog is a work in progress and may contain notes for versions which ha
 - Changed the default port for `RPC_ADDR` from a random available port to `60557`. _Some_ documentation already assumed `60557` was the default port. Now all documentation has been updated for consistency with this change. ([#542](https://github.com/0xProject/0x-mesh/pull/542)). 
 - Fixed a potential nil pointer exception in log hooks ([#543](https://github.com/0xProject/0x-mesh/pull/543)).
 - Fixed a bug where successful closes of an rpc subscription were being reported as errors ([#544](https://github.com/0xProject/0x-mesh/pull/544)).
+- We now log the error and stack trace if an RPC method panics. Before, these errors were swallowed by the panic recovery logic in `go-ethereum`'s `rpc` package. ([#545](https://github.com/0xProject/0x-mesh/pull/545))
 
 ## v6.0.1-beta
 
--- a/cmd/mesh/rpc_handler.go
+++ b/cmd/mesh/rpc_handler.go
@@ -8,7 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"runtime"
+	"runtime/debug"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


cmd/mesh/rpc_handler.go
```diff
@@ -86,8 +110,19 @@ func (handler *rpcHandler) AddOrders(signedOrdersRaw []*json.RawMessage, opts rp
...
-func (handler *rpcHandler) AddPeer(peerInfo peerstore.PeerInfo) error {
+func (handler *rpcHandler) AddPeer(peerInfo peerstore.PeerInfo) (errRes error) {
 	log.Debug("received AddPeer request via RPC")
+	// Catch panics, log stack trace and return RPC error message
+	defer func() {
+		if err := recover(); err != nil {
+			log.WithFields(log.Fields{
+				"error":      err,
+				"method":     "AddPeer",
+				"stackTrace": getStackTrace(),
```
albrow: I think you can just use [`debug.Stack`](https://golang.org/pkg/runtime/debug/#Stack) here:

```suggestion
				"stackTrace": string(debug.Stack()),
```

cmd/mesh/rpc_handler.go
```diff
@@ -86,8 +110,19 @@ func (handler *rpcHandler) AddOrders(signedOrdersRaw []*json.RawMessage, opts rp
 }
 
 // AddPeer is called when an RPC client calls AddPeer,
-func (handler *rpcHandler) AddPeer(peerInfo peerstore.PeerInfo) error {
+func (handler *rpcHandler) AddPeer(peerInfo peerstore.PeerInfo) (errRes error) {
```
albrow: Convention is to just call this `err`.

```suggestion
func (handler *rpcHandler) AddPeer(peerInfo peerstore.PeerInfo) (err error) {
```

cmd/mesh/rpc_handler.go
```diff
@@ -86,8 +110,19 @@ func (handler *rpcHandler) AddOrders(signedOrdersRaw []*json.RawMessage, opts rp
...
 	log.Debug("received AddPeer request via RPC")
+	// Catch panics, log stack trace and return RPC error message
+	defer func() {
+		if err := recover(); err != nil {
```
albrow: [`recover()` returns an `interface{}` and is not guaranteed to return an `error`](https://golang.org/pkg/builtin/#recover). Typically I would write this as:

```go
if r := recover(); r != nil {
    var ok bool
    var internalErr error
    if internalErr, ok = r.(error); !ok {
        // If r is not of type error, convert it.
        internalErr = fmt.Errorf("Recovered from non-error: (%T) %v", e, e)
    }
    // Log internalErr here...
}
```

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This changelog is a work in progress and may contain notes for versions which ha
 
 ## v6.0.2-beta
 
+### Features ✅
```
albrow: Hmm... I wonder if this should be considered a bug fix instead of a feature? It's not something that end users are ever going to "use" per se.

If you do want this to be a feature, then we should bump the upcoming release to `6.1.0-beta`.

fabioberger: Agreed.

Issue #470, opened 2019-10-24
blockwatch: Don't fast-sync if more than 128 blocks behind


fabioberger:
Currently, we always fast-sync when a Mesh node is behind the latest block. If it is far behind, it can take up quite some memory (see: #407). In addition, if we are more than 128 blocks behind, the resultant order revalidations would require a backing Ethereum node with `archive` mode enabled. Most do not have this.

Instead, if Mesh is more than 128 blocks behind, we should simply revalidate all orders once and re-set the latest block in the DB to the current head of the chain.

dekz:
~32 minutes for 128 blocks isn't a huge outage. Network blip, infura/alchemy down aren't rare.

If mesh blows past all of the blocks during that outage, what will a Market Maker see in terms of book keeping? Will they get a bunch of `EXPIRED` updated where some of them should have `FILLED` events on them? How does this effect a Market Makers book keeping and audit history?

For a "relayer" this is the right approach, unless they too were showing a users history.

Issue #407, opened 2019-09-12
fast-sync uses too much memory when trying to sync a large number of blocks


albrow:
I've observed an issue that comes up when starting Mesh for the first time after a while. One of our Mesh nodes was constantly restarting. The last message we see before restarting is "Some blocks have elapsed since last boot...". Here's a screenshot of the logs:

<img width="1337" alt="Screen Shot 2019-09-12 at 10 25 32 AM" src="https://user-images.githubusercontent.com/800857/64820936-633f8b00-d566-11e9-9dbf-6c02288b99f2.png">

I used `docker stats` to see that memory usage was steadily climbing until the container is eventually forcibly killed by Docker.

<img width="1142" alt="Screen Shot 2019-09-12 at 10 26 56 AM" src="https://user-images.githubusercontent.com/800857/64821027-a568cc80-d566-11e9-8623-6a17e97df2f1.png">

I believe this is happening because the "fast-sync" feature in Mesh keeps all event logs in memory while trying to catch up to the latest block. We can solve this issue by either (1) processing the logs in batches instead of trying to keep them all in memory or (2) placing a cap on the maximum number of blocks that Mesh will attempt to fast-sync. If too many blocks have passed, we would be better off starting from scratch and revalidating all existing orders.

The workaround for now is to manually delete the __db/__ folder, causing Mesh to start from scratch instead of trying to use the fast-sync feature.

PR #546, opened 2019-11-19
core,blockwatch: Limit fast-sync to last 128 blocks

fabioberger:
Fixes: https://github.com/0xProject/0x-mesh/issues/470 and https://github.com/0xProject/0x-mesh/issues/407

New logic is:
- If less than 128 blocks elapsed, fast-sync block events
- If more than 128 blocks, re-set stored blocks in DB and re-validate all stored orders.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -440,7 +440,7 @@ func (app *App) Start(ctx context.Context) error {
 		latestBlockNumber := latestBlock.Number
 
 		blocksElapsed = latestBlockNumber.Int64() - latestBlockProcessedNumber.Int64()
-		if blocksElapsed < maxBlocksStoredInNonArchiveNode {
+		if blocksElapsed > 0 && blocksElapsed < maxBlocksStoredInNonArchiveNode {
 			log.WithField("blocksElapsed", blocksElapsed).Info("Some blocks have elapsed since last boot. Backfilling block events (this can take a while)...")
 			// Note: this is a blocking call so we won't continue set up until its finished.
 			err := app.blockWatcher.BackfillEventsIfNeeded(innerCtx)
--- a/core/core.go
+++ b/core/core.go
@@ -19,7 +19,6 @@ import (
 	"github.com/0xProject/0x-mesh/ethereum/blockwatch"
 	"github.com/0xProject/0x-mesh/ethereum/dbstack"
 	"github.com/0xProject/0x-mesh/ethereum/ethrpcclient"
-	"github.com/0xProject/0x-mesh/ethereum/miniheader"
 	"github.com/0xProject/0x-mesh/ethereum/ratelimit"
 	"github.com/0xProject/0x-mesh/expirationwatch"
 	"github.com/0xProject/0x-mesh/keys"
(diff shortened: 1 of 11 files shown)


core/core.go
```diff
@@ -421,11 +426,32 @@ func (app *App) Start(ctx context.Context) error {
...
+	blocksElapsed := big.NewInt(0)
+	// If has processed blocks before
+	if latestBlockProcessed != nil {
+		latestBlockProcessedNumber := latestBlockProcessed.Number
+		latestBlock, err := app.blockWatcher.GetLatestBlock()
+		if err != nil {
+			return err
+		}
+		latestBlockNumber := latestBlock.Number
+		blocksElapsed = big.NewInt(0).Sub(latestBlockNumber, latestBlockProcessedNumber)
```
albrow: When profiling performance in the browser, methods in the `math/big` package show up often in the list of those responsible for high CPU usage. Generally, we need to go through and optimize our usage of `math/big` and only use it when necessary. This is a case where I don't think it's necessary. I understand that the Ethereum RPC interface gives us the current block number as a `big.Int` and that's fine. Can we still convert to `int64` as early as possible and then do the math on `int64` types?

core/core.go
```diff
@@ -436,6 +462,14 @@ func (app *App) Start(ctx context.Context) error {
 		blockWatcherErrChan <- app.blockWatcher.Watch(innerCtx)
 	}()
 
+	if blocksElapsed.Int64() >= maxBlocksStoredInNonArchiveNode {
```
albrow: This if statement `blocksElapsed.Int64() >= maxBlocksStoredInNonArchiveNode` effectively appears twice. Can we move this chunk of code to immediately after `app.db.ClearAllMiniHeaders()` so we don't have to check if we've exceeded `maxBlocksStoredInNonArchiveNode` twice?

core/core.go
fabioberger: Agreed. Made this change here and will keep this in mind in future work.

core/core.go
fabioberger: I initially had it above, but would rather we re-validate all orders _after_ the BlockWatcher has been started since it _is_ a blocking operation. While we're re-validating orders, further blocks events could be missed that impact the validity of orders.

core/core.go
```diff
@@ -421,11 +425,35 @@ func (app *App) Start(ctx context.Context) error {
...
+	if latestBlockProcessed != nil {
+		latestBlockProcessedNumber := latestBlockProcessed.Number
+		latestBlock, err := app.blockWatcher.GetLatestBlock()
+		if err != nil {
+			return err
+		}
+		latestBlockNumber := latestBlock.Number
+
+		blocksElapsed = latestBlockNumber.Int64() - latestBlockProcessedNumber.Int64()
+		if blocksElapsed < maxBlocksStoredInNonArchiveNode {
```
```suggestion
		if blocksElapsed > 0 && blocksElapsed < maxBlocksStoredInNonArchiveNode {
```

meshdb/meshdb.go
```diff
@@ -240,6 +240,20 @@ func (m *MeshDB) FindLatestMiniHeader() (*miniheader.MiniHeader, error) {
 	return miniHeaders[0], nil
 }
 
+// ClearAllMiniHeaders removes all stored MiniHeaders from the database.
+func (m *MeshDB) ClearAllMiniHeaders() error {
+	var storedHeaders []*miniheader.MiniHeader
+	if err := m.MiniHeaders.FindAll(&storedHeaders); err != nil {
```
albrow: This is really inefficient and could be improved with some changes to the `db` package. I'll make a separate issue for that.

For now, can you use a transaction here? Would ensure that:

1. We have consistency between the calls to `FindAll` and `Delete` and the state doesn't change in the middle of this function.
2. We perform all the deletes at once in a single atomic batch.

Issue #337, opened 2019-08-05
Additional custom order validation


fabioberger:
With the ability to subscribe to specific topics (#336) the question inevitably rises about how one can enforce that a specific topic is only used for sharing a subset of 0x orders. The solution to this is to introduce the ability for Mesh operators to add custom order validation to their node. 

E.g., if you are using your Mesh node to share only ETH/DAI orders, you could subscribe to a topic representing that trading pair and also add custom validation to your Mesh node that enforces that the maker/taker asset in all orders shared on that topic are ETH/DAI.

### Possible approaches

Several approaches were investigated for how to support custom validation:

1. Golang Plugins -- This is a native feature that allows a Go program to dynamically load another go program by path. Although this would have the highest performance characteristics, it currently only works on Linux/Unix and therefore probably isn't supported in WASM.
2. Hashicorps' [go-plugin](https://github.com/hashicorp/go-plugin) package uses gRPC and a spawned separate process to allow calling into a custom plugin from a Go program. This is less performant then (1) but still pretty good since the plugin is running on the same machine, however requiring spinning up a separate process and having a client/server gRPC interaction makes this unlikely to work in a browser, and probably overkill for our use-case.
3. Implement a JSON-RPC subscription `mesh_subscribe, ['orderValidation', 'TOPIC_ID']` which will batch-send orders broadcast on a particular topic to get validated by the client. The client will then return the validation results (array of accepted & rejected orderHashes) to the Mesh node. This interface can also be exposed as a JS callback in the browser. 

Although (3) is probably the least efficient approach of the 3, if we add support for IPC JSON-RPC connections, it becomes as efficient as (2) for non-WASM nodes. It is also guaranteed in WASM.

PR #563, opened 2019-11-28
WIP: Implement custom order filtering

### LINKED ISSUE #336: Add support for custom topics

albrow:
Fixes #336 and #337. Supersedes #349.

### Overview

This PR implements custom order filtering and custom topics. A custom filter may be passed into Mesh as a [JSON Schema](https://json-schema.org/) via the new `CUSTOM_ORDER_FILTER` environment variable. Messages that contain orders that don't match this schema will be dropped. As a limitation, filtering is only possible by looking at the static fields of an order. So for example, it is not possible to filter orders by doing an on-chain check or sending an HTTP request to a third-party API. We don't expect that this limitation is going to be a problem in practice and it comes with the huge benefit of enabling cross-topic forwarding in the future (more on that later).

This PR also changes the way that pubsub topics work in Mesh. Previously, there was only one topic and it was hard-coded based on the chain ID. Now, there is a one-to-one correspondence between custom filters and topics. You can convert from custom order schema to a topic and vice versa. This is a breaking change and the old topic names are not compatible with this PR. (However, with this change we can potentially implement cross-version forwarding for forward compatibility from now on).

### New order and message schemas.

All orders must match the following JSON Schema:

```json
{
	"id": "/rootOrder",
	"allOf": [{
		"$ref": "/customOrder"
	}, {
		"$ref": "/signedOrder"
	}]
}
```

- `/signedOrder` is the same JSON Schema as before and will match only valid 0x orders.
- `/customOrder` is the custom schema passed in through the `CUSTOM_ORDER_FILTER` environment variable.

Organizing the JSON Schema for orders like this means that `CUSTOM_ORDER_FILTER` can be relatively small. It doesn't need to contain all the required fields for a signed 0x order. It just needs to contain any _additional_ requirements on top of the default ones.

The new schema for order messages sent through pubsub looks like this:

```json
{
	"id": "/rootOrderMessage",
	"properties": {
		"messageType": {
			"type": "string",
			"pattern": "order"
		},
		"order": {
			"$ref": "/rootOrder"
		},
		"topics": {
			"type": "array",
			"minItems": 1,
			"items": {
				"type": "string"
			}
		}
	},
	"required": ["messageType", "order", "topics"]
}
```

Notably this adds the `topics` field which is a list of topics that the message was sent on. This is critical for cross-topic forwarding.

#### Example custom order schemas

The following `CUSTOM_ORDER_FILTER` doesn't add any additional requirements. All valid signed 0x orders will be accepted. This is the default value if no custom filter is passed in.

```json
{}
```

The following `CUSTOM_ORDER_FILTER` matches any valid signed 0x orders with a specific sender address:

```json
{
	"properties": {
		"senderAddress": {
			"pattern": "0x00000000000000000000000000000000ba5eba11",
			"type": "string"
		}
	}
}
```

This can easily be tweaked to filter orders by asset type, maker/taker address, or fee recipient.

### New topic strings

The new topic format looks like this:

```typescript
`/0x-orders/version/${versionNumber}/chain/${chainID}/schema/${encodedSchema}`
```

- `versionNumber` is the version for the topic string. This PR changes the default version from 1 to 3. (version 2 is already being used by the `0xV3` branch).
- `chainID` is the chain ID that Mesh is configured to use.
- `encodedSchema` is the base64-encoded custom order schema. This trick allows us to extract the custom order schema from the topic string.

#### Cross-topic forwarding

Part of the reason this PR is implemented in a particular way is to enable "cross-topic forwarding" in the future. What this means is that a Mesh node that is subscribed on the default topic (all 0x orders) can receive orders on any topic and forward them to nodes that are _only_ subscribed to that topic _without_ subscribing to or even knowing about the topic ahead of time. This PR does not implement cross-topic forwarding but it does add everything we need to do so in the future without breaking backwards-compatibility.

Here's a rough overview of how it will work:

1. Mesh nodes which are using custom filters will publish every order to two topics: (a) the topic corresponding to their custom filter and (b) the default topic. Because of how the JSON Schemas are implemented with `allOf`, we know that any order which is valid for a custom topic would also be considered valid on the default topic.
2. Mesh nodes with cross-topic forwarding enabled will check every message that they receive to see which topics they were originally published to. They can then decode the custom schema from the topic string and validate that the order does match the custom schema. If it matches, they can forward the message to the custom topic. Hopefully we can do this while retaining the original pubsub signature, but it's possible we need to re-sign the message before forwarding.

A similar approach can be used to forward messages between topic versions. In other words a message that was originally sent on topic version 3 could be modified and forwarded to topic version 4. This means we could potentially break backwards compatibility at the pubsub layer without completely fragmenting liquidity (although this approach would add additional latency).


p2p/validatorset/set.go
```diff
@@ -0,0 +1,58 @@
...
+			return false
+		default:
+		}
+
+		// Otherwise continue by running this validator.
+		isValid := validator.validator(ctx, sender, msg)
+		if !isValid {
+			// TODO(albrow): Change the verbosity of this log to Trace.
+			// TODO(albrow): Should we reduce a peer's score as a penalty for invalid
+			//               messages?
```
fabioberger: I'd say so. Why wouldn't we?

core/core.go
```diff
@@ -130,6 +130,13 @@ type Config struct {
 	// enforcing a limit on maximum expiration time for incoming orders and remove
 	// any orders with an expiration time too far in the future.
 	MaxOrdersInStorage int `envvar:"MAX_ORDERS_IN_STORAGE" default:"100000"`
+	// CustomOrderFilter is a JSON-schema which will be used for validating
+	// incoming orders. If provided, Mesh will only receive orders from other
+	// peers in the network with the same filter.
+	//
+	// TODO(albrow): Link to more documentation about JSON-schemas and how this
+	// filter works.
+	CustomOrderFilter string `envvar:"CUSTOM_ORDER_FILTER" default:""`
```
fabioberger: Can we make the default `"{}"` instead of an empty string? The former is actually what the default schema looks like and elucidates to the reader more clearly the form in which the schema is expected to be received (a stringified JSON object).

orderfilter/filter.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func loadExchangeAddress(loader *jsonschema.SchemaLoader, chainID int) error {
+	contractAddresses, err := ethereum.GetContractAddressesForChainID(chainID)
+	if err != nil {
+		return err
+	}
+	// Note that exchangeAddressSchema accepts both checksummed and
+	// non-checksummed (i.e. all lowercase) addresses.
+	exchangeAddressSchema := fmt.Sprintf(`{"oneOf":[{"type":"string","pattern":%q},{"type":"string","pattern":%q}]}`, contractAddresses.Exchange.Hex(), strings.ToLower(contractAddresses.Exchange.Hex()))
```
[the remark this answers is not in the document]
albrow: Yes.

orderfilter/filter.go
```diff
@@ -0,0 +1,202 @@
...
+// Dummy declaration to ensure that ValidatePubSubMessage matches the expected
+// signature for pubsub.Validator.
+var _ pubsub.Validator = (&Filter{}).ValidatePubSubMessage
+
+// ValidatePubSubMessage is an implementation of pubsub.Validator and will
+// return true if the contents of the message pass the message JSON Schema.
+func (f *Filter) ValidatePubSubMessage(ctx context.Context, sender peer.ID, msg *pubsub.Message) bool {
+	isValid, err := f.MatchOrderMessageJSON(msg.Data)
+	if err != nil {
+		log.WithError(err).Error("MatchOrderMessageJSON returned an error")
```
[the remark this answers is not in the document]
albrow: Yeah good point. I'll reduce the verbosity to `trace`.

core/core.go
albrow: Sure. Done.

p2p/validatorset/set.go
albrow: It's just a question of code complexity. Let me see if I can make it work.

ethereum/ratelimit/rate_limiter_test.go
```diff
@@ -19,9 +19,9 @@ var (
 	maxRequestsPer24HrsWithoutBuffer      = 301000
 	maxRequestsPer24hrs                   = maxRequestsPer24HrsWithoutBuffer - maxRequestsPer24HrsBuffer
 	maxRequestsPerSecond                  = 10.0
-	twentyFourHrs                         = (24 * 60 * 60 * 1000 * time.Millisecond)
+	twentyFourHrs                         = 24 * time.Hour
 	maxExpectedDelay                      = twentyFourHrs / time.Duration(maxRequestsPer24hrs)
-	minExpectedDelay                      = time.Duration(1000) / time.Duration(maxRequestsPerSecond) * time.Millisecond
+	minExpectedDelay                      = 1 * time.Second / time.Duration(maxRequestsPerSecond)
```
albrow: @fabioberger I changed some things here in order to get the tests to pass. Let me know if this new version looks right to you.

`time.Duration(1000) / time.Duration(maxRequestsPerSecond) * time.Millisecond` in particular looks a little off to me. `time.Duration(1000)` is 1,000 nanoseconds, but I don't think that's what you intended.

fabioberger: Thanks for re-writing this @albrow. The new version is much more readable and obviously correct.

albrow:
Superseded by https://github.com/0xProject/0x-mesh/pull/630.

albrow:
Please don't delete the `feature/custom-order-filters` branch for now.

Issue #561, opened 2019-11-27
blockwatch: Refactor polling loop to always fetch until latest


fabioberger:
Currently BlockWatcher always fetches the next known block. If for whatever reason it falls behind the latest block, it's polling interval will dictate how quickly it will catch up with the latest block.

It turns out the production of blocks, while having an average of one block produced per ~13 seconds, this happens far from clock-work fashion. In analyzing the last 20k blocks, I observed it could take as little as 1 second, or as long as 112sec to mine a block. 5% of blocks took less than 2sec to mine, 12% took less than 3sec and 17% less than 4sec. 

If the default block polling interval is set to 5sec, the current design can cause the Mesh node to fall several blocks behind if a few fast blocks are mined in succession. We'd have to wait `blocksPast * 5sec` time before we were caught up.

An alternative design is to always have the block poller fetch the latest block on an interval. If several blocks have passed between the last block seen and the latest block, it sends additional requests to backfill those missing. This approach guarentees that every 5seconds, we once again have caught up with the latest block.

This approach requires the fast-sync behavior we've already built, so it is now a feasible approach.

fabioberger:
Addressed in #564

PR #564, opened 2019-11-28
blockwatch: Sync to latest block available on each polling interval

fabioberger:
Fixes: https://github.com/0xProject/0x-mesh/issues/561

In addition, this PR also:
- Removes the `StartBlockDepth` config since with fast-sync, we always want to start the BlockWatcher at the latest block.
- Immediately fetches the first block when `blockWatcher.Watch()` is called instead of waiting for the first Ticker tick.
- Ensures that we only emit block events if our state of the chain has moved forward in terms of block number. If an error is thrown during a block-reorg for instance, all state changes are reverted to the last known good state. This reduces the amount of block event thrash and gives us this nice property that block events always leave the node at a higher block number which we can use to asynchronously update `fillableTakerAssetAmount`s in the DB in a subsequent PR.
- Persists changes to the stored block headers in the DB only once finalized. Updates are done atomically using transactions.


ethereum/dbstack/db_stack.go
```diff
@@ -1,91 +1,168 @@
...
-// DBStack allows performing basic stack operations on a stack of meshdb.MiniHeaders stored
-// in the DB backing our meshdb package.
-type DBStack struct {
-	// TODO(albrow): Use Transactions when db supports them instead of a mutex
-	// here. There are cases where we need to make sure no modifications are made
-	// to the database in between a read/write or read/delete.
-	mut    sync.Mutex
-	meshDB *meshdb.MeshDB
-	limit  int
+type updateType int
```
albrow: Can we de-dupe the definitions for `updateType` and `update`? Currently defined both here and in __ethereum/blockwatch/simple_stack.go__.

ethereum/dbstack/db_stack.go
```diff
@@ -1,91 +1,168 @@
...
+
+	txn := d.meshDB.MiniHeaders.OpenTransaction()
+	defer func() {
+		_ = txn.Discard()
+	}()
+
+	for _, u := range d.updates {
+		switch u.Type {
+		case pop:
+			latestMiniHeader, err := d.meshDB.FindLatestMiniHeader()
```
albrow: In the context of a `db.Transaction`, nothing is actually written to the database until `Commit` is called. That means that `FindLatestMiniHeader` always returns the same `MiniHeader`, whatever the latests is at the time the transaction was opened. I don't think this is what you intended.

If I'm right and this is a bug, we should also improve our test coverage so we can catch cases like this.

ethereum/dbstack/db_stack_test.go
```diff
@@ -54,3 +55,44 @@ func TestDBStackPushPeekPop(t *testing.T) {
 	require.NoError(t, err)
 	assert.Len(t, miniHeaders, expectedLen)
 }
+
+func TestDBStackReset(t *testing.T) {
+	meshDB, err := meshdb.New("/tmp/leveldb_testing/" + uuid.New().String())
+	require.NoError(t, err)
+	stack, err := New(meshDB, 10)
```
albrow: We should improve this test by doing more than one push/pop. We should also include a case where the same miniheader is pushed and popped more than once. We should also include a test where we call both `Checkpoint` and `Reset`.

ethereum/dbstack/db_stack.go
```diff
@@ -1,91 +1,168 @@
...
+		meshDB:      meshDB,
+		limit:       retentionLimit,
+		miniHeaders: miniHeaders,
+		updates:     []*update{},
+	}
+	return d, nil
+}
+
+// Peek returns the top of the stack
+func (d *DBStack) Peek() (*miniheader.MiniHeader, error) {
```
albrow: Since `DBStack` now also holds miniheaders in memory it seems like it is a lot more similar to `SimpleStack`. If I understand correctly, the only real changes are in the `Checkpoint` and `Reset` methods. Maybe we could reduce some duplicated code by having `DBStack` own a `SimpleStack` as a (possibly embedded) struct field?

ethereum/blockwatch/block_watcher.go
```diff
@@ -185,46 +194,82 @@ func (w *Watcher) GetAllRetainedBlocks() ([]*miniheader.MiniHeader, error) {
...
+
+	allEvents := []*Event{}
+	var syncErr error
+	for i := 0; i < numBlocksToFetch; i++ {
+		// Optimization: If numBlocksToFetch is 1, we already know what the nextHeader is, so avoid
+		// fetching it again. If there is more then 1 block to fetch, compute each from the last
+		// stored and fetch it
+		nextHeader := latestHeader
+		if numBlocksToFetch != 1 {
+			lastStoredHeader, syncErr := w.stack.Peek()
```
albrow: There's a subtle bug here. Because you used the `:=` operator, `syncErr` is redefined as a new variable within the scope of the for loop. Even though the two variables have the same name, changing the value of the one inside of the for loop will not change the one on the outside. See this playground example: https://play.golang.org/p/MsDG77VWgUr.

This is a common gotcha in Go and we should try to avoid declaring and redefining `syncErr` like this altogether if possible. It looks like the main reason you wrote the code this way is to clean up at the end of the for loop by either calling `Reset` or `Checkpoint` and potentially sending events through the `blockFeed`. I wonder if there is another way to structure this code? Maybe we could make use of `defer`?

The presence of this bug also makes me nervous about our test coverage. Can we improve our tests so that we would catch cases like this?

ethereum/blockwatch/block_watcher.go
```diff
@@ -185,46 +194,82 @@ func (w *Watcher) GetAllRetainedBlocks() ([]*miniheader.MiniHeader, error) {
...
-	return nil
+
+	return syncErr
+}
+
+func (w *Watcher) shouldRevertChanges(lastStoredHeader *miniheader.MiniHeader, events []*Event) bool {
+	if len(events) == 0 || lastStoredHeader == nil {
+		return false
+	}
+	// If we haven't progressed in terms of block number, revert back to previous "latest" block.
```
albrow: Can you elaborate on this? Shouldn't re-orgs only ever occur if one chain reached a higher block number than the one we were previously considering the "latest"?

ethereum/blockwatch/block_watcher.go
```diff
@@ -185,46 +194,82 @@ func (w *Watcher) GetAllRetainedBlocks() ([]*miniheader.MiniHeader, error) {
...
+	// This ensures block events always leave the node further ahead, preventing unnecessary thrash
+	// during block-reorgs (which tend to cluster)
+	newLatestHeader := events[len(events)-1].BlockHeader
+	return newLatestHeader.Number.Int64() <= lastStoredHeader.Number.Int64()
```
albrow: If we just need to do a comparison and nothing else, `Cmp` is probably faster here and uses less memory.

ethereum/blockwatch/block_watcher.go
```diff
@@ -185,46 +194,82 @@ func (w *Watcher) GetAllRetainedBlocks() ([]*miniheader.MiniHeader, error) {
...
+			return nil
+		}
+		numBlocksToFetch = int(latestHeader.Number.Int64() - lastStoredHeader.Number.Int64())
 	}
-	if err != nil {
-		return err
+
+	allEvents := []*Event{}
+	var syncErr error
+	for i := 0; i < numBlocksToFetch; i++ {
```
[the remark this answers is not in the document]
fabioberger: Added the following conditional:

```
if numBlocksToFetch >= constants.MaxBlocksStoredInNonArchiveNode {
		return fmt.Errorf(
			"cannot sync blocks more than %d blocks behind latest (%d). Your `BLOCK_POLLING_INTERVAL` is probably set too high",
			constants.MaxBlocksStoredInNonArchiveNode,
			numBlocksToFetch,
		)
	}
```

This should only happen if their `BLOCK_POLLING_INTERVAL` was manually set to a very infrequent interval.

ethereum/dbstack/db_stack.go
fabioberger: Good catch. Fixed and added regression test.

ethereum/dbstack/db_stack_test.go
```diff
@@ -54,3 +55,44 @@ func TestDBStackPushPeekPop(t *testing.T) {
 	require.NoError(t, err)
 	assert.Len(t, miniHeaders, expectedLen)
 }
+
+func TestDBStackReset(t *testing.T) {
```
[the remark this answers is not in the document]
fabioberger: Now that I've refactored DBStack to contain an instance of SimpleStack and all common logic is in one place, the DBStack tests now also test the SimpleStack.

ethereum/dbstack/db_stack.go
fabioberger: Ended up deduping by making DBStack instantiate a SimpleStack within it.

ethereum/blockwatch/block_watcher.go
fabioberger: Good catch. I've re-written this logic so that we use the regular `err` var when assigning, and only if an error occurs do we assign it to `syncErr` before `break`ing. 

The test coverage for BlockWatcher is at ~80% and almost all lines that aren't covered are error conditions. I'll work on bumping up coverage and will start with a regression test for the above.

ethereum/blockwatch/block_watcher.go
fabioberger: That is correct. However, syncing to the latest block will involve multiple ETH RPC calls and if any of them fail, we reach this part of the code, and so we check that if we haven't progressed higher before the failure happened, we revert the changes made.

E.g. 1, found 4 blocks are missing, fetched first block with logs, then second, on fetching the 3rd block we received a non-200 response. Since two blocks were added, we don't revert.

E.g. 2, found block re-org, removed a block, tried to fetch replacement block but got non-200 response, since we only removed a block at this point, we are not further ahead so revert the changes made.

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,167 @@
...
+type SimpleStack struct {
+	limit              int
+	miniHeaders        []*miniheader.MiniHeader
+	updates            []*Update
+	mu                 sync.Mutex
+	latestCheckpointID int
+}
+
+// NewSimpleStack instantiates a new SimpleStack
+func NewSimpleStack(retentionLimit int, miniHeaders []*miniheader.MiniHeader) *SimpleStack {
```
albrow: This should just be called `New` since it's already namespaced in the `simplestack` package.

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,167 @@
...
+	Type       updateType
+	MiniHeader *miniheader.MiniHeader
+}
+
+// SimpleStack is a simple in-memory stack used in tests
+type SimpleStack struct {
+	limit              int
+	miniHeaders        []*miniheader.MiniHeader
+	updates            []*Update
+	mu                 sync.Mutex
```
albrow: Can you make this an `RWMutex`?

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,167 @@
...
+			return fmt.Errorf("Unrecognized update type encountered: %d", u.Type)
+		}
+	}
+	s.updates = []*Update{}
+	return nil
+}
+
+// GetUpdates returns the updates applied since the last checkpoint
+func (s *SimpleStack) GetUpdates() []*Update {
+	return s.updates
```
albrow: To guarantee goroutine-safety you need to make a copy of `s.updates` here. Otherwise the the caller will have a reference to the underlying array that is not protected by a mutex.

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,167 @@
...
+	s.miniHeaders = s.miniHeaders[:len(s.miniHeaders)-1]
+	s.updates = append(s.updates, &Update{
+		Type:       Pop,
+		MiniHeader: top,
+	})
+	return top, nil
+}
+
+// Push adds a miniheader.MiniHeader to the stack
+func (s *SimpleStack) Push(miniHeader *miniheader.MiniHeader) error {
```
albrow: Is there a reason to have separate `Push` and `push` methods? I don't like that it is possible to call `push` without holding a lock on the mutex. The fact that it's unexported helps, but I would prefer if we just had one method that always uses a mutex. I also don't see any reason to split up some methods (e.g., `Push`/`push`, `Pop`/`pop`) but not others (e.g. `Checkpoint`, `Clear`).

ethereum/blockwatch/block_watcher.go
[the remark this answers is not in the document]
albrow: Or if they experience network issues or Infura/Alchemy/someone's own Ethereum node goes down temporarily. Shouldn't we just do the same thing here that we do for "fast sync"?

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,167 @@
+package simplestack
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/0xProject/0x-mesh/ethereum/miniheader"
+)
+
+type updateType int
```
albrow: As a general rule, if a variable is exported, the type of that variable should also be exported.

Otherwise, it can add annoying restrictions on what you can do as the caller/user of this package. For example you could not write a function outside of this package that accepts `updateType` as an argument (sometimes comes up in helper functions for tests). 

```suggestion
type UpdateType int
```

ethereum/blockwatch/block_watcher.go
```diff
@@ -185,46 +200,99 @@ func (w *Watcher) GetAllRetainedBlocks() ([]*miniheader.MiniHeader, error) {
...
+	if numBlocksToFetch >= constants.MaxBlocksStoredInNonArchiveNode {
+		return fmt.Errorf(
+			"cannot sync blocks more than or equal to %d blocks behind latest (%d). Your `BLOCK_POLLING_INTERVAL` is probably set too high",
+			constants.MaxBlocksStoredInNonArchiveNode,
+			numBlocksToFetch,
+		)
+	}
+
+	allEvents := []*Event{}
+	var syncErr error
```
albrow: I like this solution better and it definitely seems less error-prone. Can you add a brief comment explaining what `syncErr` is, when it should be set, and when it will be returned? It is a little unusual and I think worth explicitly noting.

ethereum/blockwatch/block_watcher.go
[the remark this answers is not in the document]
fabioberger: Went with returning an error and having Mesh crash, since in addition to re-setting the blocks stored in the DB, we'd also need OrderWatcher to re-validate all orders at the latest block. This will be done upon Mesh re-boot.

ethereum/blockwatch/block_watcher.go
```diff
@@ -185,46 +200,99 @@ func (w *Watcher) GetAllRetainedBlocks() ([]*miniheader.MiniHeader, error) {
...
+	}
+	if len(allEvents) == 0 {
+		return syncErr
+	}
+	if w.shouldRevertChanges(lastStoredHeader, allEvents) {
+		if err := w.stack.Reset(w.stackCheckpointID); err != nil {
+			return err
+		}
+	} else {
+		w.stackCheckpointID, err = w.stack.Checkpoint()
```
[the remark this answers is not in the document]
fabioberger: I see. Found a way to get rid of the instance variable. We now checkpoint right at the beginning of the `syncChain` method, and use that checkpointID when calling `Reset()`. The only awkward part is that we also checkpoint again at the end of the method if we decide to persist the changes in the in-memory stack to the DB.

The reason why `Checkpoint()` persists the in-memory stack changes to the DB is because we are supporting a generic `Stack` interface in BlockWatcher (could be in-memory only or in-memory & DB persisted), and `Checkpoint()` is the logical place to persist to the DB, making the changes "official". 

The call to `Checkpoint()` at the beginning of `syncChain` is therefore largely to avoid the footgun issue brought up. It will always just result in incrementing the checkpointID, since the other changes made by `Checkpoint` will have already been made by the previous run of `syncChain`.

ethereum/simplestack/simple_stack.go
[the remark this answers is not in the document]
fabioberger: Done. But what about the `*Update`'s it contains within it? Couldn't the caller still modify those as well?

ethereum/simplestack/simple_stack.go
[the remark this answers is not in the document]
fabioberger: `push` and `pop` are both called from within `Reset` which already holds a mutex and so we need a way to access the `pop` and `push` functionality without attempting to acquire the mutex we're already holding. I made these methods private since they are only meant to be used internally by other methods that have already acquired a mutex.

ethereum/simplestack/simple_stack.go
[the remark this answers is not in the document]
albrow: Good point. Maybe we should just make them not pointers? The structs aren't very big.

ethereum/simplestack/simple_stack.go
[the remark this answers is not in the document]
albrow: In that case can you add a comment to `pop` and `push` indicating that the caller should already hold a lock on the mutex?

ethereum/blockwatch/block_watcher.go
```diff
@@ -78,20 +88,15 @@ type Watcher struct {
 }
 
 // New creates a new Watcher instance.
-func New(config Config) (*Watcher, error) {
-	checkpointID, err := config.Stack.Checkpoint()
-	if err != nil {
-		return nil, err
-	}
+func New(config Config) *Watcher {
 	bs := &Watcher{
```
albrow: nit: This should probably just be called `watcher`. I'm not sure where the variable name `bs` came from but I'm guessing it made more sense at some point in the past.

ethereum/dbstack/db_stack.go
```diff
@@ -120,4 +120,4 @@ func (d *DBStack) Checkpoint() (int, error) {
 		return 0, err
 	}
 	return d.simpleStack.Checkpoint()
-}
+}
```
albrow: nit: Can you add a new line to the end of this file? Your editor should be configured to do this automatically. The new line was there before and it looks like you removed it.

ethereum/simplestack/simple_stack.go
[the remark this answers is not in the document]
fabioberger: Let's leave as is.

PR #578, opened 2019-12-05
POC: core, node, rpc: Add support for advertising and discovering RFQ quote providers

fabioberger:
This PR provides a proof-of-concept implementation of the functionality to advertise and discover RFQ quote providers over Mesh's DHT. The quote provider would run a Mesh node at the same IP address as their quote provider server endpoint, and periodically refresh their advertisement on the DHT. Any other Mesh user, can find the endpoints for quote providers via their own Mesh node.

Rendez-vous string format:

`/0x-rfq/version/${versionNumber}/chain/${chainID}/standard/${standard}/assetPair/${assetPair}`

Standards examples: `zaidan-v1.0`, `zaidan-v2.0`, `rfqLite-v1.0`

AssetPair examples: `weth/dai`, `0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2/0x6b175474e89094c44da98b954eedeac495271d0f`

### `mesh_advertiseAsQuoteProvider`

A quote provider can advertise themselves as providing quotes for a specific asset pair over an endpoint that adheres to a specific standard. The advertisement will be placed on the DHT for traders to find.

Request:
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "mesh_advertiseAsQuoteProvider",
    "params": {
        "standard": "zaidan-v1.0",
        "assetPair": "weth/dai",
        "ttl": "3hr"
    }
}
```

Response:
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "result": true,
}
```

### `mesh_getQuoteProviders`

A trader can request a list of quote provider IP addresses that offer quotes for a specific asset pair and that comply to a specific standard. They can then use a client designed for the specified standard to start requesting quotes from these endpoints.

Request:
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "mesh_getQuoteProviders",
    "params": {
        "standard": "zaidan-v1.0",
        "assetPair": "weth/dai"
    }
}
```

Response:
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "result": [
        "192.168.178.158",
        "167.172.201.142",
        "18.204.221.103",
        ...
    ],
}
```

### DIFF
--- a/cmd/mesh/rpc_handler.go
+++ b/cmd/mesh/rpc_handler.go
@@ -176,6 +176,61 @@ func (handler *rpcHandler) GetStats() (result *rpc.GetStatsResponse, err error) 
 	return getStatsResponse, nil
 }
 
+// AdvertiseAsQuoteProvider is called when an RPC client requests to advertise themselves as a quote provider
+func (handler *rpcHandler) AdvertiseAsQuoteProvider(standard, assetPair string, ttl time.Duration) (err error) {
+	log.Debug("received AdvertiseAsQuoteProvider request via RPC")
+	// Catch panics, log stack trace and return RPC error message
+	defer func() {
+		if r := recover(); r != nil {
+			internalErr, ok := r.(error)
+			if !ok {
+				// If r is not of type error, convert it.
+				internalErr = fmt.Errorf("Recovered from non-error: (%T) %v", r, r)
+			}
+			log.WithFields(log.Fields{
+				"error":      internalErr,
+				"method":     "AdvertiseAsQuoteProvider",
+				"stackTrace": string(debug.Stack()),
+			}).Error("RPC method handler crashed")
+			err = errors.New("method handler crashed in AdvertiseAsQuoteProvider RPC call (check logs for stack trace)")
+		}
+	}()
+	err = handler.app.AdvertiseAsQuoteProvider(standard, assetPair, ttl)
+	if err != nil {
+		log.WithField("error", err.Error()).Error("internal error in AdvertiseAsQuoteProvider RPC call")
+		return constants.ErrInternal
+	}
+	return nil
+}
+
+// GetQuoteProviders is called when an RPC client requests to retrieve all the RFQ quote providers advertised
+// on the DHT for a specific asset pair and adhering to a specific RFQ specification
+func (handler *rpcHandler) GetQuoteProviders(standard, assetPair string) (quoteProviders []string, err error) {
+	log.Debug("received GetQuoteProviders request via RPC")
+	// Catch panics, log stack trace and return RPC error message
+	defer func() {
+		if r := recover(); r != nil {
+			internalErr, ok := r.(error)
+			if !ok {
+				// If r is not of type error, convert it.
+				internalErr = fmt.Errorf("Recovered from non-error: (%T) %v", r, r)
+			}
+			log.WithFields(log.Fields{
+				"error":      internalErr,
+				"method":     "GetQuoteProviders",
+				"stackTrace": string(debug.Stack()),
+			}).Error("RPC method handler crashed")
+			err = errors.New("method handler crashed in GetQuoteProviders RPC call (check logs for stack trace)")
+		}
+	}()
+	quoteProviders, err = handler.app.GetQuoteProviders(standard, assetPair)
+	if err != nil {
+		log.WithField("error", err.Error()).Error("internal error in GetQuoteProviders RPC call")
+		return nil, constants.ErrInternal
+	}
+	return quoteProviders, nil
+}
+
 // SubscribeToOrders is called when an RPC client sends a `mesh_subscribe` request with the `orders` topic parameter
 func (handler *rpcHandler) SubscribeToOrders(ctx context.Context) (result *ethrpc.Subscription, err error) {
 	log.Debug("received order event subscription request via RPC")
--- a/core/core.go
+++ b/core/core.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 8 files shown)


hrharder:
This is excellent! Will be incredibly useful for fostering a marketplace of RFQ/Dealer systems and I'm glad you've figured out a way to tie it in with mesh and leverage the DHT.

I'm curious about the best way to represent the `assetPair`, specifically in the context of an RFQ standard like Zaidan (v2) where "arbitrary swap" markets are supported rather than just conventional bid/asks on a `BASE/QUOTE` pair. 

For example, in the [draft Zaidan JSONRPC spec](https://github.com/ParadigmFoundation/zaidan-dealer-specification#schema-market), we've decided to represent a "market" as a single `makerAsset` provided by the dealer, and multiple `takerAssets` that can be provided by traders when requesting a quote. The main reason being, if we stuck to a market defined as a single base/quote asset, we would need n^2 markets to indicate that we support "arbitrary swaps" where trade sizes can be denominated in the base or quote asset, rather than just the base. This is explored more [here.](https://github.com/ParadigmFoundation/zaidan-dealer-specification#quotes) We also chose this route to allow "prices" (which depend on the notion of a quote asset) to be defined at higher levels in terms of either asset, depending on the use case and trade scenario.

The way the new endpoint and Rendezvous string is specified would require a dealer supporting arbitrary swaps between 50 different assets to advertise ~2500 different base/quote pairs, which  I'm not sure is ideal.

Of course, this may just be something unavoidable. I would agree that a `base/quote` representation for a supported market makes the most sense for most people, and is certainly the most familiar, but I thought this point may be worth mentioning.

fabioberger:
Thanks a lot for the feedback @hrharder. I cannot think of a way to supporting "arbitrary swaps" in the DHT specification. We have two potential options if we want to support "arbitrary swaps":

1. Remove the BASE/QUOTE component out of the rendez-vous string. Each RFQ specification must then provide an endpoint where the supported markets can be discovered (similar to the [dealer_getMarkets](https://github.com/ParadigmFoundation/zaidan-dealer-specification#method-dealer_getmarkets) endpoint in Zaidan). Clients would then ping this endpoint for each quote provider discovered and cache the markets it supports client-side. This is definitively less efficient but removes the need for quote providers to advertise themselves more than once on the DHT.

2. The rendez-vous string only includes the `makerAsset` the quote provider is willing to supply. This reduces the number of advertisements back to down to 50. The client must then still query the market endpoint to see if the quote provider is willing to accept the desired `takerAsset`. There will be fewer false-positives when querying quote providers then solution (1).

Any thoughts on either of the above?

hrharder:
@fabioberger I like both of these. I'll try to think through the trade-offs for each, but I expect you may be better equipped to guide a final decision given some of the intricacies w/r/t to the DHT, and the expectations of existing market makers (specifically around how desirable it is to even support the discovery of "arbitrary swap" quote providers).

My only hesitation with this approach in (1) is, as you say, the loss of efficiency for traders (meaning consumers of provided quotes) who are looking for a specific trading pair. Instead of a single query to the Mesh `mesh_getQuoteProviders` endpoint with the pair they want to trade on, they'd have to collect a list of quote providers, then query each one to find the pair they want, potentially having to use multiple different request formats if there are multiple RFQ standards being offered by quote providers. I'm unsure if this awkwardness is acceptable -- I'll look to your input here (see questions below).

The approach in (2), as you again point out, reduces this complexity perhaps slightly by allowing traders to filter by the maker asset. My hesitation here is that the relationship between a `base/quote` market and a quote provider serving a specific maker asset is not always incredibly apparent. And — if traders still have to go out and ping the matching quote providers to see if they support the right taker asset, is this that much better than (1)?

I think having a sense of your answers to the following questions would help me here:

1. Do we even want to allow discovery of "arbitrary swap" quote providers, or does it just add too much complexity in this avenue? We could instead decide that a `base/quote` format is more desirable, and then encourage quote providers to only advertise their most popular pairs, leaving discovery of an expanded asset swap universe to traders querying individual quote providers at their discretion.
1. How concerned should we be about "DHT bloat," if at all? If the answer is "very", then perhaps your suggestion above as (1) is best regardless of decision around the desire to support `base/quote` vs "arbitrary swap", as it would mean only one entry per quote provider, no matter how many assets/markets they support. If "DHT bloat" (likely not a good term here) is not a concern _at all_ then perhaps the `base/quote` format is OK, and quote providers just advertise ALL markets they support, even if that means ~2500 entries per provider. If the concern re. "bloat" falls somewhere between "very" and "not at all" then I would say your option (2) is best, as it still provides some query-ability to end users without requiring thousands of entries per provider.

Apologies for the lengthy response, and let me know if this isn't clear.

hrharder:
@fabioberger 

**TL;DR for the above:** If we're okay with tens-to-hundreds of thousands of DHT entries for this feature, perhaps the `base/quote` format is ideal, as it provides the most familiar (and best, IMO) UX for end users, esp. due to the ability to query for providers that serve the market's you're interested in. Not to mention, we don't know if "arbitrary swaps" is something traders or quote providers will even want to support. If that level of utilization of the DHT for this feature is unacceptable, perhaps one entry per quote provider is the only option, even if that means sacrificing some query-ability. And finally, even though it seems like a nice compromise between the above options, I question how valuable it is to have a rendezvous-string with only the maker asset.

hrharder:
I may have been overthinking this... A potentially elegant solution for `assetPair` may be `[makerAsset,]/[takerAsset,]`. Is there a reason this wouldn't work?

For example, `[WETH,DAI,ZRX]/[WETH,DAI,ZRX]` would denote arbitrary swaps between WETH, DAI, and ZRX. This means you could have down to 1 rendezvous-string for a dealer that supported arbitrary swaps between `n` assets. Of course, if someone wanted a bit more granularity, it would become `>1`, but certainly less than `n (n-1)`. An unfortunate side-effect is the `WETH/WETH` market, so during decoding we'd just have to enforce you cant have a market between two identical assets.

This likely should be by address in practice to reduce ambiguity, but still seems a lot shorter, and more decodable, than anything else. Thoughts?

albrow:
There are a lot of issues with adding entries to the DHT and we'd need a lot more time to explore reputation systems and other mechanisms to solve these issues. Our current plan is not to add this to Mesh right now.

We might reconsider in the future, but I don't think we should leave this PR open until then. It's already pretty far behind the `development` branch. We should open a fresh PR in the future if we decide to go this route.

PR #580, opened 2019-12-10
Go Client RPC Endpoint Tests

jalextowle:
## Description
This PR augments the existing tests for the Golang 0x-mesh RPC client. Originally, these tests were contained in the `rpc/client_server_test.go` file. The main focus of this PR was to make it so that these tests would target an actual mesh node. This entailed refactoring the `integration-tests` directory so that the tools that it provides were in an isolated `utils.go` file and adding the `client_server_test.go` file to the `integration-tests` directory. 

The previous test infrastructure created dummy rpc handlers because it allows the tester to have a large amount of control over the execution of the rpc server. Unfortunately, this approach also increases the overhead of maintaining the tests, which was the motivation for this PR implementing integration tests. The integration tests have significantly less control over the rpc server's interactions, which ultimately causes the new tests to be more complex. This increased complexity motivated the generalization of several of the utilities in the `utils.go` file.


integration-tests/constants.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/0xProject/0x-mesh/constants"
+)
+
+const (
+	ethereumRPCURL  = "http://localhost:8545"
+	ethereumChainID = 1337
+
+	rpcPort        = 60501
+	protocolString = "/meshsub/1.0.0"
```
albrow: This isn't part of the public interface and I don't think there's any real reason to include it in the tests. The tests will already fail here and inside the `p2p` package if for some reason peers are not using the same protocol ID for pubsub.

integration-tests/constants.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// Various config options/information for the standalone node. Like the
+	// bootstrap node, we know the private key/peer ID ahead of time.
+	standalonePeerID      = "16Uiu2HAmM9j68mgGGSFkXsuzbGJA8ezVHtQ2H9y6mRJAPhx6xtj9"
+	standaloneDataDir     = "./data/standalone-"
+	standaloneRPCEndpoint = "ws://localhost:"
+	standaloneRPCAddr     = "localhost:"
+)
+
+var makerAddress = constants.GanacheAccount1
```
albrow: These could just be combined into a single `var` group:

```go
var (
    makerAddress = constants.GanacheAccount1
    takerAddress = constants.GanacheAccount2
    // ...
)
```

integration-tests/constants.go
```diff
@@ -0,0 +1,36 @@
...
+	// peer ID ahead of time.
+	bootstrapAddr    = "/ip4/127.0.0.1/tcp/60500/ws"
+	bootstrapPeerID  = "16Uiu2HAmGd949LwaV4KNvK2WDSiMVy7xEmW983VH75CMmefmMpP7"
+	bootstrapList    = "/ip4/127.0.0.1/tcp/60500/ws/ipfs/16Uiu2HAmGd949LwaV4KNvK2WDSiMVy7xEmW983VH75CMmefmMpP7"
+	bootstrapDataDir = "./data/bootstrap-0"
+
+	// Various config options/information for the standalone node. Like the
+	// bootstrap node, we know the private key/peer ID ahead of time.
+	standalonePeerID      = "16Uiu2HAmM9j68mgGGSFkXsuzbGJA8ezVHtQ2H9y6mRJAPhx6xtj9"
+	standaloneDataDir     = "./data/standalone-"
```
albrow: Niether `standaloneDataDir`, `standaloneRPCEndpoint`, or `standaloneRPCAddr` are complete or correspond to a specific standalone node. Can we call these something like `standaloneDataDirPrefix`?

integration-tests/constants.go
```diff
@@ -0,0 +1,36 @@
...
+	// for the bootstrap node is checked in to version control so we know it's
+	// peer ID ahead of time.
+	bootstrapAddr    = "/ip4/127.0.0.1/tcp/60500/ws"
+	bootstrapPeerID  = "16Uiu2HAmGd949LwaV4KNvK2WDSiMVy7xEmW983VH75CMmefmMpP7"
+	bootstrapList    = "/ip4/127.0.0.1/tcp/60500/ws/ipfs/16Uiu2HAmGd949LwaV4KNvK2WDSiMVy7xEmW983VH75CMmefmMpP7"
+	bootstrapDataDir = "./data/bootstrap-0"
+
+	// Various config options/information for the standalone node. Like the
+	// bootstrap node, we know the private key/peer ID ahead of time.
+	standalonePeerID      = "16Uiu2HAmM9j68mgGGSFkXsuzbGJA8ezVHtQ2H9y6mRJAPhx6xtj9"
```
albrow: Which standalone node does this peer ID belong to? Can you clarify in the comments? It looks like the integration tests spin up at least 4 or 5 nodes.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+	ethClient := ethclient.NewClient(ethRPCClient)
+	signedTestOrder := scenario.CreateZRXForWETHSignedTestOrder(t, ethClient, makerAddress, takerAddress, wethAmount, zrxAmount)
+	expectedOrderHash, err := signedTestOrder.ComputeOrderHash()
+	require.NoError(t, err, "could not compute order hash for standalone order")
+
+	expectedFillableTakerAssetAmount := signedTestOrder.TakerAssetAmount
+
+	expectedPage := 0
+	expectedPerPage := 5
+	expectedSnapshotID := ""
```
albrow: Looks like none of `expectedPage`, `expectedPerPage`, or `expectedSnapshotID` are actually used for any assertions. We should improve this test by adding more than one order and making sure that the RPC endpoint handles pagination correctly.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+
+	// Set up the dummy handler with an addOrdersHandler
```
albrow: The code doesn't match the comment. "dummy handler" was only used in the old RPC tests in __rpc/client_server_test.go__, which was removed in this PR. This also applies to a few other places.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+
+	go func() {
+		defer wg.Done()
+		startStandaloneNode(t, ctx, count, logMessages)
+	}()
+
+	log, err := waitForLogSubstring(ctx, logMessages, "started RPC server")
+	require.NoError(t, err, "RPC server didn't start")
+	var jsonLog map[string]interface{}
+	err = json.Unmarshal([]byte(log), &jsonLog)
```
albrow: You are committing the private key for each standalone node in this PR. If we're doing that, then we don't need to extract the peer ID from the logs. We should either:

1. Don't commit the private key for each standalone node in this PR.
2. Extract the expected peer ID directly from the private key instead of reading the logs.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+	expectedGetStatsResponse := &rpc.GetStatsResponse{
+		Version:              "development",
+		PubSubTopic:          "/0x-orders/network/1337/version/1",
+		Rendezvous:           "/0x-mesh/network/1337/version/1",
+		PeerID:               jsonLog["myPeerID"].(string),
+		EthereumChainID:      1337,
+		LatestBlock:          rpc.LatestBlock{},
+		NumOrders:            0,
+		NumPeers:             0,
+		MaxExpirationTime:    "115792089237316195423570985008687907853269984665640564039457584007913129639935",
```
albrow:
```suggestion
		MaxExpirationTime:  constants.UnlimitedExpirationTime.String(),
```

nit: This is a little easier to understand IMO.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+	client, err := rpc.NewClient(standaloneRPCEndpoint + strconv.Itoa(rpcPort+count))
+	require.NoError(t, err)
+
+	getStatsResponse, err := client.GetStats()
+
+	// HACK(jalextowle): Zeroing the Number and Hash fields of LatestBlock
+	//                   allows us to test more of the "GetStats" response
+	//                   without being too restrictive about the blockchain
+	//                   that is being used.
+	getStatsResponse.LatestBlock = rpc.LatestBlock{}
```
albrow: We should still test that `LatestBlock` is included in the RPC response and is not zero instead of just ignoring it.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+		startStandaloneNode(t, ctx, count, logMessages)
+	}()
+
+	_, err := waitForLogSubstring(ctx, logMessages, "started RPC server")
+	require.NoError(t, err, "RPC server didn't start")
+
+	client, err := rpc.NewClient(standaloneRPCEndpoint + strconv.Itoa(rpcPort+count))
+	require.NoError(t, err)
+
+	orderEventChan := make(chan []*zeroex.OrderEvent)
```
albrow: We should use a buffered channel for `orderEventChan`. As the documentation explains:

> Client buffers up to 8000 notifications before considering the subscriber dead. The subscription Err channel will receive ErrSubscriptionQueueOverflow. Use a sufficiently large buffer on the channel or ensure that the channel usually has at least one reader to prevent this issue.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+	expectedFillableTakerAssetAmount := signedTestOrder.TakerAssetAmount
+	signedTestOrders := []*zeroex.SignedOrder{signedTestOrder}
+	_, err = client.AddOrders(signedTestOrders)
+	require.NoError(t, err)
+
+	// Reset the hash so that the order events can be compared without needing to hash
+	// the response order.
+	signedTestOrder.ResetHash()
+
+	orderEvent := <-orderEventChan
```
albrow: Can we use a timeout here via a `select` statement? Otherwise if something goes wrong this test will hang until the test run times out (currently set to 30s).

You can look at [another part of the code where we do this](https://github.com/0xProject/0x-mesh/blob/017d42d5a46e7376a7e3bb8879a59b095ca8772c/ethereum/blockwatch/block_watcher_test.go#L60) for an example.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+	assert.Equal(t, signedTestOrder, orderInfo.SignedOrder, "signedOrder did not match")
+	assert.Equal(t, expectedFillableTakerAssetAmount, orderInfo.FillableTakerAssetAmount, "fillableTakerAssetAmount did not match")
+
+	// Cancel the context and wait for all outstanding goroutines to finish.
+	cancel()
+	wg.Wait()
+}
+
+func TestAddPeer(t *testing.T) {
+	t.Skip("The AddPeer test is currently skipped because of nondeterministic behavior that causes it to intermittently fail")
```
albrow: Just a heads up. I'm not reviewing this test since it's skipped and not a high priority.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,465 @@
...
+	client, err := rpc.NewClient(standaloneRPCEndpoint + strconv.Itoa(rpcPort+count))
+	require.NoError(t, err)
+
+	heartbeatChan := make(chan string)
+	clientSubscription, err := client.SubscribeToHeartbeat(ctx, heartbeatChan)
+	defer clientSubscription.Unsubscribe()
+	require.NoError(t, err)
+	assert.NotNil(t, clientSubscription, "clientSubscription not nil")
+
+	heartbeat := <-heartbeatChan
```
albrow: Can we also add a timeout here with `select`?

integration-tests/utils.go
```diff
@@ -509,3 +385,9 @@ func setupSubTest(t *testing.T) func(t *testing.T) {
 		blockchainLifecycle.Revert(t)
 	}
 }
+
+// FIXME(jalextowle): This was taken from "ethereum/ratelimit." Should it be put elsewhere?
+func getUTCMidnightOfDate(date time.Time) time.Time {
```
albrow: Let's just export this in the `ethereum/ratelimit` package for now.

integration-tests/browser_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/0xProject/0x-mesh/zeroex"
+	"github.com/chromedp/chromedp"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBrowserIntegration(t *testing.T) {
+	if !integrationTestsEnabled {
+		t.Skip("Integration tests are disabled. You can enable them with the --integration flag")
```
albrow: The whole point of this `--integration` flag was that, prior to this PR, the integration tests had some additional requirements and took a lot longer to run. We didn't want to include them in `make test-all` or `make test-go` because of this.

However, in this PR there are now multiple "integration tests" that don't involve the browser and don't have any special requirements. Can we rename this flag to something like `--enable-browser-integration-tests`? You will also need to update the Makefile accordingly.

integration-tests/utils.go
```diff
@@ -22,48 +17,20 @@ import (
...
-var eighteenDecimalsInBaseUnits = new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
-var wethAmount = new(big.Int).Mul(big.NewInt(50), eighteenDecimalsInBaseUnits)
-var zrxAmount = new(big.Int).Mul(big.NewInt(100), eighteenDecimalsInBaseUnits)
-
 // Since the tests take so long, we don't want them to run as part of the normal
 // testing process. They will only be run if the "--integration" flag is used.
 var integrationTestsEnabled bool
 
 func init() {
+	var zero int32 = 0
```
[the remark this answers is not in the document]
jalextowle: Okay, this makes sense. I did it the other way so that `nodeCount` could actually be a pointer, but with this context, this seems better.

integration-tests/rpc_integration_test.go
jalextowle: I added a `context.WithTimeout` to this test. This should kill the test after 10 seconds. I'm going to go ahead and also add the `select`, but this seems redundant. What do you think?

integration-tests/browser_integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	go func() {
+		defer wg.Done()
+		startBootstrapNode(t, ctx)
+	}()
+
+	// standaloneLogMessages is a channel through which log messages from the
+	// standalone node will be sent. We use a large buffer so it doesn't cause
+	// goroutines to block.
+	standaloneLogMessages := make(chan string, 1024)
+	// count := int(atomic.AddInt32(&nodeCount, 1))
```
fabioberger: Is this commented out code temporary?

integration-tests/browser_integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+		require.NoError(t, err, "RPC server didn't start")
+		rpcClient, err := rpc.NewClient(standaloneRPCEndpointPrefix + strconv.Itoa(rpcPort+count))
+		require.NoError(t, err)
+		results, err := rpcClient.AddOrders([]*zeroex.SignedOrder{standaloneOrder})
+		require.NoError(t, err)
+		assert.Len(t, results.Accepted, 1, "Expected 1 order to be accepted over RPC")
+		assert.Len(t, results.Rejected, 0, "Expected 0 orders to be rejected over RPC")
+	}()
+
+	// Start a sinple HTTP server to serve the web page for the browser node.
```
fabioberger:
```suggestion
	// Start a simple HTTP server to serve the web page for the browser node.
```

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,416 @@
...
+
+	// Create a new valid order.
+	ethClient := ethclient.NewClient(ethRPCClient)
+	signedTestOrder := scenario.CreateZRXForWETHSignedTestOrder(t, ethClient, makerAddress, takerAddress, wethAmount, zrxAmount)
+
+	// Send the "AddOrders" request to the rpc server.
+	validationResponse, err := client.AddOrders([]*zeroex.SignedOrder{signedTestOrder})
+	require.NoError(t, err)
+
+	// Ensure that the validation validation results contain only the order that was
```
fabioberger:
```suggestion
	// Ensure that the validation results contain only the order that was
```

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,416 @@
...
+
+	// Send the "SubscribeToHeartbeat" request through the rpc client and assert
+	// that the subscription is not nil.
+	heartbeatChan := make(chan string)
+	clientSubscription, err := client.SubscribeToHeartbeat(ctx, heartbeatChan)
+	defer clientSubscription.Unsubscribe()
+	require.NoError(t, err)
+	assert.NotNil(t, clientSubscription, "clientSubscription not nil")
+
+	// Ensure that ta valid heartbeat was received
```
fabioberger:
```suggestion
	// Ensure that a valid heartbeat was received
```

integration-tests/browser_integration_test.go
jalextowle: Yes, I'm currently fixing some bugs in this PR.

cmd/mesh/rpc_handler.go
```diff
@@ -86,7 +86,7 @@ func (handler *rpcHandler) GetOrders(page, perPage int, snapshotID string) (resu
 			return nil, err
 		}
 		// We don't want to leak internal error details to the RPC client.
-		log.WithField("error", err.Error()).Error("internal error in AddOrders RPC call")
+		log.WithField("error", err.Error()).Error("internal error in GetOrders RPC call")
```
fabioberger: Nice catch! 🙏

Makefile
```diff
@@ -61,7 +61,7 @@ test-go-serial:
 
 .PHONY: test-integration
 test-integration:
-	go test ./integration-tests -timeout 185s --integration
+	go test ./integration-tests -timeout 185s --enable-browser-integration-tests
```
albrow: Technically this means we are running the non-browser integration tests twice. Once in `test-go` and again here. How about we change this Make task so that it _only_ runs the browser tests?

```
.PHONY: test-browser-integration
test-browser-integration:
	go test ./integration-tests -timeout 185s --enable-browser-integration-tests -run BrowserIntegration
```

Then we also need to update __.circleci/config.yml__.

integration-tests/browser_integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/0xProject/0x-mesh/zeroex"
+	"github.com/chromedp/chromedp"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBrowserIntegration(t *testing.T) {
+	if !browserIntegrationTestsEnabled {
+		t.Skip("Integration tests are disabled. You can enable them with the --enable-browser-integration-tests flag")
```
albrow:
```suggestion
		t.Skip("Browser integration tests are disabled. You can enable them with the --enable-browser-integration-tests flag")
```

integration-tests/constants.go
```diff
@@ -0,0 +1,35 @@
...
+	// Various config options/information for the bootstrap node. The private key
+	// for the bootstrap node is checked in to version control so we know it's
+	// peer ID ahead of time.
+	bootstrapAddr    = "/ip4/127.0.0.1/tcp/60500/ws"
+	bootstrapPeerID  = "16Uiu2HAmGd949LwaV4KNvK2WDSiMVy7xEmW983VH75CMmefmMpP7"
+	bootstrapList    = "/ip4/127.0.0.1/tcp/60500/ws/ipfs/16Uiu2HAmGd949LwaV4KNvK2WDSiMVy7xEmW983VH75CMmefmMpP7"
+	bootstrapDataDir = "./data/bootstrap-0"
+
+	// Various config options/information for the standalone node. Like the
+	// bootstrap node, we know the private key/peer ID ahead of time.
```
albrow: This comment doesn't match the underlying code and also doesn't hold true anymore. We don't always know the peer ID ahead of time for all the standalone nodes that are started in the integration tests.

integration-tests/rpc_integration_test.go
```diff
@@ -0,0 +1,421 @@
...
+	// Make a new "GetOrders" request with different pagination parameters.
+	expectedPage = 1
+	expectedPerPage = 5
+	getOrdersResponse, err := client.GetOrders(expectedPage, expectedPerPage, expectedSnapshotID)
+	require.NoError(t, err)
+	assert.Len(t, getOrdersResponse.OrdersInfos, expectedPerPage)
+
+	// Ensure that the getOrdersResponse has the correct pagination
+	for i := range getOrdersResponse.OrdersInfos {
+		assert.Equal(t, initialGetOrdersResponse.OrdersInfos[i+5*expectedPage], getOrdersResponse.OrdersInfos[i], "Incorrect order of second getOrdersResponse")
```
albrow: We don't really guarantee that the order of orders is the same between requests, it's just a side-effect of how orders are stored in the current database implementation. We shouldn't rely on this in our tests.

Instead of this, what do you think of sending two or three requests with pagination and then testing that _all_ the orders where returned (in any order)?

jalextowle: That sounds like a good way to test for pagination. Is there a place that we document this guarantee?

integration-tests/utils.go
```diff
@@ -420,6 +299,27 @@ func waitForReceivedOrderLog(ctx context.Context, logMessages <-chan string, exp
 	})
 }
 
+// Ensure that the signed orders
```
albrow: Can you finish this comment?

PR #584, opened 2019-12-12
Add new ENABLE_ETHEREUM_RPC_RATE_LIMITING environment variable

albrow:
This new environment variable will allow users to completely disable Ethereum RPC rate limiting. Disabling the feature does come with some risks and can lead to abnormally high bills for services like Infura and Alchemy depending on network conditions. Still, some users, and in particular market makers, might prefer higher costs to the alternative of not receiving order updates quickly.

### DIFF
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -97,6 +97,14 @@ export interface Config {
     // 30 rps for Infura's free tier, and can be increased to 100 rpc for pro
     // users, and potentially higher on alternative infrastructure.
     ethereumRPCMaxRequestsPerSecond?: number;
+    // Determines whether or not Mesh should limit the number of Ethereum RPC
+    // requests it sends. It defaults to true. Disabling Ethereum RPC rate
+    // limiting can reduce latency for receiving order events in some network
+    // conditions, but can also potentially lead to higher costs or other rate
+    // limiting issues outside of Mesh, depending on your Ethereum RPC provider.
+    // If set to false, ethereumRPCMaxRequestsPer24HrUTC and
+    // ethereumRPCMaxRequestsPerSecond will have no effect.
+    enableEthereumRPCRateLimiting?: boolean;
     // A set of custom addresses to use for the configured network ID. The
     // contract addresses for most common networks are already included by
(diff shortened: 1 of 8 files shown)


core/core.go
```diff
@@ -217,10 +225,16 @@ func New(config Config) (*App, error) {
 	}
 
 	// Initialize ETH JSON-RPC RateLimiter
-	clock := clock.New()
-	ethRPCRateLimiter, err := ratelimit.New(config.EthereumRPCMaxRequestsPer24HrUTC, config.EthereumRPCMaxRequestsPerSecond, meshDB, clock)
-	if err != nil {
-		return nil, err
+	var ethRPCRateLimiter ratelimit.RateLimiter
+	if config.EnableEthereumRPCRateLimiting == false {
+		ethRPCRateLimiter = ratelimit.NewUnlimited()
```
albrow: @fabioberger I opted to re-use the existing `fakeLimiter` type and just renamed the constructor from `NewFakeLimiter` to `NewUnlimited`. This way we can still count and log the number of RPC requests that were allowed. It also required the fewest code changes. Let me know if there would be any issues with this approach.

browser/ts/index.ts
```diff
@@ -97,6 +97,14 @@ export interface Config {
...
     // users, and potentially higher on alternative infrastructure.
     ethereumRPCMaxRequestsPerSecond?: number;
+    // Determines whether or not Mesh should limit the number of Ethereum RPC
+    // requests it sends. It defaults to true. Disabling Ethereum RPC rate
+    // limiting can reduce latency for receiving order events in some network
+    // conditions, but can also potentially lead to higher costs or other rate
+    // limiting issues outside of Mesh, depending on your Ethereum RPC provider.
+    // If set to false, ethereumRPCMaxRequestsPer24HrUTC and
+    // ethereumRPCMaxRequestsPerSecond will have no effect.
+    enableEthereumRPCRateLimiting?: boolean;
```
fabioberger: Can we move this config _above_ the 24hr and perSecond configs? Here and in the other files too.

albrow: 👍Good idea.

Issue #585, opened 2019-12-12
Rate limiter accounting bug


fabioberger:
We noticed two Mesh nodes that were experiencing the following error right at midnight UTC:

```
rate: Wait(n=1) would exceed context deadline
```

The error originates from the underlying `golang.org/x/time/rate` dependency we use in our ETH RPC rate limiting logic. This specific error message occurs if the limiter determines that the timeout on the supplied context to the `Wait()` method will timeout _before_ the request can be granted. We set the context timeout to our calls to `Wait()` to 30sec or 1min depending on the request type.

When someone calls `WaitN(ctx, n)` with a value `n` that is larger then all the tokens that have accrued in the bucket, the call will block until enough time has elapsed for n tokens to be available. In addition to blocking here, it will already adjust it's internal accounting, such that there is now a negative number of tokens available for other callers to `WaitN()/Wait()`. When this token number is negative, another call to `Wait()` will trigger the computation to check if the token number will become positive _before_ the supplied context is cancelled. 

When UTC midnight comes around, our rate limiter wishes to re-set all tokens that have accrued to the bucket. We currently do this by keeping track of the number of tokens granted in the last 24hr UTC window and subtracting this from the amount we know have accrued to get the amount still remaining in the bucket. We then call `WaitN()` with this number. _If_ our accounting is incorrect, we might try to deduct _more_ tokens then are currently in the bucket, causing the rate package's internal accounting to go into negative territory and cause further requests for tokens to get rejected because their 30sec/1min timeout would get hit before the tokens are available.

Our hunch is therefore that our internal accounting is off and we should try avoiding the need for any accounting by simply re-instantiating the `limiter` instance.

PR #587, opened 2019-12-12
WIP: Completely reset 24 hour rate limiter every midnight UTC

albrow:
This is a potential fix for #585. We think that the issue might be due to a mismatch in our accounting of how many requests have been granted and the accounting inside of the `rate.Limiter`. If that is the problem, this PR fixes it by creating a fresh `rate.Limiter` every time at midnight UTC. Instead of trying to track the number of tokens we need to drain, we simply drain all of them.


albrow:
Superseded by https://github.com/0xProject/0x-mesh/pull/596

Issue #567, opened 2019-11-30
core: Add timestamp to GetOrders snapshot of when snapshot taken


fabioberger:
Currently we don't indicate to the caller at what point in time the DB snapshot was taken at which they are able to fetch all orders stored by the Mesh node.

### Implementation

- Add a `snapshotTimestamp` field to the `GetOrders` response as well as a `createdAt` field to the `Snapshot` struct.

fabioberger:
Addressed in #591

PR #591, opened 2019-12-13
rpc,core: Add timestamp to GetOrders snapshots

fabioberger:
Fixes: https://github.com/0xProject/0x-mesh/issues/567

### DIFF
--- a/cmd/mesh/rpc_handler.go
+++ b/cmd/mesh/rpc_handler.go
@@ -85,6 +85,9 @@ func (handler *rpcHandler) GetOrders(page, perPage int, snapshotID string) (resu
 		if _, ok := err.(core.ErrSnapshotNotFound); ok {
 			return nil, err
 		}
+		if _, ok := err.(core.ErrPerPageZero); ok {
+			return nil, err
+		}
 		// We don't want to leak internal error details to the RPC client.
 		log.WithField("error", err.Error()).Error("internal error in AddOrders RPC call")
 		return nil, constants.ErrInternal
--- a/core/core.go
+++ b/core/core.go
@@ -598,12 +598,23 @@ func (e ErrSnapshotNotFound) Error() string {
 	return fmt.Sprintf("No snapshot found with id: %s. To create a new snapshot, send a request with an empty snapshotID", e.id)
 }
 
+// ErrPerPageZero is the error returned when a GetOrders request specifies perPage to 0
+type ErrPerPageZero struct{}
+
+func (e ErrPerPageZero) Error() string {
+	return "perPage cannot be zero"
+}
+
 // GetOrders retrieves paginated orders from the Mesh DB at a specific snapshot in time. Passing an empty
 // string as `snapshotID` creates a new snapshot and returns the first set of results. To fetch all orders,
 // continue to make requests supplying the `snapshotID` returned from the first request. After 1 minute of not
 // received further requests referencing a specific snapshot, the snapshot expires and can no longer be used.
 func (app *App) GetOrders(page, perPage int, snapshotID string) (*rpc.GetOrdersResponse, error) {
 	<-app.started
+
+	if perPage <= 0 {
+		return nil, ErrPerPageZero{}
+	}
 
 	ordersInfos := []*rpc.OrderInfo{}
 	var snapshot *db.Snapshot
@@ -646,14 +657,6 @@ func (app *App) GetOrders(page, perPage int, snapshotID string) (*rpc.GetOrdersR
 			ExpirationTimestamp: expirationTimestamp,
 		}
 		app.muIdToSnapshotInfo.Unlock()
-	}
-
-	if perPage <= 0 {
-		return &rpc.GetOrdersResponse{
-			SnapshotID:        snapshotID,
-			SnapshotTimestamp: createdAt,
-			OrdersInfos:       ordersInfos,
-		}, nil
 	}
 
 	notRemovedFilter := app.db.Orders.IsRemovedIndex.ValueFilter([]byte{0})
--- a/core/core.go
+++ b/core/core.go
@@ -141,6 +141,7 @@ type Config struct {
 
 type snapshotInfo struct {
 	Snapshot            *db.Snapshot
+	CreatedAt           time.Time
 	ExpirationTimestamp time.Time
 }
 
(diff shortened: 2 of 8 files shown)


core/core.go
```diff
@@ -597,22 +598,27 @@ func (e ErrSnapshotNotFound) Error() string {
 	return fmt.Sprintf("No snapshot found with id: %s. To create a new snapshot, send a request with an empty snapshotID", e.id)
 }
 
+// ErrPerPageZero is the error returned when a GetOrders request specifies perPage to 0
+type ErrPerPageZero struct{}
+
+func (e ErrPerPageZero) Error() string {
```
albrow: I don't think we need a special type for this error. In general, you only need to use typed errors for cases where the caller might be expected to programmatically react to the error. In other words: "If the function returns this error, I need to do this, but if it returns a different error I need to do that." I can't imagine that being the case here.

If you _do_ want to still use a specific named error, this should just be a variable. Error types should be used when you need to include additional information about the error, e.g., in struct fields. Otherwise, you can just use a variable. See [this guide on naming errors](https://github.com/golang/go/wiki/Errors#naming). I would also suggest tweaking the language a bit since it is also returned when `perPage` is less than zero.

```
var ErrInvalidPerPage = errors.New("perPage cannot be zero or less")
```

Though again, IMO it makes more sense to just return `errors.New("perPage cannot be zero or less")` inside the function instead of declaring this ahead of time.

rpc/client_server_test.go
```diff
@@ -218,6 +220,7 @@ func TestGetOrdersSuccess(t *testing.T) {
 	assert.Len(t, getOrdersResponse.OrdersInfos, 1)
 
 	assert.Equal(t, returnedSnapshotID, getOrdersResponse.SnapshotID, "SnapshotID did not match")
+	assert.Equal(t, expectedSnapshotTimestamp, getOrdersResponse.SnapshotTimestamp, "SnapshotTimestamp did not match")
```
albrow: If you get unlucky with the timing of this test, the timestamp in the response can be one second after `expectedSnapshotTimestamp`, which would cause the test to fail. Can we use [`assert.WithinDuration`](https://godoc.org/github.com/stretchr/testify/assert#WithinDuration) to account for this?

rpc/clients/typescript/src/ws_client.ts
```diff
@@ -278,17 +279,19 @@ export class WSClient {
      * @param perPage number of signedOrders to fetch per paginated request
      * @returns all orders, their hash and their fillableTakerAssetAmount
      */
-    public async getOrdersAsync(perPage: number = 200): Promise<OrderInfo[]> {
+    public async getOrdersAsync(perPage: number = 200): Promise<GetOrdersResponse> {
```
albrow: This is a breaking change so we need to bump the next release number up a major version.

rpc/clients/typescript/src/ws_client.ts
```diff
@@ -278,17 +279,19 @@ export class WSClient {
...
         let page = 0;
-        const getOrdersResponse: GetOrdersResponse = await this._wsProvider.send('mesh_getOrders', [
+        const rawGetOrdersResponse: RawGetOrdersResponse = await this._wsProvider.send('mesh_getOrders', [
             page,
             perPage,
             snapshotID,
         ]);
-        snapshotID = getOrdersResponse.snapshotID;
-        let ordersInfos = getOrdersResponse.ordersInfos;
+        snapshotID = rawGetOrdersResponse.snapshotID;
```
albrow: Can we move this conversion to a helper function/method? The type conversions in this file (and in the __browser/__ code) are one of the things we need to improve test coverage for and this would make it easier to test. The previous code didn't really do any conversion.

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -84,7 +84,8 @@ describe('WSClient', () => {
 
             const client = new WSClient(`ws://localhost:${SERVER_PORT}`);
             const perPage = 1;
-            const orderInfos = await client.getOrdersAsync(perPage);
+            const getOrdersResponse = await client.getOrdersAsync(perPage);
+            const orderInfos = getOrdersResponse.ordersInfos;
```
albrow: Can you add `expect` statements here for the timestamp and snapshot ID?

core/core.go
fabioberger: We do programmatically react to this error. See line 88 in `rpc_handler.go`

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -84,7 +84,8 @@ describe('WSClient', () => {
 
             const client = new WSClient(`ws://localhost:${SERVER_PORT}`);
             const perPage = 1;
-            const orderInfos = await client.getOrdersAsync(perPage);
+            const getOrdersResponse = await client.getOrdersAsync(perPage);
```
jalextowle: Optional: Nit: Since the only part of `getOrdersResponse` that is used in this test is `ordersInfo`, you could also use object destructuring here.

```
const { ordersInfo } = await client.getOrdersAsync(perPage);
```

fabioberger: We care about all three here now.

cmd/mesh/rpc_handler.go
```diff
@@ -85,6 +85,9 @@ func (handler *rpcHandler) GetOrders(page, perPage int, snapshotID string) (resu
 		if _, ok := err.(core.ErrSnapshotNotFound); ok {
 			return nil, err
 		}
+		if _, ok := err.(core.ErrPerPageZero); ok {
```
fabioberger: @albrow we use this error programmatically here

core/core.go
albrow: Ah sorry I missed that. Makes sense to me.

rpc/clients/typescript/src/types.ts
```diff
@@ -396,11 +396,18 @@ export interface ValidationResults {
...
 }
 
-export interface GetOrdersResponse {
+export interface RawGetOrdersResponse {
     snapshotID: string;
+    snapshotTimestamp: string;
     ordersInfos: RawAcceptedOrderInfo[];
 }
 
+export interface GetOrdersResponse {
```
albrow: Let's document this new exported interface, especially the fact that `snapshotTimestamp` is seconds since the Unix Epoch in UTC.

PR #596, opened 2019-12-16
Simplify Ethereum RPC rate limiting

albrow:
This PR simplifies our implementation of Ethereum RPC rate limiting in order to address some potential bugs and make it easier to test. In general, it also loosens the restrictions of the rate limiter to make it less likely to affect non-spammers under normal network conditions.

This is most likely a fix for #585.

Summary of changes:

1. Removed the `rate.Limiter` that was used to enforce `ETHEREUM_RPC_MAX_REQUESTS_PER_24_HR_UTC`. This limit is now enforced via a simple counter which is reset every UTC midnight. This greatly simplifies our implementation and makes it easier to write tests. It also reduces the chances of artificially limiting the rate of requests for non-spammers (especially around UTC midnight). But it does come with the downside of not guaranteeing that RPC requests will be evenly distributed throughout the 24 hour period.
2. Removed the buffer for `maxRequestsPer24Hrs`. The consequences for exceeding this limit are usually not severe, so we feel the buffer is no longer necessary.
3. Removed the `lowestPossibleMaxRequestsPer24Hrs` constraint. We already have a check in the `core` package to prevent you from setting `ETHEREUM_RPC_MAX_REQUESTS_PER_24_HR_UTC` too low based on `BLOCK_POLLING_INTERVAL`.
4. Added a small burst to the per second `rate.Limiter`. Currently set to `maxRequestsPerSecond/2`. IMO the old implementation which set the burst to `1` was too restrictive. It guaranteed that we would never exceed `maxRequestsPerSecond` for any one second period but also enforced a _minimum delay_ of `1/maxRequestsPerSecond` seconds between subsequent requests. This delay was enforced even after long periods of not sending any requests. This works out to 30ms with the default settings, which is nearly twice the median response time for some Ethereum hosting providers (our Alchemy dashboard is showing 16ms). This can be a significant delay for some time-sensitive/performance-sensitive use cases.

### DIFF
--- a/ethereum/ratelimit/rate_limiter.go
+++ b/ethereum/ratelimit/rate_limiter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
@@ -22,6 +23,8 @@ const (
 	lowestPossibleMaxRequestsPer24Hrs = 40000
 )
 
+var ErrTooManyRequestsIn24Hours = errors.New("too many Ethereum RPC requests have been sent this 24 hour period")
+
 // RateLimiter is the interface one must satisfy to be considered a RateLimiter
 type RateLimiter interface {
 	Wait(ctx context.Context) error
@@ -33,7 +36,6 @@ type RateLimiter interface {
 // rateLimiter is a rate-limiter for requests
 type rateLimiter struct {
 	maxRequestsPer24Hrs   int
-	twentyFourHourLimiter *rate.Limiter
 	perSecondLimiter      *rate.Limiter
 	currentUTCCheckpoint  time.Time // Start of current UTC 24hr period
 	grantedInLast24hrsUTC int       // Number of granted requests issued in last 24hr UTC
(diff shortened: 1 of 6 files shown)


ethereum/ratelimit/rate_limiter.go
```diff
@@ -143,29 +105,17 @@ func (r *rateLimiter) Start(ctx context.Context, checkpointInterval time.Duratio
 			case <-ctx.Done():
 				return
 			case <-r.aClock.After(untilNextUTCCheckpoint):
-				// Compute how many grants have accrued and gone unused and remove that
-				// many from the bucket so that it starts empty for the next 24hr period
+				// Reset the number of requests granted and the set the next UTC
```
fabioberger:
```suggestion
				// Reset the number of requests granted and set the next UTC
```

ethereum/ratelimit/rate_limiter_test.go
```diff
@@ -112,57 +102,17 @@ func TestScenario2(t *testing.T) {
...
-		if i == 0 {
-			expectedDuration := (500 * time.Millisecond) + maxExpectedDelay
-			delta := elapsed - expectedDuration
-			assert.True(t, delta < expectedDuration/10, "Delta between expected and elapsed duration too large")
-		} else {
-			// Subsequent requests take 24hrs / maxRequestsPer24hrs
-			delta := math.Abs(float64(maxExpectedDelay - elapsed))
-			assert.True(t, time.Duration(delta) < expectedDeltaMaxExpectedDelay, "Delta between maxExpectedDelay and rate limit delay too large")
-		}
+	// Subsequent reuqests should result in ErrTooManyRequestsIn24Hours
```
fabioberger:
```suggestion
	// Subsequent requests should result in ErrTooManyRequestsIn24Hours
```

ethereum/ratelimit/rate_limiter_test.go
```diff
@@ -247,7 +203,7 @@ func TestScenario4(t *testing.T) {
 
 	// If we are not properly converting times to UTC, instantiation will fail with err
 	// `Wait(n=450000) exceeds limiter's burst 300000`
```
fabioberger:
```suggestion
	// `Wait(n=150000) exceeds limiter's burst 100000`
```

Issue #583, opened 2019-12-11
Possible nil pointer exception in browser


albrow:
Augur reported this issue on version `7.2.1-beta-0xv3`:

![unknown](https://user-images.githubusercontent.com/800857/70657570-eaa5a480-1c10-11ea-9378-92c11668a34a.png)

albrow:
Possibly related to https://github.com/0xProject/0x-mesh/issues/537. The stack trace is similar.

PR #597, opened 2019-12-17
Log error when validationResults contains unknown order hash

albrow:
See https://github.com/0xProject/0x-mesh/issues/583. The nil pointer exception appears to be occurring because `validationResults` contains an order but that order's hash is not found in `orderHashToDBOrder`.

I took a look at the code path, and I couldn't determine why this might be the case. This PR doesn't fix the underlying issue, but it does log the order hash and stop us from panicking which can help us understand what's going on.


albrow:
> This looks good. Do we want to potentially also log all the orderHashes orderHashToDBOrder does contain? Could be helpful in debugging.

Good idea. I just decided to log all of `orderHashToDBOrder` and `validationResults`. Whatever is happening is likely really rare and this will help us understand what's going on.

Issue #557, opened 2019-11-22
rpc: Add timestamps to order events


fabioberger:
Currently we emit order events without any timestamp to elucidate when the order update happened. We should.

- For events emanating from block events, we should use the block timestamp.
- For newly discovered orders received from RPC or P2P connections, a timestamp of after the order has been received and validated should be used.

chowryan:
Will timestamp be added to the `mesh_getOrders` snapshot api too? or is this just being added to the `order` topics stream?

fabioberger:
@chowryan opened a [separate issue](https://github.com/0xProject/0x-mesh/issues/567) for snapshot timestamps but will implement both in the same PR.

fabioberger:
Addressed in #602

PR #602, opened 2019-12-19
zeroex,rpc,browser: Add Timestamp to OrderEvents

fabioberger:
Fixes: #557 

For Mesh-specific events, we use the current UTC time. For on-chain state triggered events, we use the block height at which the order was re-validated.

### DIFF
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -458,7 +458,7 @@ interface WrapperOrderEvent {
  * or filled.
  */
 export interface OrderEvent {
-    timestamp: number;
+    timestampMs: number;
     orderHash: string;
     signedOrder: SignedOrder;
     endState: OrderEventEndState;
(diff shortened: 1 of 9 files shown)


browser/ts/index.ts
```diff
@@ -842,6 +844,8 @@ function signedOrderToWrapperSignedOrder(signedOrder: SignedOrder): WrapperSigne
 function wrapperOrderEventToOrderEvent(wrapperOrderEvent: WrapperOrderEvent): OrderEvent {
     return {
         ...wrapperOrderEvent,
+        // tslint:disable-next-line:custom-no-magic-numbers
+        timestamp: Math.round(new Date(wrapperOrderEvent.timestamp).getTime() / 1000),
```
albrow: I don't think we should reduce the precision here by dividing by 1000 and rounding. It seems like the extra precision could be useful.

How about either leaving the date as a string or converting it to milliseconds since the Unix Epoch instead of seconds? If we do go with milliseconds I would also suggest changing the name to something like `timestampMilliseconds` to reduce ambiguity.

albrow:
I requested a small change and also want to make sure CI passes.

PR #609, opened 2019-12-20
Merge V3 support into `development` branch

fabioberger:


albrow:
@fabioberger I think we agreed to go with a different approach. Can we close this?

PR #610, opened 2019-12-21
Special instructions for running w/ Ganache

feuGeneA:
Documenting so that other people won't hit https://github.com/0xProject/0x-monorepo/pull/2413


albrow:
@feuGeneA please read https://github.com/0xProject/0x-mesh/blob/master/CONTRIBUTING.md. New PRs need to be based on the `development` branch.

@fabioberger do we really need to set to 50ms for Ganache? That seems really low.

fabioberger:
@albrow if devs don't want to sleep in their tests, then the block polling must be set to a really low value unfortunately.

feuGeneA:
@fabioberger No, I did not try to see if it works with higher values.  It was a monorepo-wide blocker, so I just got it working ASAP, and then I got focused on other things.

albrow:
@feuGeneA @fabioberger I don't think this suggested change is the right approach. The suggested numbers are basically pulled out of nowhere and obfuscate the real underlying problem, which is that Mesh does not process mined blocks instantaneously. There is no possible value for `BLOCK_POLLING_INTERVAL` that guarantees that blocks will be processed when you need them to be. If you need Mesh to process the blocks before continuing a test, it is more reliable to just add a sleep statement (though obviously this is still not ideal).

I'm leaning towards just closing this. Any opposition?

PR #611, opened 2019-12-23
Typescript RPC Client Integration Tests

jalextowle:
Updates most of the typescript RPC client tests to be integration tests rather than unit tests. Some functions that were exempt from this refactor include `onClose`, `onReconnection`, and `destroy` because unit tests actually do a good job of testing the desired behavior.

As well as creating integration tests, this PR introduces typescript snippets that will hopefully become common idioms for interacting with mesh nodes in typescript.


rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -1,507 +1,434 @@
...
+                (async () => {
+                    // Add an order and then cancel it.
+                    const order = await orderFactory.newSignedOrderAsync({});
+                    const validationResults = await deployment.client.addOrdersAsync([ order ]);
+                    expect(validationResults.accepted.length).to.be.eq(1);
+                    await exchange.cancelOrder(order).awaitTransactionSuccessAsync({ from: makerAddress });
+
+                    // Subscribe to order events and assert that only a single cancel event was received.
+                    const now = new Date(Date.now()).getTime();
+                    await deployment.client.subscribeToOrdersAsync((orderEvents: OrderEvent[]) => {
```
fabioberger: It's not guaranteed that you'll receive events if thdey occurred before you subscribe. Let's therefore subscribe before we cancel the order.

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -1,507 +1,434 @@
...
+                        expect(parameters.feeRecipientAddress.toLowerCase()).to.be.eq(order.feeRecipientAddress);
+                        expect(parameters.makerAddress.toLowerCase()).to.be.eq(makerAddress);
+                        expect(parameters.makerAssetData).to.be.eq(order.makerAssetData);
+                        expect(parameters.orderHash).to.be.eq(orderHashUtils.getOrderHashHex(order));
+                        expect(parameters.senderAddress.toLowerCase()).to.be.eq(makerAddress);
+                        expect(parameters.takerAssetData).to.be.eq(order.takerAssetData);
+                        done();
+                    });
+                })().catch(done);
+            });
```
fabioberger: Let's add a test where we create a subscription, and then unsubscribe with the `subscriptionID` received upon subscription.

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -1,507 +1,434 @@
...
-            expect(BigNumber.isBigNumber(validationResults.accepted[0].signedOrder.expirationTimeSeconds)).to.equal(
-                true,
+
+        afterEach(async () => {
+            deployment.client.destroy();
+            deployment.mesh.stopMesh();
+        });
+
+        async function deployErc20TokenAsync(name: string, symbol: string): Promise<DummyERC20TokenContract> {
+            return DummyERC20TokenContract.deployFrom0xArtifactAsync(
```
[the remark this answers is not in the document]
jalextowle: I'll have to get used to not always deploying contracts manually 😄

rpc/clients/typescript/test/utils/ws_server.ts
```diff
@@ -0,0 +1,99 @@
...
+            setTimeout(reject, timeout || MeshManager.DEFAULT_TIMEOUT);
+        });
+    }
+
+    /**
+     * Kill the mesh process of this mesh instance.
+     */
+    public stopMesh(): void {
+        this._killed = true;
+        this._mesh.kill();
```
albrow: If we're done with `this._mesh` and you're sure we don't need to use it anymore, let's use `SIGKILL` here to make sure that the process definitely exits. We probably don't care about preserving data or shutting down gracefully. The default signal, `SIGTERM`, can be intercepted and is not guaranteed to kill the process.

```suggestion
        this._mesh.kill('SIGKILL');
```

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This changelog is a work in progress and may contain notes for versions which ha
 
 ### Bug fixes 🐞
 
+- Fixed a typo ("rendervouz" --> "rendezvous") in GetStatsResponse. ([#611](https://github.com/0x-mesh/pull/611)).
```
albrow: @jalextowle this isn't the right place to add the entry to the changelog. `v8.0.0-beta-0xv3` has already been released. You need to add a heading for a new version called `v8.0.1` (without the `beta` or `0xv3` suffix this time).

PR #612, opened 2019-12-24
Added fix for webworkers

joeykrug:
Window == self in a browser in general, but in a webworker, there is no window, only self. So this switches to self so it works everywhere!


albrow:
There are a few other places where Mesh and its dependencies use `window`. This PR is probably not sufficient and we need to do more testing to see if running Mesh in a Web Worker is even possible.

PR #614, opened 2019-12-28
orderwatch: Avoid conflicting txn error when updating block headers in DB

fabioberger:

### DIFF
--- a/db/transaction.go
+++ b/db/transaction.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"sync/atomic"
 
@@ -9,10 +10,19 @@ import (
 )
 
 var (
-	ErrDiscarded             = errors.New("transaction has already been discarded")
-	ErrCommitted             = errors.New("transaction has already been committed")
-	ErrConflictingOperations = errors.New("cannot perform more than one operation (insert/delete/update) on the same model within a transaction")
+	ErrDiscarded = errors.New("transaction has already been discarded")
+	ErrCommitted = errors.New("transaction has already been committed")
 )
+
+// ConflictingOperationsError is returned when two conflicting operations are attempted within the same
+// transaction
+type ConflictingOperationsError struct {
+	operation string
+}
+
+func (e ConflictingOperationsError) Error() string {
+	return fmt.Sprintf("cannot perform more than one operation (%s) on the same model within a transaction", e.operation)
+}
 
 // Transaction is an atomic database transaction for a single collection which
 // can be used to guarantee consistency.
@@ -139,7 +149,7 @@ func (txn *Transaction) Insert(model Model) error {
 		return err
 	}
 	if txn.affectedIDs.Contains(string(model.ID())) {
-		return ErrConflictingOperations
+		return ConflictingOperationsError{operation: "insert"}
 	}
 	if err := insertWithTransaction(txn.colInfo, txn.readWriter, model); err != nil {
 		return err
@@ -159,7 +169,7 @@ func (txn *Transaction) Update(model Model) error {
 		return err
 	}
 	if txn.affectedIDs.Contains(string(model.ID())) {
-		return ErrConflictingOperations
+		return ConflictingOperationsError{operation: "update"}
 	}
 	if err := updateWithTransaction(txn.colInfo, txn.readWriter, model); err != nil {
 		return err
@@ -178,7 +188,7 @@ func (txn *Transaction) Delete(id []byte) error {
 		return err
 	}
 	if txn.affectedIDs.Contains(string(id)) {
-		return ErrConflictingOperations
+		return ConflictingOperationsError{operation: "delete"}
 	}
 	if err := deleteWithTransaction(txn.colInfo, txn.readWriter, id); err != nil {
 		return err
--- a/db/transaction.go
+++ b/db/transaction.go
@@ -21,7 +21,7 @@ type ConflictingOperationsError struct {
 }
 
 func (e ConflictingOperationsError) Error() string {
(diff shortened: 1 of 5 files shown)


db/transaction.go
```diff
@@ -2,18 +2,28 @@ package db
...
 )
 
+// ConflictingOperationsError is returned when two conflicting operations are attempted within the same
+// transaction
+type ConflictingOperationsError struct {
+	operation string
+}
+
+func (e ConflictingOperationsError) Error() string {
+	return fmt.Sprintf("cannot perform more than one operation (%s) on the same model within a transaction", e.operation)
```
albrow:
```suggestion
	return fmt.Sprintf("error on %s: cannot perform more than one operation on the same model within a transaction", e.operation)
```

Just a slight re-wording here. The current version doesn't make it clear whether `operation` is the first operation or the second one.

zeroex/orderwatch/order_watcher.go
```diff
@@ -1480,16 +1529,21 @@ type orderDeleter interface {
...
+				"error": err.Error(),
+				"order": order,
+			}).Error("Failed to permanently delete order")
+			return nil
+		}
+		if _, ok := err.(db.NotFoundError); ok {
+			logger.WithFields(logger.Fields{
+				"error": err.Error(),
+				"order": order,
+			}).Warn("Attempted to delete order that no longer exists")
```
albrow: How do you feel about removing this log message? I don't think we're tracking it or planning to take any further action related to it.

albrow:
@fabioberger can we also add a regression test here? Or at the very least add a test for `updateBlockHeadersStoredInDB`?

zeroex/orderwatch/order_watcher.go
fabioberger: Yep, fine by me.

fabioberger:
Added comprehensive tests for `updateBlockHeadersStoredInDB`

zeroex/orderwatch/order_watcher.go
```diff
@@ -1011,6 +1008,62 @@ func (w *Watcher) trimOrdersAndGenerateEvents() ([]*zeroex.OrderEvent, error) {
 	return orderEvents, nil
 }
 
+// updateBlockHeadersStoredInDB updates the block headers stored in the DB. Since our DB txns don't support
+// multiple operations involving the same entry, we make sure we only perform either an insertion or a deletion
+// for each block in this method.
+func (w *Watcher) updateBlockHeadersStoredInDB(miniHeadersColTxn *db.Transaction, events []*blockwatch.Event) error {
```
albrow: It looks like we don't actually use `w` inside this method. If we change this from a method to a function, it will be a lot easier to test. We wouldn't have to set up any blockchain state and it could be run in parallel (as long as we use a randomly generated name for the db directory).

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -1027,6 +1027,243 @@ func TestOrderWatcherCleanup(t *testing.T) {
...
+	defer teardownSubTest(t)
+
+	meshDB, err := meshdb.New("/tmp/leveldb_testing/" + uuid.New().String())
+	require.NoError(t, err)
+
+	ctx, cancelFn := context.WithCancel(context.Background())
+	defer cancelFn()
+	_, orderWatcher := setupOrderWatcher(ctx, t, ethRPCClient, meshDB)
+
+	// Scenario 1: Header 1 exists in DB. Get's removed and then re-added.
```
albrow: These should all be independent test cases in order to make sure they aren't interfering with each other. If we change `updateBlockHeadersStoredInDB` to a function, it should be pretty straightforward to use table-driven tests here.

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -1027,6 +1027,279 @@ func TestOrderWatcherCleanup(t *testing.T) {
 	}
 }
 
+// Scenario 1: Header 1 exists in DB. Get's removed and then re-added.
+func TestOrderWatcherUpdateBlockHeadersStoredInDBScenario1(t *testing.T) {
```
albrow: @fabioberger these tests are isolated, but they are repetitive and IMO overly verbose. Snippets like this are duplicated many times:

```go
miniHeaders := []*miniheader.MiniHeader{}
err = meshDB.MiniHeaders.FindAll(&miniHeaders)
require.NoError(t, err)
assert.Len(t, miniHeaders, 1)
```

It distracts from what we are really trying to test: "this input should result in that output".

What I was suggesting was using table-driven tests, similar to what we do [in the `db` package](https://github.com/0xProject/0x-mesh/blob/dee8353e2e51862920745f5006e5fd38ab162b4b/db/query_test.go#L265) and a few other places in the repo. This style of testing is based on https://github.com/golang/go/wiki/TableDrivenTests and is considered a best practice in the Go ecosystem.

Can you rewrite these tests in the table-driven style? Something like:

```go
testCases := []struct{
    events []*blockwatch.Event
    expectedMiniHeaders []*miniheader.MiniHeader
}{
    {
        events: []*blockwatch.Event{
            // Put the events here. These will be passed directly to updateBlockHeadersStoredInDB
        },
        expectedMiniHeaders: []*miniheader.MiniHeader{
            // Put the expected mini headers here. What do we expect to find in the database after calling updateBlockHeadersStoredInDB?
        }
    }
}

for _, testCase := range testCases {
    // 1. Call updateBlockHeadersStoredInDB
    // 2. Call meshDB.MiniHeaders.FindAll(&miniHeaders)
    // 3. Expect that the miniheaders in the database match testCase.expectedMiniHeaders
}
```

This approach makes the tests far more DRY and easier to understand.

fabioberger: I see, sorry for missing that. Done.

Issue #608, opened 2019-12-20
orderwatch: handleBlockEvents expects multiple updates per txn


fabioberger:
Currently the way `handleBlockEvents` is written, we update order DB entries both in `handleOrderExpirations` and `generateOrderEventsIfChanged`. This is only a problem because our DB transactions do not support applying multiple updates to a single DB record. 

To fix this, we should merge the logic in `handleOrderExpirations` into `generateOrderEventsIfChanged` and enforce that only a single DB update is performed per order in `handleBlockEvents`.

This bug does not seem urgent given the following: There are two outcomes of `handleOrderExpirations`, the order is either expired or unexpired. If expired, the DB is updated as such and the next update would have applied the same outcome. The error thrown is logged but does not otherwise disrupt `handleBlockEvents`. If the order became unexpired (which is rare), we would likewise update the DB properly. If the order also became filled/cancelled in the same block that caused it to be "unexpired", we would have the incorrect `fillableTakerAmount` in the DB. This is even rarer than becoming unexpired in the first place but still problematic. Thus, we should fix this issue, but it should very rarely affect the correctness of the watcher in the mean time.

PR #623, opened 2020-01-06
orderwatch: Avoid updating orders multiple times in same txn (part 2)

fabioberger:
Fixes: https://github.com/0xProject/0x-mesh/issues/608

Currently it is possible that we attempt to update orders that expire or unexpire twice within the same txn. This causes the second update to noop since our DB txns do not support multiple updates within a single txn. This PR stops this from happening.

### DIFF
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -342,6 +342,10 @@ func (w *Watcher) removedCheckerLoop(ctx context.Context) error {
 	}
 }
 
+// handleOrderExpirations takes care of generating expired and unexpired order events for orders that do not require re-validation.
+// Since expiry is now done according to block timestamp, we can figure out which orders have expired/unexpired statically. We do not
+// process blocks that require re-validation, since the validation process will already emit the necessary events and we cannot make
+// multiple updates to an order within a single DB transaction.
 func (w *Watcher) handleOrderExpirations(ordersColTxn *db.Transaction, latestBlockTimestamp, previousLatestBlockTimestamp time.Time, ordersToRevalidate map[common.Hash]*meshdb.Order) ([]*zeroex.OrderEvent, error) {
 	orderEvents := []*zeroex.OrderEvent{}
 	var defaultTime time.Time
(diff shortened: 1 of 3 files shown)


zeroex/orderwatch/order_watcher.go
```diff
@@ -342,14 +342,19 @@ func (w *Watcher) removedCheckerLoop(ctx context.Context) error {
 	}
 }
 
-func (w *Watcher) handleOrderExpirations(ordersColTxn *db.Transaction, latestBlockTimestamp time.Time, previousLatestBlockTimestamp time.Time) ([]*zeroex.OrderEvent, error) {
+func (w *Watcher) handleOrderExpirations(ordersColTxn *db.Transaction, latestBlockTimestamp, previousLatestBlockTimestamp time.Time, ordersToRevalidate map[common.Hash]*meshdb.Order) ([]*zeroex.OrderEvent, error) {
```
albrow: This function has become so complex that I think a brief comment would really help legibility. What does the function do? What are the different parameters and how are they used?

zeroex/orderwatch/order_watcher.go
```diff
@@ -1142,35 +1159,52 @@ func (w *Watcher) generateOrderEventsIfChanged(
...
+				orderEvent := &zeroex.OrderEvent{
+					Timestamp:                validationBlockTimestamp,
+					OrderHash:                acceptedOrderInfo.OrderHash,
+					SignedOrder:              order.SignedOrder,
+					EndState:                 zeroex.ESOrderFillabilityIncreased,
+					FillableTakerAssetAmount: acceptedOrderInfo.FillableTakerAssetAmount,
+					ContractEvents:           orderHashToEvents[order.Hash],
+				}
+				orderEvents = append(orderEvents, orderEvent)
 			}
```
jalextowle: Optional: There is some code duplication in these blocks that can be removed.

fabioberger: Will leave as is so we don't need to add an additional conditional nesting.

Issue #551, opened 2019-11-21
Is GossipSub working as expected?


weijiewu8:
If GossipSub is working as expected then orders should be propagated really quickly. 
Consider the time point when an order is propagated for the first time by some peer in Mesh, for all peers that are online at that moment, they should receive the order really quickly (hopefully within 1-2 seconds). This is because according to the mechanism of GossipSub, a subset of the neighboring peers (between 4 to 12, by default) should receive the order from the first sender, then each of these neighboring peers should immediately forward this order to its own neighboring peers (again, between 4 to 12), and the process should be going on so that nearly all peers in Mesh should receive the order really quickly. 

However, we found many cases per day where an order can take more than 10 seconds to reach the peers, and there are occasionally cases that take minutes. It's questionable whether GossipSub is working as we expect.

albrow:
This should be fixed by #692. We just need to run more simulations/tests to confirm.

PR #630, opened 2020-01-08
Implement custom order filtering

### LINKED ISSUE #336: Add support for custom topics

albrow:
This PR simply rebases https://github.com/0xProject/0x-mesh/pull/563 on the new `development` branch. I'm keeping the old PR around just in case anything was forgotten/lost during the messy rebase.

Fixes #336 and #337.

### Overview

This PR implements custom order filtering and custom topics. A custom filter may be passed into Mesh as a [JSON Schema](https://json-schema.org/) via the new `CUSTOM_ORDER_FILTER` environment variable. Messages that contain orders that don't match this schema will be dropped. As a limitation, filtering is only possible by looking at the static fields of an order. So for example, it is not possible to filter orders by doing an on-chain check or sending an HTTP request to a third-party API. We don't expect that this limitation is going to be a problem in practice and it comes with the huge benefit of enabling cross-topic forwarding in the future (more on that later).

This PR also changes the way that pubsub topics work in Mesh. Previously, there was only one topic and it was hard-coded based on the chain ID. Now, there is a one-to-one correspondence between custom filters and topics. You can convert from custom order schema to a topic and vice versa. This is a breaking change and the old topic names are not compatible with this PR. (However, with this change we can potentially implement cross-version forwarding for forward compatibility from now on).

### New order and message schemas.

All orders must match the following JSON Schema:

```json
{
	"id": "/rootOrder",
	"allOf": [{
		"$ref": "/customOrder"
	}, {
		"$ref": "/signedOrder"
	}]
}
```

- `/signedOrder` is the same JSON Schema as before and will match only valid 0x orders.
- `/customOrder` is the custom schema passed in through the `CUSTOM_ORDER_FILTER` environment variable.

Organizing the JSON Schema for orders like this means that `CUSTOM_ORDER_FILTER` can be relatively small. It doesn't need to contain all the required fields for a signed 0x order. It just needs to contain any _additional_ requirements on top of the default ones.

The new schema for order messages sent through pubsub looks like this:

```json
{
	"id": "/rootOrderMessage",
	"properties": {
		"messageType": {
			"type": "string",
			"pattern": "order"
		},
		"order": {
			"$ref": "/rootOrder"
		},
		"topics": {
			"type": "array",
			"minItems": 1,
			"items": {
				"type": "string"
			}
		}
	},
	"required": ["messageType", "order", "topics"]
}
```

Notably this adds the `topics` field which is a list of topics that the message was sent on. This is critical for cross-topic forwarding.

#### Example custom order schemas

The following `CUSTOM_ORDER_FILTER` doesn't add any additional requirements. All valid signed 0x orders will be accepted. This is the default value if no custom filter is passed in.

```json
{}
```

The following `CUSTOM_ORDER_FILTER` matches any valid signed 0x orders with a specific sender address:

```json
{
	"properties": {
		"senderAddress": {
			"pattern": "0x00000000000000000000000000000000ba5eba11",
			"type": "string"
		}
	}
}
```

This can easily be tweaked to filter orders by asset type, maker/taker address, or fee recipient.

### New topic strings

The new topic format looks like this:

```typescript
`/0x-orders/version/${versionNumber}/chain/${chainID}/schema/${encodedSchema}`
```

- `versionNumber` is the version for the topic string. This PR changes the default version from 2 to 3.
- `chainID` is the chain ID that Mesh is configured to use.
- `encodedSchema` is the base64-encoded custom order schema. This trick allows us to extract the custom order schema from the topic string.

#### Cross-topic forwarding

Part of the reason this PR is implemented in a particular way is to enable "cross-topic forwarding" in the future. What this means is that a Mesh node that is subscribed on the default topic (all 0x orders) can receive orders on any topic and forward them to nodes that are _only_ subscribed to that topic _without_ subscribing to or even knowing about the topic ahead of time. This PR does not implement cross-topic forwarding but it does add everything we need to do so in the future without breaking backwards-compatibility.

Here's a rough overview of how it will work:

1. Mesh nodes which are using custom filters will publish every order to two topics: (a) the topic corresponding to their custom filter and (b) the default topic. Because of how the JSON Schemas are implemented with `allOf`, we know that any order which is valid for a custom topic would also be considered valid on the default topic.
2. Mesh nodes with cross-topic forwarding enabled will check every message that they receive to see which topics they were originally published to. They can then decode the custom schema from the topic string and validate that the order does match the custom schema. If it matches, they can forward the message to the custom topic. Hopefully we can do this while retaining the original pubsub signature, but it's possible we need to re-sign the message before forwarding.

A similar approach can be used to forward messages between topic versions. In other words a message that was originally sent on topic version 3 could be modified and forwarded to topic version 4. This means we could potentially break backwards compatibility at the pubsub layer without completely fragmenting liquidity (although this approach would add additional latency).


integration-tests/browser/yarn.lock
```diff
@@ -63,7 +63,7 @@
     lodash.values "^4.3.0"
 
 "@0x/mesh-browser@./../../browser/":
-  version "8.0.0-beta-0xv3"
+  version "3.0.1-beta"
```
albrow: This is an auto-generated file and yarn is grabbing the version number from the __package.json__ for the `@0x/mesh-browser` package _locally_. I know this looks weird but it's a side-effect of using a relative import in our __package.json__ instead of importing from npm. I say we leave it be.

albrow:
This PR is ready for review. In the interest of not making this PR even bigger than it already is, the following will be added in separate PRs:

1. In-depth documentation about how to create custom order filters. This will probably exist as a page in the "Advanced Topic" section on https://0x-org.gitbook.io/mesh/.
2. Updating peer scores when receiving an order that doesn't pass your custom filter. This will help you stay connected to peers that are compatible with your custom order filter, which will increase the efficiency of sharing orders. However there are already some important changes and optimizations that will have an even greater impact on order propagation speed and have higher priority (e.g. https://github.com/0xProject/0x-mesh/issues/638 and https://github.com/0xProject/0x-mesh/issues/551).

orderfilter/filter.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, err
+	}
+	if err := loader.AddSchema("/customOrder", jsonschema.NewStringLoader(customOrderSchema)); err != nil {
+		return nil, err
+	}
+	return loader, nil
+}
+
+func NewFromTopic(topic string) (*Filter, error) {
+	// TODO(albrow): Use a cache for topic -> filter
```
fabioberger: Is this still slated for this PR?

orderfilter/filter.go
```diff
@@ -0,0 +1,210 @@
...
+	if f.topic == "" {
+		f.topic = f.generateTopic()
+	}
+	return f.topic
+}
+
+func (v *Filter) generateTopic() string {
+	var holder interface{} = struct{}{}
+	_ = canonicaljson.Unmarshal([]byte(v.rawCustomOrderSchema), &holder)
+	canonicalOrderSchemaJSON, _ := canonicaljson.Marshal(holder)
```
fabioberger: Can you add a comment to why this unmarshalling and marshalling is necessary?

p2p/node.go
```diff
@@ -280,6 +280,41 @@ func New(ctx context.Context, config Config) (*Node, error) {
...
+	}
+	validators.Add("message rate limiting", rateValidator.Validate)
+
+	// Add the custom validator if there is one.
+	if config.CustomMessageValidator != nil {
+		validators.Add("custom", config.CustomMessageValidator)
+	}
+
+	// Register the set of validators for all topics that we publish and/or
+	// subscribe to.
```
fabioberger: So does this mean we perform this validation before we publish messages, and on messages received from a subscription? I'm wondering if custom message validation is necessary on the former? Thoughts?

p2p/validatorset/set.go
```diff
@@ -0,0 +1,67 @@
...
+		case <-ctx.Done():
+			return false
+		default:
+		}
+
+		// Otherwise continue by running this validator.
+		isValid := validator.validator(ctx, sender, msg)
+		if !isValid {
+			// TODO(albrow): Should we reduce a peer's score as a penalty for invalid
+			//               messages?
```
fabioberger: Yes, I think we should.

orderfilter/filter.go
albrow: No. It's something we should add as part of cross-topic forwarding (or shortly afterwards). Currently we don't call `NewFromTopic` except maybe in the tests. The purpose of this function is to work backwards from a topic string to an order filter.

p2p/validatorset/set.go
albrow: @fabioberger See [my comment on this PR](https://github.com/0xProject/0x-mesh/pull/630#issuecomment-573922844). I would prefer to do this in a separate PR. What do you think?

browser/ts/index.ts
```diff
@@ -126,6 +179,26 @@ export interface Config {
...
+    // Here is an example filter which will only allow orders with a specific
+    // makerAssetData:
+    //
+    //    {
+    //        properties: {
+    //            makerAssetData: {
+    //                const: "0xf47261b0000000000000000000000000871dd7c2b4b25e1aa18728e9d5f2af4c4e431f5c"
+    //            }
+    //        }
+    //    }
```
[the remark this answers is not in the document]
albrow: > What about in MAP orders? Is there a way to saw that every asset in the MAP asset data is a GodsUnchained card?

Yes that should be possible. I'm not very good at regex-foo but the pattern would basically be: `The MultiAssetProxy bytes4 prefix followed by one or more of (the ERC721AssetProxy bytes4 prefix followed by a token id)`.

orderfilter/filter.go
albrow: I added the following comment:

```go
	// Note(albrow): We use canonicaljson to elminate any differences in spacing,
	// formatting, and the order of field names. This ensures that two filters
	// that are semantically the same JSON object always encode to exactly the
	// same canonical topic string.
	//
	// So for example:
	//
	//     {
	//         "foo": "bar",
	//         "biz": "baz"
	//     }
	//
	// Will encode to the same topic string as:
	//
	//     {
	//         "biz":"baz",
	//         "foo":"bar"
	//     }
	//
```

p2p/node.go
albrow: Yes, this is a consequence of how the `pubsub` works and can't be easily changed. Luckily validating messages with `jsonschema` is very fast.

fabioberger: Ok, that's fine. Can you please add a comment making this clear so we know that this validation is not strictly necessary but we've decided to do it for now?

p2p/validatorset/set.go
fabioberger: Sounds good.

p2p/node.go
albrow: 👍 I added a comment explaining this.

PR #631, opened 2020-01-09
Increase max order size and create separate max message size

### LINKED ISSUE #615: Up max order size such that a multiAsset order exchanging 35 assets fit

albrow:
Fixes #615.

This PR increases the maximum order size to 16k bytes. It also introduces a separate limit called `MaxMessageSizeInBytes` and a corresponding error `ErrMaxMessageSize`. Previously we were conflating these two limits, but they should technically be different.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This changelog is a work in progress and may contain notes for versions which ha
 
 ### Features ✅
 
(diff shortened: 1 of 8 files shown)


albrow:
@jalextowle can you double check that this order size will not exceed the gas cost limits on mainnet?

PR #634, opened 2020-01-10
Release version 8.1.0

albrow:


albrow:
Blocked by https://github.com/0xProject/0x-mesh/pull/635. Will rebase and update after that PR is merged.

docs/browser/README.md
```diff
@@ -1,3 +1,5 @@
+# @0x/mesh-browser
```
albrow: @fabioberger this doesn't look right. There are two `@0x/mesh-browser` headings. Is this maybe a bug in the `cut-release` script?

fabioberger: This looks like a faulty merge. The `cut-release` doesn't do any more extensive regexes, just version regexes.

PR #635, opened 2020-01-10
Fix browser example and make sure it builds in CI

albrow:
- Updated order format for compatibility with 0x v3
- Upgraded to latest version of `@0x/utils` to fix an issue with `BigNumber`
- Added a new CI step to make sure the browser example builds

### DIFF
--- a/browser/package.json
+++ b/browser/package.json
@@ -1,31 +1,31 @@
 {
-  "name": "@0x/mesh-browser",
-  "version": "3.0.1-beta",
-  "description": "TypeScript and JavaScript bindings for running Mesh directly in the browser.",
-  "main": "./lib/index.js",
-  "license": "Apache-2.0",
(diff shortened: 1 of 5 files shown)


browser/package.json
```diff
@@ -1,31 +1,31 @@
...
-    "typescript": "^3.5.3"
-  },
-  "dependencies": {
-    "@0x/order-utils": "^10.0.1",
-    "@0x/utils": "^5.1.1",
-    "base64-arraybuffer": "^0.2.0",
-    "browserfs": "^1.4.3"
-  }
+    "name": "@0x/mesh-browser",
+    "version": "1.0.0",
```
albrow: The version here doesn't actually matter since it always get's updated when we run `make cut-release`. I think `1.0.0` makes more sense than `3.0.1-beta`.

PR #637, opened 2020-01-10
Fixed the intermitent failures on CI

jalextowle:
Fixed the intermittent issues on CI (my fault). Sorry for the trouble, but hopefully this fixes the issue!


rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -220,13 +219,10 @@ blockchainTests.resets('WSClient', env => {
                     }
 
                     // Subscribe to orders and wait for order events.
-                    let now: number;
                     const subscription = deployment.client.subscribeToOrdersAsync((orderEvents: OrderEvent[]) => {
                         expect(orderEvents.length).to.be.eq(orders.length);
                         for (const orderEvent of orderEvents) {
                             expect(orderEvent.endState).to.be.eq(OrderEventEndState.Added);
-                            // tslint:disable-next-line:custom-no-magic-numbers
-                            assertRoughlyEquals(now, orderEvent.timestampMs, secondsToMs(4));
```
albrow: Can you elaborate on why you removed this? Don't we care about the timestamp?

jalextowle: We do, but I couldn't actually find an upper bound on the timestamp difference. I thought that 4 seconds would be enough, but I've experienced failures locally with that maximum (and 5 seconds). Considering how large of a gap that is, I didn't feel like the assertion was meaningful. That said, I'm happy to add it back if you disagree with this reasoning.

albrow: Let's add it back in but set the range to something like 10 seconds. That still tests that the `timestampMs` is not null and is not zero. It also protects us from order of magnitude mistakes like forgetting to convert seconds to milliseconds.

PR #639, opened 2020-01-13
readme: Add status section

fabioberger:


README.md
```diff
@@ -8,7 +8,13 @@
 0x Mesh is a peer-to-peer network for sharing orders that adhere to the
 [0x order message format](https://0x.org/docs/guides/v2-specification#order-message-format).
 
-WARNING: This project is still under active development. Expect breaking changes before the official release.
+## Project status
+
+we have reached the point where Mesh is being used by some teams in production. We feel that for many use cases, Mesh is stable enough for production use. However, we caution that there are some issues and shortcomings in it's current state, which generally fall into two categories:
```
albrow: I'm pretty sure I wrote this part from the release notes, but I noticed some small mistakes 😛

```suggestion
We have reached the point where Mesh is being used by some teams in production. We feel that for many use cases, Mesh is stable enough for production. However, we caution that there are some issues and shortcomings in its current state, which generally fall into two categories:
```

Issue #619, opened 2020-01-02
Can't start a new instance of Mesh in the same browser window


albrow:
If you are using custom contract addresses, attempting to call `new Mesh()` in the browser more than once results in the following error:

```
config.CustomContractAddresses is invalid: cannot add contract addresses for chain ID 12346: addresses for this chain id are already defined
```

While we don't really want two Mesh instances running simultaneously, it would be useful to be able to _restart_ it by creating a new instance. For example, if Mesh crashes and is no longer running, you would ideally be able to start fresh with `new Mesh()` so users can continue using the page with little to no interruption. The only solution now is to refresh the page, which isn't ideal.

albrow:
Should be fixed by https://github.com/0xProject/0x-mesh/pull/640.

PR #640, opened 2020-01-13
core,ethereum: Allow testnet contract addressses override

fabioberger:
Several developers have been slowed down by having to wait until a smart contract fix be officially included in Mesh. This PR allows them to point their Mesh node to the fixed smart contract on testnets while still guarding mainnet contracts from being swapped out.

### DIFF
--- a/ethereum/contract_addresses.go
+++ b/ethereum/contract_addresses.go
@@ -17,8 +17,8 @@ func GetContractAddressesForChainID(chainID int) (ContractAddresses, error) {
 }
 
 func AddContractAddressesForChainID(chainID int, addresses ContractAddresses) error {
-	if _, alreadExists := ChainIDToContractAddresses[chainID]; alreadExists {
-		return fmt.Errorf("cannot add contract addresses for chain ID %d: addresses for this chain id are already defined", chainID)
+	if _, alreadExists := ChainIDToContractAddresses[chainID]; alreadExists && chainID == 1 {
+		return fmt.Errorf("cannot add contract addresses for chain ID %d: addresses for this chain id are hard-coded", chainID)
 	}
 	if addresses.Exchange == constants.NullAddress {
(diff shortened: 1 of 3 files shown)


ethereum/contract_addresses.go
```diff
@@ -17,8 +17,8 @@ func GetContractAddressesForChainID(chainID int) (ContractAddresses, error) {
 }
 
 func AddContractAddressesForChainID(chainID int, addresses ContractAddresses) error {
-	if _, alreadExists := ChainIDToContractAddresses[chainID]; alreadExists {
-		return fmt.Errorf("cannot add contract addresses for chain ID %d: addresses for this chain id are already defined", chainID)
+	if _, alreadExists := ChainIDToContractAddresses[chainID]; alreadExists && chainID == 1 {
+		return fmt.Errorf("cannot add contract addresses for chain ID %d: addresses for this chain id are hard-coded", chainID)
```
albrow: Historically, this error also happened if a user tried to define contract addresses for the same `chainID` twice (e.g. via a typo in their JSON or by attempting to initialize Mesh more than once. See https://github.com/0xProject/0x-mesh/issues/619).

Why don't we change this to:

```suggestion
if chainID == 1 {
	return fmt.Errorf("cannot add contract addresses for chainID 1 : addresses for mainnet are hard-coded and cannot be changed")
}
```

This would actually fix https://github.com/0xProject/0x-mesh/issues/619, but does mean we miss potential typos in the JSON object that is passed through in the `CUSTOM_CONTRACT_ADDRESSES` environment variable, specifically for cases where the same key is specified twice. I'm okay with that change, and if we really cared about it we could validate the JSON separately.

fabioberger: Yeah, I feel like the multiple keys issue is definitely an edge-case and we could add separate validation for that.

PR #641, opened 2020-01-13
dockerfile: Fix dep install step

fabioberger:
Our migration to Go Modules (https://github.com/0xProject/0x-mesh/pull/618) broke our `dockerfile`'s dep installation step. This PR fixes it.

### DIFF
--- a/dockerfiles/mesh-bootstrap/Dockerfile
+++ b/dockerfiles/mesh-bootstrap/Dockerfile
@@ -13,7 +13,7 @@ WORKDIR /go/src/github.com/0xProject/0x-mesh
 
 ADD . ./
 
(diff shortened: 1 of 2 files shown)


dockerfiles/mesh-bootstrap/Dockerfile
```diff
@@ -13,7 +13,7 @@ WORKDIR /go/src/github.com/0xProject/0x-mesh
 
 ADD . ./
 
-RUN make deps-go-no-lockfile
+RUN make gobin
```
albrow: `gobin` is used for installing tools that we run from the command line (specifically `gowasmexec`, which is used in the tests). I don't actually think we need it here. We can just skip this step.

```suggestion
```

PR #642, opened 2020-01-13
rpc,typescript: Add Paginated getOrders method

fabioberger:
Fixes: #86

### DIFF
--- a/rpc/clients/typescript/src/ws_client.ts
+++ b/rpc/clients/typescript/src/ws_client.ts
@@ -337,6 +337,24 @@ export class WSClient {
         return getOrdersResponse;
     }
     /**
+     * Get page of 0x signed orders stored on the Mesh node at the specified snapshot
+     * @param page Page index at which to retrieve orders
+     * @param perPage number of signedOrders to fetch per paginated request
+     * @param snapshotID The DB snapshot at which to fetch orders. If omitted, a new snapshot is created
+     * @returns the snapshotID, snapshotTimestamp and all orders, their hashes and fillableTakerAssetAmounts
+     */
+    public async getOrdersForPageAsync(page: number, perPage: number = 200, snapshotID?: string): Promise<GetOrdersResponse> {
+        const finalSnapshotID = snapshotID === undefined ? '' : snapshotID;
+
+        const rawGetOrdersResponse: RawGetOrdersResponse = await this._wsProvider.send('mesh_getOrders', [
+            page,
+            perPage,
+            finalSnapshotID,
+        ]);
+        const getOrdersResponse = WSClient._convertRawGetOrdersResponse(rawGetOrdersResponse);
+        return getOrdersResponse;
+    }
+    /**
      * Subscribe to the 'orders' topic and receive order events from Mesh. This method returns a
      * subscriptionId that can be used to `unsubscribe()` from this subscription.
      * @param   cb   callback function where you'd like to get notified about order events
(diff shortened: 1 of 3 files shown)


rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -173,7 +173,7 @@ blockchainTests.resets('WSClient', env => {
                 // timestamp is approximately equal (within 1 second) because the server
                 // will receive the request slightly after it is sent.
                 const now = new Date(Date.now()).getTime();
-                const perPage = 10;
+                const perPage = ordersLength / 2;
```
fabioberger: This test wasn't actually making multiple paginated requests before so I modified it so that it does.

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -195,6 +195,55 @@ blockchainTests.resets('WSClient', env => {
             });
         });
 
+        describe('#getOrdersForPageAsync', async () => {
+            it('properly makes paginated request and returns signedOrders', async () => {
+                const ordersLength = 10;
+                const orders = [];
+                for (let i = 0; i < ordersLength; i++) {
+                    orders[i] = await orderFactory.newSignedOrderAsync({});
```
jalextowle: I love to see the monorepo's tools coming into mesh!

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -195,6 +195,55 @@ blockchainTests.resets('WSClient', env => {
...
+                page = 1;
+                response = await deployment.client.getOrdersForPageAsync(page, perPage, response.snapshotID);
+                expect(uuidValidate(response.snapshotID)).to.be.true();
+
+                // Combine orders found in first and second paginated requests
+                responseOrders = [...responseOrders, ...response.ordersInfos];
+
+                // Verify that all of the orders that were added to the mesh node
+                // were returned in the two `getOrders` rpc response
+                for (const order of orders) {
```
jalextowle: Maybe we should make this a function since this pattern is now used twice. Not a big deal tho

rpc/clients/typescript/src/ws_client.ts
```diff
@@ -336,6 +336,24 @@ export class WSClient {
...
         return getOrdersResponse;
     }
+    /**
+     * Get page of 0x signed orders stored on the Mesh node at the specified snapshot
+     * @param page Page index at which to retrieve orders
+     * @param perPage number of signedOrders to fetch per paginated request
+     * @param snapshotID The DB snapshot at which to fetch orders. If omitted, a new snapshot is created
+     * @returns the snapshotID, snapshotTimestamp and all orders, their hashes and fillableTakerAssetAmounts
+     */
+    public async getOrdersForPageAsync(page: number, perPage: number = 200, snapshotID?: string): Promise<GetOrdersResponse> {
```
albrow: @fabioberger can we use this new `getOrdersForPageAsync` in the implementation of `getOrdersAsync` to make the code DRYer?

rpc/clients/typescript/test/ws_client_test.ts
albrow: I also think we should do this.

rpc/clients/typescript/src/ws_client.ts
fabioberger: Great idea, done ✅

PR #643, opened 2020-01-13
Fix the intermitent failures on CI (part 2)

jalextowle:
Fixes two more issues that were causing intermittent CI failures.


rpc/clients/typescript/test/utils/ws_server.ts
```diff
@@ -44,10 +44,23 @@ export async function startServerAndClientAsync(): Promise<MeshDeployment> {
...
 
     const mesh = new MeshHarness();
-    const log = await mesh.waitForPatternAsync(/started RPC server/);
-    // TODO(jalextowle): Remove this log once the underlying issue has
-    // been found.
-    console.log(log.toString()); // tslint:disable-line
+    const startingPattern = /started RPC server/;
+    const patternChunk = await mesh.waitForPatternAsync(startingPattern);
+
+    // Since chunks can contain more than one log, process the chunks until a
```
albrow: @jalextowle I would not expect there to be more than one log that matches the pattern `/started RPC server/`. Can you explain why that would be the case?

jalextowle: The event listeners forward "chunks" rather than logs. Some of these "chunks" are really just strings of the format: 

```
log_containing_pattern
log_not_containing_pattern
log_not_containing_pattern
```

This is invalid json, which caused the test to fail with an error like `found unexpected { at position 196 in JSON`. The fix is to make sure that we are only parsing a single log.

All this said, I am going to move this logic to `waitForPatternAsync` so that it is more intuitive to use.

rpc/clients/typescript/test/utils/ws_server.ts
```diff
@@ -75,11 +73,14 @@ export class MeshHarness {
             throw new Error('mesh instance has already been killed');
         }
         return new Promise<string>((resolve, reject) => {
-            this._mesh.stderr.on('data', async data => {
-                if (pattern.test(data.toString())) {
-                    resolve(data.toString());
+            const stream = jsonstream.parse(true);
+            stream.on('data', data => {
```
albrow: Nice! This is much cleaner 😄

PR #648, opened 2020-01-15
Upgrade to Go 1.13

### LINKED ISSUE #480: Upgrade to Go 1.13
Not necessarily as straightforward as changing the version number. I was playing around with the latest version of Go, and it looks like it breaks our Wasm builds. Some of our tests might also be affected by [this change to the `testing` package](https://golang.org/doc/go1.13#testing).

albrow:
Fixes #480. This will help us include some small but important optimizations (especially in the browser) and keep up with the Go ecosystem. libp2p has already switched to 1.13.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     environment:
       BASH_ENV: ~/.nvm/nvm.sh
     docker:
-      - image: circleci/golang:1.12.13-browsers
+      - image: circleci/golang:1.13.4-browsers
       - image: 0xorg/ganache-cli:istanbul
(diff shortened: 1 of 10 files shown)


fabioberger:
Should we add a note to the CHANGELOG about this?

albrow:
> Should we add a note to the CHANGELOG about this?

My vote is no. It shouldn't affect users at all and I did update the __CONTRIBUTING.md__ guide in this PR.

albrow:
Let's wait to merge for now. Need to get confirmation about the intended usage of `this` in part of __wasm_exec.js__.

PR #650, opened 2020-01-15
Upgrade kad-dht to 0.3.0 to fix a regression

albrow:
In https://github.com/0xProject/0x-mesh/pull/539 we upgraded `github.com/libp2p/go-libp2p-kad-dht` to fix a memory leak. It looks like we regressed as part of the switch to Go Modules and an older version of the package was re-introduced. This PR fixes the problem by explicitly specifying version 0.3.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ This changelog is a work in progress and may contain notes for versions which ha
 - Add `getOrdersForPageAsync` method to `@0x/mesh-rpc-client` WS client interface so that clients can paginate through the retrieved orders themselves ([#642](https://github.com/0xProject/0x-mesh/pull/642)).
 
 
+## v8.1.1
+
+### Bug fixes 🐞
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,13 @@ This changelog is a work in progress and may contain notes for versions which ha
 - Add `getOrdersForPageAsync` method to `@0x/mesh-rpc-client` WS client interface so that clients can paginate through the retrieved orders themselves ([#642](https://github.com/0xProject/0x-mesh/pull/642)).
 
 
+## v8.1.1
+
+### Bug fixes 🐞
+
+- Fixed a regression which can reuslt in memory leaks. ([#650](https://github.com/0xProject/0x-mesh/pull/650)).
```
jalextowle: Nit: `reuslt` --> `result`.

Also, in the future it could be good to link to the bug report if a dependency introduces an issue like this.

Issue #577, opened 2019-12-05
Implement getOrders in the browser


albrow:
We have a `getOrders` RPC endpoint for getting all orders but we don't have an equivalent method in the `mesh-browser` package.

joeykrug:
Would it be possible to call getOrders with a certain order schema (passed as a parameter at the time of the call)? 

Example use case:

- User loads a market on augur directly, user would want to getOrders(order_format_for_that_market) first, then afterwards getOrders for the rest of the markets from mesh just using the order schema for augur orders (so not market specific ones)

albrow:
@joeykrug We're open to it, but `getOrders` in this context returns all the orders that are already stored locally. Since it doesn't involve sending requests or communicating with other nodes, there is probably not much benefit to optimizing it. We would rather focus on optimizations that will make a bigger impact, like improving the algorithm for sharing old orders (see https://github.com/0xProject/0x-mesh/issues/638).

albrow:
This was implemented in #655 and will be included in version 9.0.0.

Issue #576, opened 2019-12-05
Implement getStats in the browser


albrow:
We have an RPC endpoint called `getStats` but there is currently not a corresponding method in the `mesh-browser` package.

albrow:
This was implemented in #654 and will be included in version 9.0.0.

PR #653, opened 2020-01-16
Refactor types so that rpc is not imported in core or browser packages

albrow:
I was working on https://github.com/0xProject/0x-mesh/issues/577 and https://github.com/0xProject/0x-mesh/issues/576. It feels really awkward to import the `rpc` package into the `browser` package, not to mention it adds unnecessary bloat to the WebAssembly bundle. This PR re-arranges the types so that `core` and `browser` do not need to import `rpc`.

We can probably find more things to move into to the `common/types` package over time.


integration-tests/browser_integration_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !js
```
jalextowle: Did something change, or should this have always been here?

rpc/client.go
```diff
@@ -1,13 +1,14 @@
+// +build !js
```
jalextowle: Same question. It doesn't look like much changed besides using `types` instead of `rpc`.

rpc/service.go
```diff
@@ -1,3 +1,5 @@
+// +build !js
```
jalextowle: Ditto

integration-tests/browser_integration_test.go
albrow: This file never worked in the browser, (among other things, it uses `chromedp` and `os/exec`). But until this PR, it at least _built_ in the browser.

What changed was that I added `+build !js` to the `rpc` package to make sure that we don't include it in browser builds. As a consequence we also needed to add the tag here.

rpc/client.go
albrow: This change is intentional. The `rpc` package should not be included in browser builds. The fact that it was included in the past was an oversight. Adding this build tag makes it so that it cannot be included. Without it, it's fairly easy to accidentally include it via transitive dependencies.

jalextowle: Thanks for the answer, that makes sense 👍

integration-tests/browser_integration_test.go
jalextowle: Ah I see. That makes sense.

PR #655, opened 2020-01-17
Implement getOrdersAsync getOrdersForPageAsync for browsers

### LINKED ISSUE #577: Implement getOrders in the browser

albrow:
Fixes #577. This PR adds both `getOrdersAsync` and `getOrdersForPageAsync` to the `@0x/mesh-browser` package.

### DIFF
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -252,6 +252,30 @@ export interface Stats {
     ethRPCRateLimitExpiredRequests: number;
 }
 
+interface WrapperOrderInfo {
+    orderHash: string;
+    signedOrder: WrapperSignedOrder;
+    fillableTakerAssetAmount: string;
+}
+
+export interface OrderInfo {
+    orderHash: string;
+    signedOrder: SignedOrder;
+    fillableTakerAssetAmount: BigNumber;
+}
+
+interface WrapperGetOrdersResponse {
+    snapshotID: string;
+    snapshotTimestamp: number;
+    ordersInfos: WrapperOrderInfo[];
+}
+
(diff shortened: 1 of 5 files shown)


browser/ts/index.ts
```diff
@@ -761,6 +786,66 @@ export class Mesh {
...
+        let ordersInfos = getOrdersResponse.ordersInfos;
+
+        let allOrderInfos: OrderInfo[] = [];
+
+        do {
+            allOrderInfos = [...allOrderInfos, ...ordersInfos];
+            page++;
+            getOrdersResponse = await this.getOrdersForPageAsync(page, perPage, snapshotID);
+            ordersInfos = getOrdersResponse.ordersInfos;
+        } while (ordersInfos.length > 0);
```
jalextowle: Just out of curiosity, is there a reason why we're using a `do..while` loop here?

albrow: You'll have to ask @fabioberger. This is copied over from the TypeScript RPC client since we don't have any way to share code between them right now.

fabioberger: @jalextowle we need to do an initiate request and then feed the `snapshotID` returned into subsequent requests. Just made sense at the time I wrote it. Any issues with it?

jalextowle: Nope. I like `do..while` loops, but I almost never see them used. I just wondered if there was a particular reason that we used them. It sounds like they save a "bootstrapping" statement before a for or while loop.

Thanks for answering 😄

PR #663, opened 2020-01-22
zeroex,ethereum,meshdb: Add support for validating & watching Chai bridge orders

fabioberger:
Fixes: https://github.com/0xProject/0x-mesh/pull/663


ethereum/contract_addresses.go
```diff
@@ -82,6 +86,8 @@ var ChainIDToContractAddresses = map[int]ContractAddresses{
 		DevUtils:            common.HexToAddress("0x04674b391d3a58f4350e5306d5251a74e91f65da"),
 		WETH9:               common.HexToAddress("0xc778417e063141139fce010982780140aa0cd5ab"),
 		ZRXToken:            common.HexToAddress("0xff67881f8d12f372d91baae9752eb3631ff0ed00"),
+		ChaiBridge:          common.HexToAddress("0x0000000000000000000000000000000000000000"),
+		ChaiToken:           common.HexToAddress("0x0000000000000000000000000000000000000000"),
```
fabioberger: @abandeali1 do you know what the contract addresses are for these other testnets, including our Ganache snapshot?

meshdb/meshdb.go
```diff
@@ -509,12 +523,29 @@ func parseContractAddressesAndTokenIdsFromAssetData(assetData []byte) ([]singleA
...
+	case "ERC20Bridge":
+		var decodedAssetData zeroex.ERC20BridgeAssetData
+		err := assetDataDecoder.Decode(assetData, &decodedAssetData)
+		if err != nil {
+			return nil, err
+		}
+		tokenAddress := decodedAssetData.TokenAddress
+		// HACK(fabio): Despite Chai ERC20Bridge orders encoding the Dai address as
+		// the tokenAddress, we actually want to react to the Chai token's contract
+		// events, so we actually return it instead.
```
fabioberger: @abandeali1 can you confirm that this is the intended behavior?

ethereum/contract_addresses.go
abandeali1: I'm not aware of Chai being deployed on any of the testnets, so we also did not deploy the `ChaiBridge` contract. Is that a blocker at all?

meshdb/meshdb.go
abandeali1: Yep, the maker is actually holding Chai, not Dai, so we'll only need to watch the Chai contract's events.

zeroex/asset_data_decoder.go
```diff
@@ -23,10 +23,14 @@ const ERC1155AssetDataID = "a7cb5fb7"
...
+const ERC20BridgeAssetDataID = "dc1600f3"
+
 const erc20AssetDataAbi = "[{\"inputs\":[{\"name\":\"address\",\"type\":\"address\"}],\"name\":\"ERC20Token\",\"type\":\"function\"}]"
 const erc721AssetDataAbi = "[{\"inputs\":[{\"name\":\"address\",\"type\":\"address\"},{\"name\":\"tokenId\",\"type\":\"uint256\"}],\"name\":\"ERC721Token\",\"type\":\"function\"}]"
 const erc1155AssetDataAbi = "[{\"constant\":false,\"inputs\":[{\"name\":\"address\",\"type\":\"address\"},{\"name\":\"ids\",\"type\":\"uint256[]\"},{\"name\":\"values\",\"type\":\"uint256[]\"},{\"name\":\"callbackData\",\"type\":\"bytes\"}],\"name\":\"ERC1155Assets\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]"
 const multiAssetDataAbi = "[{\"inputs\":[{\"name\":\"amounts\",\"type\":\"uint256[]\"},{\"name\":\"nestedAssetData\",\"type\":\"bytes[]\"}],\"name\":\"MultiAsset\",\"type\":\"function\"}]"
+const erc20BuyerAssetDataAbi = "[{\"inputs\":[{\"name\":\"tokenAddress\",\"type\":\"address\"},{\"name\":\"bridgeAddress\",\"type\":\"address\"},{\"name\":\"calldata\",\"type\":\"bytes\"}],\"name\":\"ERC20Bridge\",\"type\":\"function\"}]"
```
mzhu25: nit: 
```suggestion
const erc20BridgeAssetDataAbi = "[{\"inputs\":[{\"name\":\"tokenAddress\",\"type\":\"address\"},{\"name\":\"bridgeAddress\",\"type\":\"address\"},{\"name\":\"calldata\",\"type\":\"bytes\"}],\"name\":\"ERC20Bridge\",\"type\":\"function\"}]"
```

zeroex/asset_data_decoder.go
```diff
@@ -79,7 +90,11 @@ func NewAssetDataDecoder() *AssetDataDecoder {
 	}
 	multiAssetDataABI, err := abi.JSON(strings.NewReader(multiAssetDataAbi))
 	if err != nil {
-		log.WithField("erc20AssetDataAbi", erc20AssetDataAbi).Panic("erc20AssetDataAbi should be ABI parsable")
+		log.WithField("multiAssetDataAbi", multiAssetDataAbi).Panic("multiAssetDataAbi should be ABI parsable")
+	}
+	erc20BuyerAssetDataABI, err := abi.JSON(strings.NewReader(erc20BuyerAssetDataAbi))
```
mzhu25: ditto

ethereum/contract_addresses.go
mzhu25: We could pretty easily deploy Chai and ChaiBridge on Kovan, since MCD is on there. The others might be a pain in the ass though...

zeroex/asset_data_decoder_test.go
```diff
@@ -77,3 +77,20 @@ func TestDecodeERC1155AssetData(t *testing.T) {
...
+	d := NewAssetDataDecoder()
+
+	var actualDecodedAssetData ERC20BridgeAssetData
+	err := d.Decode(assetData, &actualDecodedAssetData)
+	require.NoError(t, err)
+
+	expectedDecodedAssetData := ERC20BridgeAssetData{
+		TokenAddress:  common.HexToAddress("0x6b175474e89094c44da98b954eedeac495271d0f"),
+		BridgeAddress: common.HexToAddress("0x77c31eba23043b9a72d13470f3a3a311344d7438"),
+		Calldata:      common.Hex2Bytes("0x"),
```
jalextowle: We should test that `calldata` is decoded correctly. From the source code (https://github.com/0xProject/0x-monorepo/blob/development/contracts/asset-proxy/contracts/src/bridges/ChaiBridge.sol#L39), we can see that any `calldata` that is provided is valid. It's entirely possible that users will come up with novel uses of calldata, so I think that we need to test that we can decode it.

ethereum/contract_addresses.go
fabioberger: It's probably not a blocker, I'll just reject Chai Bridge orders on other testnets for now so that we don't accept Chai Bridge orders where the bridge address is set to the null address.

fabioberger: @mzhu25 could we at least deploy on our Ganache snapshot so we can write a test for this?

zeroex/asset_data_decoder_test.go
fabioberger: @jalextowle, it looks like the calldata is not used by the Chai Bridge Contract. How are we supposed to know how to decode it? My understanding is it's simply a byte-array that can include _any_ data. Is that incorrect?

ethereum/contract_addresses.go
fabioberger: Turns out this is difficult. Punting on this for now.

zeroex/asset_data_decoder_test.go
jalextowle: That’s correct. When I say “we should test that calldata is decoded correctly” I mean that we should decode it as bytes. The only change that is required is to add a new test where the bytes field isn’t empty.

fabioberger: Added a test for it.

PR #665, opened 2020-01-23
Add support for passing in a Web3 Provider in the browser

### LINKED ISSUE #581: Support passing in a web3 provider in the browser
Currently, when running Mesh in the browser, you need to provide an Ethereum RPC URL. This works great for some applications but isn't ideal for others. For example, it may expose your Infura/Alchemy API key and lead to abuse.

Most of the 0x TypeScript tooling supports [passing in a web3 provider](https://0x.org/docs/guides/web3-provider-explained), which can be provided by a wallet. We should support this use case when running Mesh in the browser.

The biggest challenge here will be passing the provider through to the Wasm bytecode and making it work correctly with `go-ethereum` and our generated Go contract wrappers.

albrow:
Fixes #581.

This PR adds the ability to use a Web3 Provider in the browser, which takes the place of `config.EthereumRPCURL` and will be used for all Ethereum RPC requests.

Most of the magic is in the new `providerwrapper` package. This package implements `ethclient.RPCClient` by wrapping a Web3 provider and calling `sendAsync`.

I also updated the integration tests to use a Web3 Provider. This is the bare minimum we can do for testing. After https://github.com/0xProject/0x-mesh/pull/649 is merged we should plan on adding more unit tests for a lot of the conversion code in this PR.

This PR depends on some changes to two separate dependencies:

1. For `go-ethereum`, I [refactored `ethclient.RPCClient` into an interface](https://github.com/0xProject/go-ethereum/commits/wasm_signer_core_3).
2. For `plaid/go-envvar`, I added [support for ignoring struct fields](https://github.com/plaid/go-envvar/pull/17).

I updated our __go.mod__ file to use our forks of these dependencies for now.

### DIFF
--- a/browser/go/providerwrapper/providerwrapper.go
+++ b/browser/go/providerwrapper/providerwrapper.go
@@ -48,6 +48,18 @@ func (c *RPCClient) CallContext(ctx context.Context, result interface{}, method 
 	//         jsonrpc: string;
 	//     }
 	//
+	//     interface JSONRPCResponseError {
+	//         message: string;
+	//         code: number;
+	//     }
+	//
+	//     interface JSONRPCResponsePayload {
+	//         result: any;
+	//         id: number;
+	//         jsonrpc: string;
+	//         error?: JSONRPCResponseError;
+	//     }
+	//
 	//     type JSONRPCErrorCallback = (err: Error | null, result?: JSONRPCResponsePayload) => void;
 	//
 	//     sendAsync(payload: JSONRPCRequestPayload, callback: JSONRPCErrorCallback): void;
@@ -107,7 +119,9 @@ func (c *RPCClient) CallContext(ctx context.Context, result interface{}, method 
 	case err := <-errChan:
 		return err
 	case jsResult := <-resultChan:
-		// TOOD(albrow): Handle jsResult.error?
+		if rpcErr := jsResult.Get("error"); !jsutil.IsNullOrUndefined(rpcErr) {
+			return jsErrorToRPCError(rpcErr)
+		}
 		if result == nil {
 			return nil
 		}
@@ -141,4 +155,30 @@ func (c *RPCClient) EthSubscribe(ctx context.Context, channel interface{}, args 
 // Close is a no-op in this implementation.
 func (c *RPCClient) Close() {
 	// no-op for now.
+}
+
+var _ rpc.Error = &rpcError{}
+
+type rpcError struct {
+	message string
+	code    int
+}
+
+func (e rpcError) Message() string {
+	return e.message
+}
+
+func (e rpcError) Error() string {
+	return fmt.Sprintf("Web3Provider returned RPC error (code %d): %s", e.code, e.message)
+}
+
+func (e rpcError) ErrorCode() int {
+	return e.code
+}
+
+func jsErrorToRPCError(jsError js.Value) rpc.Error {
+	return &rpcError{
+		message: jsError.Get("message").String(),
+		code:    jsError.Get("code").Int(),
+	}
 }
--- a/browser/go/providerwrapper/providerwrapper.go
+++ b/browser/go/providerwrapper/providerwrapper.go
@@ -15,6 +15,7 @@ import (
 	"github.com/0xProject/0x-mesh/browser/go/jsutil"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/rpc"
+	log "github.com/sirupsen/logrus"
 )
 
 var _ ethclient.RPCClient = &RPCClient{}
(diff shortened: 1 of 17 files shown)


browser/ts/index.ts
```diff
@@ -199,6 +201,9 @@ export interface Config {
     // all the required fields) are automatically included. For more information
     // on JSON Schemas, see https://json-schema.org/
     customOrderFilter?: JsonSchema;
+    // Offers the ability to use your own web3 provider for all Ethereum RPC
+    // requests instead of the default.
+    web3Provider?: Provider;
```
albrow: @fabioberger Is `Provider ` the right type to use here or should we be accepting a more general type like `SupportedProvider` or `ZeroExProvider`? The main thing to keep in mind is that we need a consistent interface that our Go code can call out to. Maybe there's a way to convert to a standard interface?

browser/go/providerwrapper/providerwrapper.go
```diff
@@ -0,0 +1,183 @@
...
+// BatchCall sends all given requests as a single batch and waits for the server
+// to return a response for all of them. The wait duration is bounded by the
+// context's deadline.
+//
+// In contrast to CallContext, BatchCallContext only returns errors that have occurred
+// while sending the request. Any error specific to a request is reported through the
+// Error field of the corresponding BatchElem.
+//
+// Note that batch calls may not be executed atomically on the server side.
+func (c *RPCClient) BatchCallContext(ctx context.Context, b []rpc.BatchElem) error {
```
albrow: I didn't implement `BatchCallContext` or `EthSubscribe` since they don't seem to actually be used in the browser. We can implement them later if needed.

browser/go/providerwrapper/providerwrapper.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (e rpcError) Error() string {
+	return fmt.Sprintf("Web3Provider returned RPC error (code %d): %s", e.code, e.message)
+}
+
+func (e rpcError) ErrorCode() int {
+	return e.code
+}
+
+func jsErrorToRPCError(jsError js.Value) rpc.Error {
```
albrow: @fabioberger do we ever check the message, code, or type of error in block watcher or order watcher? Want to make sure that this is compatible with the rest of our code.

browser/go/providerwrapper/providerwrapper.go
fabioberger: Sounds good. We might use `EthSubscribe` is we decide to use a WS subscription for watching new blocks, but we don't use it now.

browser/go/providerwrapper/providerwrapper.go
fabioberger: We do not, but we do include them in logs sometimes. Why would they not be compatible?

browser/ts/index.ts
fabioberger: Can we set it to `SupportedProvider` and then call `@0x/utils`'s [`providerUtils.standardizeOrThrow()`](https://github.com/0xProject/0x-monorepo/blob/development/packages/utils/src/provider_utils.ts#L33) method? That'll spit out a `ZeroExProvider` that we can rely on to always have a consistent interface. Unfortunately there was never a formal standard for what a "provider" looks like and so there are _many_ implementations floating around.

browser/go/providerwrapper/providerwrapper.go
albrow: Can you point me to the code where we log these?

examples/browser/src/index.ts
```diff
@@ -1,10 +1,11 @@
-import { Mesh, OrderEvent, SignedOrder, BigNumber } from '@0x/mesh-browser';
+import { Mesh, OrderEvent, SignedOrder, BigNumber, Provider } from '@0x/mesh-browser';
 
 (async () => {
     // Configure Mesh to use mainnet and Infura.
```
jalextowle: Nit: This comment needs to be updated

examples/browser/src/index.ts
```diff
@@ -26,24 +27,24 @@ import { Mesh, OrderEvent, SignedOrder, BigNumber } from '@0x/mesh-browser';
     // This order is for demonstration purposes only and is invalid. It will be
     // rejected by Mesh. You can replace it with a valid order.
     const order: SignedOrder = {
-        chainId: 1,
-        makerAddress: '0xa3eCE5D5B6319Fa785EfC10D3112769a46C6E149',
-        makerAssetData: '0xf47261b0000000000000000000000000e41d2489571d322189246dafa5ebde1f4699f498',
-        makerAssetAmount: new BigNumber('1000000000000000000'),
-        makerFee: new BigNumber('0'),
-        makerFeeAssetData: '0x',
+        signature:
```
jalextowle: Is there a particular order that is being used here?

albrow: Yeah I decided to just grab one from api.0x.org. The old one was invalid because it didn't have the correct signature.

browser/go/providerwrapper/providerwrapper.go
fabioberger: The `go-ethereum` `rpc.Client` checks the JSON-RPC response and if it contains the `Error` key, it returns it's contents as the error from the call. This error object gets logged as in: https://github.com/0xProject/0x-mesh/blob/master/ethereum/blockwatch/block_watcher.go#L174

albrow: 👍 I found the concrete implementation in go-ethereum and made our implementation more closely match it. I think we should be good.

PR #671, opened 2020-01-24
Update DevUtils addresses again

jalextowle:
Updated the DevUtils address again.


zeroex/ordervalidator/order_validator.go
```diff
@@ -357,6 +357,7 @@ func (o *OrderValidator) BatchValidate(ctx context.Context, rawSignedOrders []*z
 						log.WithFields(log.Fields{
 							"error":     err.Error(),
 							"numOrders": len(trimmedOrders),
+							"orders":    trimmedOrders,
```
albrow: @jalextowle [There were about 1,600 occurrences](https://kibana.mesh.0x.org/goto/27e67dfdaf42866e3131acbcd8359ad8) of this warning over the past 24 hours. Is this going to result in too much data in the logs?

fabioberger: Yeah, let's remove this. This warning occurs a lot. Anything that causes the RPC request to fail, including network issues will trigger it. Would orderHashes suffice? If it's necessary for debugging, can we be more judicious by introspecting the returned error and only throw if it's a smart contract revert reason?

jalextowle: No, order hashes don't help. The only thing that is useful is the contents of the order. I can do the introspection instead.

zeroex/ordervalidator/order_validator.go
```diff
@@ -354,10 +355,24 @@ func (o *OrderValidator) BatchValidate(ctx context.Context, rawSignedOrders []*z
 					d := b.Duration()
 					if d == maxDuration {
 						<-semaphoreChan
-						log.WithFields(log.Fields{
-							"error":     err.Error(),
-							"numOrders": len(trimmedOrders),
-						}).Warning("Gave up on GetOrderRelevantStates request after backoff limit reached")
+						var fields log.Fields
+						match, err := regexp.MatchString("abi: improperly formatted output", err.Error())
```
albrow: This shadows the definition for `err`, which will break the line of code below `if err.Error() == "VM execution error."`. How about renaming it to `regexErr`?

```suggestion
						match, regexErr := regexp.MatchString("abi: improperly formatted output", err.Error())
```

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This changelog is a work in progress and may contain notes for versions which ha
 - Added `getOrdersAsync` and `getOrdersForPageAsync` to the `@0x/mesh-browser` package. ([#655](https://github.com/0xProject/0x-mesh/pull/655)).
 - Added support for passing in your own Web3 provider when using the `@0x/mesh-browser` package. ([#665](https://github.com/0xProject/0x-mesh/pull/665)).
 
+### Bug fixes 🐞
```
albrow: This probably needs to be added to new section for version 8.1.3.

PR #673, opened 2020-01-24
Backport Release 7.1.0-beta

fabioberger:
This backports a bunch of recent fixes needed by TokenTrove who is still running a 0x V2 node alongside their 0x V3 node. See CHANGELOG entry for full list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
+## v7.1.0-beta
+
+### Features ✅
+
+ - Reduce startup time for Mesh node by only waiting for a block to be processed if Mesh isn't already sync'ed up to the latest block. ([#622](https://github.com/0xProject/0x-mesh/pull/622))
+ - Developers can now override the contract addresses for any testnet using the `CUSTOM_CONTRACT_ADDRESSES` env config ([#640](https://github.com/0xProject/0x-mesh/pull/640)).
+  - Add `getOrdersForPageAsync` method to `@0x/mesh-rpc-client` WS client interface so that clients can paginate through the retrieved orders themselves ([#642](https://github.com/0xProject/0x-mesh/pull/642)).
+
+### Bug fixes 🐞
+
+ - Fixed a bug where we attempted to update the same order multiple times in a single DB txn, causing the later update to noop. ([#623](https://github.com/0xProject/0x-mesh/pull/623)).
+- Fixed a bug which could cause Mesh to exit if a re-org condition occurs causing a block to be added and removed within the same block sync operation. ([#614](https://github.com/0xProject/0x-mesh/pull/614)).
+ - Fix bug where we attempted to update the same order multiple times in a single DB txn, causing the later update to noop. ([#623](https://github.com/0xProject/0x-mesh/pull/623))
+
 ## v7.0.0-beta
 
 ### Breaking changes 🛠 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+
+### Features ✅
+
+ - Reduce startup time for Mesh node by only waiting for a block to be processed if Mesh isn't already sync'ed up to the latest block. ([#622](https://github.com/0xProject/0x-mesh/pull/622))
+ - Developers can now override the contract addresses for any testnet using the `CUSTOM_CONTRACT_ADDRESSES` env config ([#640](https://github.com/0xProject/0x-mesh/pull/640)).
+  - Add `getOrdersForPageAsync` method to `@0x/mesh-rpc-client` WS client interface so that clients can paginate through the retrieved orders themselves ([#642](https://github.com/0xProject/0x-mesh/pull/642)).
+
+### Bug fixes 🐞
+
+ - Fixed a bug where we attempted to update the same order multiple times in a single DB txn, causing the later update to noop. ([#623](https://github.com/0xProject/0x-mesh/pull/623)).
```
jalextowle: Should we backport the `DevUtils` fixes? It looks like we updated the address of mainnet DevUtils, so we should probably add a bullet to the `Bug fixes` section.

rpc/clients/typescript/src/ws_client.ts
```diff
@@ -287,36 +287,50 @@ export class WSClient {
     /**
      * Get all 0x signed orders currently stored in the Mesh node
```
jalextowle: Can we cherry-pick the updates to the ws client tests onto this branch? It seems like we should make sure our best tests get brought onto our release branches.

CHANGELOG.md
albrow: The updates to `DevUtils` only work with protocol v3 right? If we wanted to do this we would need to also deploy a new version of `DevUtils` for protocol v2.

jalextowle: We still updated `DevUtils`, so I think we should add a bullet to the changelog.

ethereum/contract_addresses.go
```diff
@@ -103,7 +103,7 @@ var ChainIDToContractAddresses = map[int]ContractAddresses{
 		Exchange:            common.HexToAddress("0x30589010550762d2f0d06f650d8e8b6ade6dbf4b"),
 		Coordinator:         common.HexToAddress("0x2ba02e03ee0029311e0f43715307870a3e701b53"),
 		CoordinatorRegistry: common.HexToAddress("0x09fb99968c016a3ff537bf58fb3d9fe55a7975d5"),
-		DevUtils:            common.HexToAddress("0x1e3616bc5144362f95d72de41874395567697e93"),
+		DevUtils:            common.HexToAddress("0x80101f90446d9e19a80109ac2767c8c957debb29"),
```
albrow: Did we meant to do this? Judging by our cheat sheet [this address is a version of DevUtils is designed for protocol v3](https://0x.org/docs/guides/0x-cheat-sheet).

CHANGELOG.md
albrow: Hmm.. I think that might have been a mistake. Made a separate comment about it.

ethereum/contract_addresses.go
fabioberger: This was an oversight. Fixing.

CHANGELOG.md
fabioberger: It was a mistake. Fixed it.

rpc/clients/typescript/src/ws_client.ts
fabioberger: Unfortunately those tests use features in the V3 0x libraries that don't exist in V2 and cherry-picking them would involve re-writing a bunch of the test setup. We agreed offline to leave this as is.

PR #683, opened 2020-01-29
orderwatch: Batch process multiple block event arrays at once

fabioberger:
I discovered this inefficiency while bench-marking the performance of the OrderWatcher. It is  possible that multiple blockEvents are queued up within the `blockEventsChan` and we _could_ process them all at once but weren't. It is possible that the events in these array would lead to the same order being re-validated, and so combining them leads to fewer RPC calls and lowers the latency after which we emit relevant order events.

### DIFF
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -278,8 +278,20 @@ func (w *Watcher) mainLoop(ctx context.Context) error {
 				"error": err.Error(),
 			}).Error("block subscription error encountered")
 		case events := <-w.blockEventsChan:
+			// Instead of simply processing the first array of events in the blockEventsChan,
+			// we might as well process _all_ events in the channel.
+			allEvents := events
+		L:
+			for {
+				select {
+				case moreEvents := <-w.blockEventsChan:
+					allEvents = append(allEvents, moreEvents...)
+				default:
+					break L
+				}
+			}
 			w.handleBlockEventsMu.Lock()
-			if err := w.handleBlockEvents(ctx, events); err != nil {
+			if err := w.handleBlockEvents(ctx, allEvents); err != nil {
 				w.handleBlockEventsMu.Unlock()
 				return err
 			}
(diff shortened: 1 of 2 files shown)


fabioberger:
Not sure whether this deserves a CHANGELOG entry. Any thoughts?

zeroex/orderwatch/order_watcher.go
```diff
@@ -288,6 +292,20 @@ func (w *Watcher) mainLoop(ctx context.Context) error {
...
 
+func drainBlockEventsChan(blockEventsChan chan []*blockwatch.Event) []*blockwatch.Event {
+	allEvents := []*blockwatch.Event{}
+L:
+	for {
+		select {
+		case moreEvents := <-blockEventsChan:
+			allEvents = append(allEvents, moreEvents...)
+		default:
+			break L
```
jalextowle: Why is the label needed here? Does break only break out of the `select`?

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -1449,6 +1449,41 @@ func TestConvertValidationResultsIntoOrderEventsUnexpired(t *testing.T) {
 	assert.Equal(t, false, orderTwo.IsRemoved)
 }
 
+func TestDrainAllBlockEventsChan(t *testing.T) {
```
jalextowle: Thanks for writing this test 🙏

zeroex/orderwatch/order_watcher.go
albrow: Yes, without this label, `break` would only break out of `select`. @fabio maybe we could rename the label to `Loop` or `OuterLoop`?

zeroex/orderwatch/order_watcher.go
```diff
@@ -288,6 +292,20 @@ func (w *Watcher) mainLoop(ctx context.Context) error {
 	}
 }
 
+func drainBlockEventsChan(blockEventsChan chan []*blockwatch.Event) []*blockwatch.Event {
+	allEvents := []*blockwatch.Event{}
+L:
+	for {
+		select {
+		case moreEvents := <-blockEventsChan:
+			allEvents = append(allEvents, moreEvents...)
```
albrow: Can we add a maximum size for `allEvents` here? Infinite loops make me nervous. There's a slight chance that new events could come in before we have the time to process them, especially on testnets with a higher rate of mining new blocks or when CPU usage is high. It's not something that strikes me as super likely, but the fix is also really easy.

```suggestion
			allEvents = append(allEvents, moreEvents...)
			if len(allEvents) >= maxBlockEventsToHandle {
			    break L
			}
```

albrow:
> Not sure whether this deserves a CHANGELOG entry. Any thoughts?

I don't have a strong preference.

PR #692, opened 2020-01-30
Improve algorithm for sharing old orders

### LINKED ISSUE #638: Improve algorithm for sharing old orders
Our current algorithm for sharing old orders is functional but has a few big inefficiencies. It wastes a lot of bandwidth sharing orders with peers who don't need them (because they already have them). This has a direct financial impact on people who want to run Mesh nodes, as the bandwidth usage is a driver of high cost on hosting providers like AWS (though notably it is a lot cheaper on Digital Ocean). It is also having a significant negative impact on CPU performance in the browser.

In general, we need to change the algorithm so that we only send orders to peers that actually need them, instead of broadcasting old orders indiscriminately. @fabioberger suggested looking at https://github.com/ipfs/go-bitswap as one solution. We might also need to design our own protocol. This new solution will likely exist outside of GossipSub.

albrow:
Fixes #638.

This PR introduces a new protocol for sharing existing orders between two peers, called the "ordersync" protocol. At a high-level, ordersync is a pull-based protocol rather than a push-based one. Upon initialization, a Mesh node will simply request orders from its peers. This PR also removes our old push-based strategy for sharing existing orders. GossipSub is now only used for sharing new orders.

The ordersync protocol is designed with forward-compatibility in mind and supports multiple "subprotocols" which define the details for what each request and response should look like. This PR includes only one subprotocol, in which all orders are returned using pagination. We can expand on this in the future to implement more efficient protocols (e.g. using cuckoo filters, a gossip layer, or prioritizing sharing orders with the best price).

Remaining TODOs:

- [x] Modify peer scores based on what happens during the ordersync protocol.
- [x] Add basic rate limiting to prevent abuse.
- [x] Add support for filters. ordersync providers should only give you orders that match your filter, but they currently return all orders.


core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// FilteredPaginationResponseMetadata is the response metadata for the
+// FilteredPaginationSubProtocol. It keeps track of the current page and SnapshotID.
+type FilteredPaginationResponseMetadata struct {
+	Page       int    `json:"page"`
+	SnapshotID string `json:"snapshotID"`
+}
+
+// Name reutrns the name of the FilteredPaginationSubProtocol
```
fabioberger:
```suggestion
// Name returns the name of the FilteredPaginationSubProtocol
```

core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,129 @@
...
+func (p *FilteredPaginationSubProtocol) Name() string {
+	return "/pagination-with-filter/version/0"
+}
+
+// GetOrders returns the orders for one page, based on the page number
+// and snapshotID in the corresponding to the given request. This is
```
fabioberger:
```suggestion
// and snapshotID corresponding to the given request. This is
```

core/core.go
```diff
@@ -580,11 +594,31 @@ func (app *App) Start(ctx context.Context) error {
 		return err
 	}
 
+	// Register and start ordersync service.
+	ordersyncSubprotocols := []ordersync.Subprotocol{
+		NewFilteredPaginationSubprotocol(app, paginationSubprotocolPerPage),
+	}
+	app.ordersyncService = ordersync.New(innerCtx, app.node, ordersyncSubprotocols)
```
jalextowle: I think the fact that we are calling `New` in a separate thread than we are calling `GetOrders` makes the case that I describe in the other file (it is the comment that describes an example heuristic) more likely. It's entirely possible that this peer will be hit with a `GetOrders` request from a new peer before it has actually finished bootstrapping it's database using the `ordersync` protocol.

I think that an even more impactful change than the heuristic I described would be for `New` to take a `done` channel that will be sent a signal whenever the `GetOrders` function has finished executing. Prior to receiving the signal, `HandleStream` would reject with a "Database not initialized" warning.

Thoughts?

core/ordersync/ordersync.go
```diff
@@ -0,0 +1,501 @@
...
+			}
+
+			if err := s.getOrdersFromPeer(ctx, peerID); err != nil {
+				log.WithFields(log.Fields{
+					"error":    err.Error(),
+					"provider": peerID.Pretty(),
+				}).Warn("could not get orders from peer via ordersync")
+				continue
+			} else {
+				// TODO(albrow): Handle case where no orders were returned from this
```
jalextowle: In line with this, should there be a threshold under which `ordersync` isn't considered a success that is nonzero? My reasoning for including this is that it's possible that a peer has significantly less orders than other peers, and it would make sense to consider that step of `ordersync` a failure. I'm curious to hear whether you think this is necessary considering that we are sampling from multiple peers.

An example heuristic would be "if one of the results of getting orders from a peer contained < 1% of the orders of the other ordersync peers, repeat." This heuristic would necessarily rely on backtracking but would prevent against calling ordersync on a newish peer.

core/ordersync/ordersync.go
```diff
@@ -0,0 +1,501 @@
...
+	case <-ctx.Done():
+		log.WithFields(log.Fields{
+			"error":    ctx.Err(),
+			"provider": stream.Conn().RemotePeer().Pretty(),
+		}).Warn("timed out waiting for ordersync response")
+		return nil, ctx.Err()
+	case rawRes := <-resChan:
+		return rawRes, nil
+	}
+}
```
jalextowle: Optional Suggestion: I think it's possible to de-dupe at least some of the code in these implementations. It would be easier with generics, but I think it could still be a bit cleaner.

core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,129 @@
...
+// the implementation for the "provider" side of the subprotocol.
+func (p *FilteredPaginationSubProtocol) GetOrders(ctx context.Context, req *ordersync.Request) (*ordersync.Response, error) {
```
jalextowle: Would it make more sense to call this something like `getAndSendOrders`? I think it could be a little confusing that `HandleStream` is on the `provider` side and `GetOrders` is on the `requester` side in the `ordersync` protocol, but in this subprotocol, these names switch places.

core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,129 @@
...
+			Page:       metadata.Page,
+			SnapshotID: ordersResp.SnapshotID,
+		},
+	}, nil
+}
+
+// HandleOrders handles the orders for one page by validating them, storing them in the database,
+// and firing the appropriate events. It also returns the next request to be sent. This is the
+// implementation for the "requester" side of the subprotocol.
+func (p *FilteredPaginationSubProtocol) HandleOrders(ctx context.Context, res *ordersync.Response) (*ordersync.Request, error) {
```
jalextowle: Same as above, consider switching the name of this. Maybe names like `HandleOrderSyncRequest` and `HandleOrderSyncResponse`?

core/core_test.go
```diff
@@ -28,3 +49,151 @@ func TestEthereumChainDetection(t *testing.T) {
...
+
+	// Connect the two nodes *after* adding orders to one of them. This should
+	// trigger the ordersync protocol.
+	err = originalNode.AddPeer(peer.AddrInfo{
+		ID:    newNode.node.ID(),
+		Addrs: newNode.node.Multiaddrs(),
+	})
+	require.NoError(t, err)
+
+	// Wait for newNode to get the orders via ordersync.
```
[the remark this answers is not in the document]
albrow: Since we added the order directly via `ValidateAndStoreValidOrders`, we are bypassing GossipSub. I think it is not possible for the order to be shared via GossipSub in this test, but let me know if I'm missing something.

core/ordersync/ordersync.go
```diff
@@ -0,0 +1,501 @@
...
+			if err := s.getOrdersFromPeer(ctx, peerID); err != nil {
+				log.WithFields(log.Fields{
+					"error":    err.Error(),
+					"provider": peerID.Pretty(),
+				}).Warn("could not get orders from peer via ordersync")
+				continue
+			} else {
+				// TODO(albrow): Handle case where no orders were returned from this
+				// peer. We need to not try them again, but also not count them toward
+				// the number of peers we have successfully synced with.
```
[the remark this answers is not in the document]
albrow: It should not be possible with the pagination subprotocol. This is mostly for the sake of forwards compatibility. E.g., we could have a subprotocol where the requester asks for orders with a hash between the range of X and Y, but the provider doesn't have any orders in that range.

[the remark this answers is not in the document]
albrow: This comment might need some adjusting since it doesn't really clearly cover the case I described above.

Makefile
```diff
@@ -40,10 +40,12 @@ test-all: test-go test-wasm-node test-wasm-browser
 .PHONY: test-go
 test-go: test-go-parallel test-go-serial test-browser-conversion
 
+
```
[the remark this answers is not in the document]
albrow: Yeah this is a carryover from conventions we used at a previous job. I like to put two spaces between Make tasks. It's subjective and probably not worth enforcing. Do you think I should just undo this commit and not make any formatting-only changes to the Makefile?

integration-tests/browser/src/index.ts
```diff
@@ -117,3 +123,7 @@ provider.start();
         console.error(err.toString());
     }
 });
+
+async function sleepAsync(ms: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
```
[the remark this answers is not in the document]
albrow: No VS Code does not remove new lines at the end of files. It is generally considered best practice to always include an explicit new line (partly historical, partly for compatibility with certain tools). This is likely a result of me switching computers several times and not copying over my editor config. I'll add it back.

core/core_test.go
```diff
@@ -28,3 +49,151 @@ func TestEthereumChainDetection(t *testing.T) {
...
+	for i := range originalOrders {
+		originalOrders[i] = scenario.CreateWETHForZRXSignedTestOrder(t, ethClient, constants.GanacheAccount1, constants.GanacheAccount2, big.NewInt(20), big.NewInt(5))
+	}
+
+	// We have to wait for latest block to be processed by the Mesh node.
+	time.Sleep(blockProcessingWaitTime)
+
+	results, err := originalNode.orderWatcher.ValidateAndStoreValidOrders(ctx, originalOrders, true, constants.TestChainID)
+	require.NoError(t, err)
+	require.NotEmpty(t, results.Accepted, "tried to add orders but some were invalid: \n%s\n", spew.Sdump(results))
```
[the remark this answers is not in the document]
albrow: This is just written backwards. We want the error message to be the same but the assertion should be `require.Empty(t, results.Rejected, ....)`. I'll fix it.

core/core_test.go
```diff
@@ -28,3 +49,151 @@ func TestEthereumChainDetection(t *testing.T) {
...
+		actualOrder := dbOrder.SignedOrder
+		assert.Equal(t, expectedOrder, actualOrder, "correct order was not stored in new node database")
+	}
+
+	// Wait for nodes to exit without error.
+	cancel()
+	wg.Wait()
+}
+
+func setupSubTest(t *testing.T) func(t *testing.T) {
```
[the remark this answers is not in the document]
albrow: Good point. Something we should consider in a separate PR.

core/ordersync/ordersync.go
albrow: Can you offer a concrete suggestion? It's pretty hard to do without generics. We need to create a variable with a concrete type (either `rawResponse` or `rawResult`) in order for encoding/decoding to work correctly.

core/ordersync/ordersync.go
albrow: I think it is too difficult to come up with a heuristic like this that is accurate, doesn't have any side effects, and is not gameable. IMO it's better to just request orders from more peers to get a better sampling. It's true that some of those peers might also be "new" but statistically it is not very likely that all of them will be. I think I'm going to go ahead and increase the default `minPeers` for now and leave it at that.

We can address this with better subprotocols in the future. The current implementation is more of a stopgap which greatly improves he previous algorithm but has some known flaws. This is one of them.

core/core.go
albrow: The downside to this approach is that for networks or topics that don't have `minPeers` peers, the ordersync protocol would fail and no one would be able to get any existing orders. I think it's better to leave the implementation as is. It is not very statistically likely that every peer you dial will be "new", especially if we increase `minPeers`.

core/ordersync_subprotocols.go
albrow: I like this suggestion and just applied it. Would prefer to keep `HandleStream` the same since it is a carryover from libp2p semantics. `ordersync.Service` is a stream handler, and other stream handlers use this naming convention in libp2p.

core/ordersync_subprotocols.go
albrow: Changed the method names in the `ordersync.Subprotocol` interface. What do you think of the current version?

Makefile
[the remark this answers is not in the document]
jalextowle: Nope, I'm totally fine with this change. I just wanted to make sure that I understood it properly.

core/ordersync/ordersync.go
jalextowle: That's reasonable. I think it's possible that we could come up with a working heuristic in future subprotocols if necessary, but I agree that it doesn't seem necessary given a greater amount of sampling.

core/core.go
jalextowle: That's a good point. I'm still somewhat concerned about a future in which there are a lot of browser nodes that are spinning up and down quickly and are being used as `ordersync` providers, but I think we can cross that bridge when we come to it considering the downsides that you mention.

core/ordersync/ordersync.go
jalextowle: We could put the log statements into functions. After looking again, that's the only change that can really be made, so it's totally up to you.

core/ordersync_subprotocols.go
jalextowle: It looks good!

core/core_test.go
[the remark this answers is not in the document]
fabioberger: Ah yep, that makes sense!

core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,129 @@
...
+func (p *FilteredPaginationSubProtocol) HandleOrderSyncResponse(ctx context.Context, res *ordersync.Response) (*ordersync.Request, error) {
+	if res.Metadata == nil {
+		return nil, errors.New("FilteredPaginationSubProtocol received response with nil metadata")
+	}
+	metadata, ok := res.Metadata.(*FilteredPaginationResponseMetadata)
+	if !ok {
+		return nil, fmt.Errorf("FilteredPaginationSubProtocol received response with wrong metadata type (got %T)", res.Metadata)
+	}
+	// TODO(albrow): Check that this order matches our current filter/topic
+	_, err := p.app.orderWatcher.ValidateAndStoreValidOrders(ctx, res.Orders, false, p.app.chainID)
```
fabioberger: @albrow in order for our order propogation dashboard to continue to make sense, we need to emit similar logs here for orders discovered and stored via `ordersync`. Relevant code-snippet from `message_handler.go`:

```
validationResults, err := app.orderWatcher.ValidateAndStoreValidOrders(ctx, orders, false, app.chainID)
if err != nil {
	return err
}

	// If the order isn't new, we don't log it's receipt or adjust peer scores
	if !acceptedOrderInfo.IsNew {
		continue
	}
	msg := orderHashToMessage[acceptedOrderInfo.OrderHash]
	// If we've reached this point, the message is valid, we were able to
	// decode it into an order and check that this order is valid. Update
	// peer scores accordingly.
	log.WithFields(map[string]interface{}{
		"orderHash": acceptedOrderInfo.OrderHash.Hex(),
		"from":      msg.From.String(),
	}).Info("received new valid order from peer")
	log.WithFields(map[string]interface{}{
		"order":     acceptedOrderInfo.SignedOrder,
		"orderHash": acceptedOrderInfo.OrderHash.Hex(),
		"from":      msg.From.String(),
	}).Trace("all fields for new valid order received from peer")
}
```

albrow: @fabioberger thanks for the tip. I just added a similar log message in this PR.

albrow:
@fabioberger @jalextowle I believe I addressed all your feedback and any remaining TODOs on my end. I want to leave some optimizations and improvements for a future PR. This one is ready for final review.

core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,174 @@
...
+	for _, order := range res.Orders {
+		if matches, err := p.orderFilter.MatchOrder(order); err != nil {
+			return nil, err
+		} else if matches {
+			filteredOrders = append(filteredOrders, order)
+		} else if !matches {
+			p.app.handlePeerScoreEvent(res.ProviderID, psReceivedOrderDoesNotMatchFilter)
+		}
+	}
+	validationResults, err := p.app.orderWatcher.ValidateAndStoreValidOrders(ctx, res.Orders, false, p.app.chainID)
```
fabioberger:
```suggestion
	validationResults, err := p.app.orderWatcher.ValidateAndStoreValidOrders(ctx, filteredOrders, false, p.app.chainID)
```

core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, acceptedOrderInfo := range validationResults.Accepted {
+		if acceptedOrderInfo.IsNew {
+			log.WithFields(map[string]interface{}{
+				"orderHash": acceptedOrderInfo.OrderHash.Hex(),
+				"from":      res.ProviderID.Pretty(),
+			}).Info("received new valid order from peer via ordersync")
```
fabioberger:
```suggestion
			}).Info("received new valid order from peer")
```

Let's not modify the log message since the data pipeline is querying for it exactly. If we want to distinguish between those discovered via `ordersync` and `gossipsub`, let's add a `source` field to the log.

core/ordersync_subprotocols.go
```diff
@@ -99,7 +120,16 @@ func (p *FilteredPaginationSubProtocol) HandleOrderSyncResponse(ctx context.Cont
...
+	for _, order := range res.Orders {
+		if matches, err := p.orderFilter.MatchOrder(order); err != nil {
+			return nil, err
+		} else if matches {
+			filteredOrders = append(filteredOrders, order)
+		} else if !matches {
+			p.app.handlePeerScoreEvent(res.ProviderID, psReceivedOrderDoesNotMatchFilter)
+		}
+	}
 	_, err := p.app.orderWatcher.ValidateAndStoreValidOrders(ctx, res.Orders, false, p.app.chainID)
```
fabioberger:
```suggestion
	_, err := p.app.orderWatcher.ValidateAndStoreValidOrders(ctx, filteredOrders, false, p.app.chainID)
```

core/ordersync/ordersync.go
```diff
@@ -0,0 +1,507 @@
...
+	subprotocols map[string]Subprotocol
+	// requestRateLimiter is a rate limiter for incoming ordersync requests. It's
+	// shared between all peers.
+	requestRateLimiter *rate.Limiter
+}
+
+// SupportedSubprotocols returns the subprotocols that are supported by the service.
+func (s *Service) SupportedSubprotocols() []string {
+	sids := []string{}
+	for sid := range s.subprotocols {
```
jalextowle: This snippet suggests that this array can have a random order (unless I'm missing something): https://play.golang.org/p/s3WTo2LVuyl. With this in mind, we aren't actually enforcing a preference on these subprotocols. What do you think about keeping the original subprotocols array in the `Service` object?

core/ordersync/ordersync.go
```diff
@@ -0,0 +1,507 @@
...
+	ctx, cancel := context.WithTimeout(parentCtx, requestResponseTimeout)
+	defer cancel()
+	reqChan := make(chan *rawRequest, 1)
+	go func() {
+		var rawReq rawRequest
+		if err := json.NewDecoder(stream).Decode(&rawReq); err != nil {
+			log.WithFields(log.Fields{
+				"error":     err.Error(),
+				"requester": stream.Conn().RemotePeer().Pretty(),
+			}).Warn("could not encode ordersync request")
```
jalextowle: Doesn't `waitForRequest` want to return this error instead of sending it in the request channel? What am I missing?

core/ordersync/ordersync.go
```diff
@@ -0,0 +1,507 @@
...
+	}
+}
+
+func waitForResponse(parentCtx context.Context, stream network.Stream) (*rawResponse, error) {
+	ctx, cancel := context.WithTimeout(parentCtx, requestResponseTimeout)
+	defer cancel()
+	resChan := make(chan *rawResponse, 1)
+	go func() {
+		var rawRes rawResponse
+		if err := json.NewDecoder(stream).Decode(&rawRes); err != nil {
```
jalextowle: Same question here, why aren't we propagating this error?

core/ordersync/ordersync.go
```diff
@@ -0,0 +1,507 @@
...
+		if err := json.NewEncoder(stream).Encode(rawReq); err != nil {
+			s.handlePeerScoreEvent(providerID, psUnexpectedDisconnect)
+			return err
+		}
+
+		rawRes, err := waitForResponse(ctx, stream)
+		if err != nil {
+			return err
+		}
+		s.handlePeerScoreEvent(providerID, psValidMessage)
```
jalextowle: I see that we parse the response below with the subprotocol, but I still question whether we should give a peer a `psValidMessage` peer score when we cannot decode their response as a valid `rawResponse`.

core/ordersync_subprotocols.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, acceptedOrderInfo := range validationResults.Accepted {
+		if acceptedOrderInfo.IsNew {
+			log.WithFields(map[string]interface{}{
+				"orderHash": acceptedOrderInfo.OrderHash.Hex(),
+				"from":      res.ProviderID.Pretty(),
+				"protocol":  "ordersync",
+			}).Info("received new valid order from pee")
```
fabioberger:
```suggestion
			}).Info("received new valid order from peer")
```

core/ordersync/ordersync.go
albrow: Arrays and slices do not have randomized order. Maps have randomized order. When we implement more subprotocols we just need to add an additional field to `Service` to keep track of the original order for the subprotocols that were passed in.

core/ordersync/ordersync.go
albrow: Good catch. We should be returning the error. Just fixed it.

core/ordersync/ordersync.go
albrow: This is addressed by the other changes. We only add the `psValidMessage` score when receiving a valid message.

core/ordersync/ordersync.go
jalextowle: Your last sentence was the point of my comment. Is there a reason not to fix this now though?

core/ordersync/ordersync.go
jalextowle: Yep, I just wanted to mention this in case not returning the err was intentional.

PR #693, opened 2020-01-30
orderwatch: Update block headers in DB even if the block processed has no events

fabioberger:


fabioberger:
I misdiagnosed the issue.

PR #694, opened 2020-01-30
Browser Conversion Tests

jalextowle:
## Overview
This pull request introduces a large addition to 0x-mesh's testing infrastructure. Tests were added that can effectively test the Go --> Typescript interface that is used to send data to Typescript code that would like to interact with a browser based mesh node. This interface was previously a source of lots of bugs and would have been difficult to maintain with manual code review alone.

### Architecture
A new framework had to be created to test this interface properly. The crux of the issue in this test is that values must be converted to their javascript counterparts _by Go code in a Wasm binary_ and then these values should be processed by real Javascript (or in this case Typescript) code. To accomplish this behavior, a new Wasm binary was created, and this binary is embedded in a testing script written in Typescript and then served to a browser that is running in a go-routine of another Go file that serves as the entry-point to the test. The next sections will more clearly outline the role of each component of the test.

#### Entry Point
The entry into these conversion tests is the file `browser/go/conversion-test/conversion_test.go`. This file has three primary responsibilities: register test cases; serve the bundled test logic; and scrape the logs created by the bundled test script to identify test passes and failures (and missing results).

Registration of test cases is actually very straightforward. Test results are reported by the bundled test script in the form of console logs. These logs have a format of `$description: true`, where `true` indicates that the test passed. It is currently expected that these logs will be emitted in a particular order, so the test cases are registered in a slice called `testCases`. The Go code that handles emitted logs from the headless browser will just iterate over the list of `testCases` and verify that there is a log that matches every entry of `testCases`. Some features were implemented that will identify some unexpected logs (we don't want to wait forever, so logs that come very late and aren't expected will not be identified). Additionally, all of the registration logic in this file is structured in such a way that most of the boilerplate writing can be avoided. If many more of these tests need to be written in the future, code generation could hypothetically create these registration functions from reading the typescript conversion test implementation (manual generation of tests should be totally sufficient for the foreseeable future).

#### Typescript Conversion Test
Because of the way that Wasm currently works in the browser, the Typescript code in the bundle needs to control most of the testing inside of the browser. This code is responsible for instantiating a WebAssembly module that will execute the Wasm buffer and for verifying the results of calling test functions exposed by the Wasm code. The Wasm boilerplate is very similar to the implementation of `integration-tests/browser/src/index.ts`. When result are tested, logs are emitted that indicate whether or not an individual test passed or failed. There is a 1:1 correspondence between the logging calls in this file and the registration calls in the "entry point." 

#### Wasm Binary
The Wasm binary is responsible for providing functions that the Typescript Conversion test can use to receive data from Wasm. As of right now, all of these functions simply return `js.Value` conversions of all of the types that have `JSValue` methods; however, this will not necessarily be the case in the future.


browser/go/conversion-test/conversion_test.go
```diff
@@ -0,0 +1,451 @@
...
+
+	// Register the test cases that should be logged.
+	registerContractEventTest()
+	registerGetOrdersResponseTest("emptyOrderInfo", 0)
+	registerGetOrdersResponseTest("oneOrderInfo", 1)
+	registerGetOrdersResponseTest("twoOrderInfos", 2)
+	registerOrderEventTest("EmptyContractEvents", 0)
+	registerOrderEventTest("ExchangeFillContractEvent", 1)
+	registerSignedOrderTest("NullAssetData")
+	registerSignedOrderTest("NonNullAssetData")
```
fabioberger: Nit: Why are some names capitalized, and others aren't?

browser/go/conversion-test/conversion_test.go
```diff
@@ -0,0 +1,451 @@
...
+	registerStatsField(description, "numPeers")
+	registerStatsField(description, "numOrdersIncludingRemoved")
+	registerStatsField(description, "numPinnedOrders")
+	registerStatsField(description, "maxExpirationTime")
+	registerStatsField(description, "startOfCurrentUTCDay")
+	registerStatsField(description, "ethRPCRequestsSentInCurrentUTCDay")
+	registerStatsField(description, "ethRPCRateLimitExpiredRequests")
+}
+
+// FIXME(jalextowle): Generalize for non-empty validation results
```
fabioberger: Is this a "nice to have" feature, or something you plan to address in a subsequent PR?

browser/conversion-tests/conversion_test.ts
```diff
@@ -0,0 +1,1003 @@
...
+    testOrderEvents(orderEvents);
+    const signedOrders = await conversionTestCases.signedOrdersAsync();
+    testSignedOrders(signedOrders);
+    const stats = await conversionTestCases.statsAsync();
+    testStats(stats);
+    const validationResults = await conversionTestCases.validationResultsAsync();
+    testValidationResults(validationResults);
+
+    // This special #jsFinished div is used to signal the headless Chrome driver
+    // that the JavaScript code is done running.
```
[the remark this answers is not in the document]
jalextowle: This is the same method that is used in the browser integration tests. The `ts/conversion_test.ts` file will place this `#jsFinished` `div` into the `index.html` page after all tests have finished executing. I'll add a comment, but hopefully that explains the origin of this method.

browser/go/conversion-test/conversion_test.go
jalextowle: No reason. I'll make them all lowercase (my preference)

browser/go/conversion-test/conversion_test.go
jalextowle: This is a dangling comment. I left this for myself to make sure that I implemented the for-loops in the `registerValidationResultsTest` function. I will remove.

browser/conversion-tests/conversion_test.ts
```diff
@@ -0,0 +1,1003 @@
...
+    printer('parameter | value', wethWithdrawalParams.value === '150000');
+
+    // FooBarBaz
+    printer = prettyPrintTestCase('contractEventTest', 'FooBarBazEvent');
+    testContractEventPrelude(printer, contractEvents[13]);
+    printer('kind', contractEvents[13].kind === 'FooBarBazEvent');
+    const fooBarBazParams = contractEvents[13].parameters as any;
+    printer('parameter | owner', fooBarBazParams.owner === hexUtils.leftPad('0x4', 20));
+    printer('parameter | spender', fooBarBazParams.spender === hexUtils.leftPad('0x5', 20));
+    printer('parameter | value', fooBarBazParams.value === '1');
```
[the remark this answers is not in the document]
jalextowle: I commented on this offline:

"I think that we use static analysis to solve this problem. Basically, we first grep for all JSValue declarations (a regex like func\ \(($(regex for go type names))\)\ JSValue\(\)). Then we’d use that capture group to get all of the names of the types that implement JSValue, and then we’d use a similar grep expression to verify that each of these types has a corresponding test in the conversion tests. Like I said below, this is easily hackable by implementing vacuous tests but would somewhat improve our ability to catch missing tests."

browser/go/conversion-test/conversion_test.go
jalextowle: This is a dangling comment. I left this for myself to make sure that I implemented the for-loops in the registerValidationResultsTest function. I will remove.

browser/conversion-tests/conversion_test.ts
```diff
@@ -0,0 +1,1003 @@
...
+    const stats = await conversionTestCases.statsAsync();
+    testStats(stats);
+    const validationResults = await conversionTestCases.validationResultsAsync();
+    testValidationResults(validationResults);
+
+    // This special #jsFinished div is used to signal the headless Chrome driver
+    // that the JavaScript code is done running.
+    const finishedDiv = document.createElement('div');
+    finishedDiv.setAttribute('id', 'jsFinished');
+    document.querySelector('body')!.appendChild(finishedDiv); // tslint:disable-line:no-non-null-assertion
```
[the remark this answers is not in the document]
jalextowle: I found out that this can be replaced with `document.body`, which is faster, more idiomatic, and doesn't require a non-null assertion (or associated comment). Seems like a big win 😄

browser/conversion-tests/conversion_test.ts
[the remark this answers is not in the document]
albrow: I came up with this hack for the integration tests. It's written this way because `chromedp` has a limited API for "waiting for X to happen". It has a function called [`WaitVisible`](https://godoc.org/github.com/chromedp/chromedp#WaitVisible) which blocks until an element is visible in the DOM. It would be cleaner if we could wait for a JavaScript event or maybe a log message, but AFAICT there is no way to do that with `chromedp`.

browser/conversion-tests/conversion_test.ts
```diff
@@ -0,0 +1,1003 @@
...
+    };
+}
+
+function testContractEvents(contractEvents: WrapperContractEvent[]): void {
+    // ERC20ApprovalEvent
+    let printer = prettyPrintTestCase('contractEventTest', 'ERC20ApprovalEvent');
+    testContractEventPrelude(printer, contractEvents[0]);
+    printer('kind', contractEvents[0].kind === 'ERC20ApprovalEvent');
+    const erc20ApprovalParams = contractEvents[0].parameters as WrapperERC20ApprovalEvent;
+    printer('parameter | owner', erc20ApprovalParams.owner === hexUtils.leftPad('0x4', 20));
```
albrow: optional idea: How about `parameters.owner` to more clearly show which part of the object we are testing for equality? As it is currently written, we would have to do a sort of mental translation between this format and how the TypeScript type is actually structured. This idea applies in a few other places as well: anywhere there is a direct mapping to a property of a TypeScript object.

```suggestion
    printer('parameters.owner', erc20ApprovalParams.owner === hexUtils.leftPad('0x4', 20));
```

browser/conversion-tests/conversion_test.ts
```diff
@@ -0,0 +1,1003 @@
...
+    // This special #jsFinished div is used to signal the headless Chrome driver
+    // that the JavaScript code is done running.
+    const finishedDiv = document.createElement('div');
+    finishedDiv.setAttribute('id', 'jsFinished');
+    document.querySelector('body')!.appendChild(finishedDiv); // tslint:disable-line:no-non-null-assertion
+})().catch(err => {
+    throw err;
+});
+
+function prettyPrintTestCase(name: string, description: string): (section: string, value: boolean) => void {
```
albrow: The code related to printing could definitely benefit from some more comments. This is a pretty big and sometimes counterintuitive part of the tests. Right now we don't really explain anywhere in the file the importance of how the output is formatted and what it is used for.

browser/go/conversion-test/conversion_test.go
```diff
@@ -0,0 +1,451 @@
...
+func registerStatsField(description string, field string) {
+	registerTest(fmt.Sprintf("(statsTest | %s | %s)", description, field))
+}
+
+func registerValidationResultsField(description string, field string) {
+	registerTest(fmt.Sprintf("(validationResultsTest | %s | %s)", description, field))
+}
+
+func registerTest(test string) {
+	testCases = append(testCases, fmt.Sprintf("\"%s: true\"", test))
```
albrow: nit: I think using backticks is a little cleaner.

```suggestion
	testCases = append(testCases, fmt.Sprintf(`"%s: true"`, test))
```

browser/go/conversion-test/main.go
```diff
@@ -0,0 +1,673 @@
...
+func main() {
+	setGlobals()
+	triggerLoadEvent()
+	select {}
+}
+
+func setGlobals() {
+	conversionTestCases := map[string]interface{}{
+		"contractEventsAsync": js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+			return jsutil.WrapInPromise(func() (interface{}, error) {
```
albrow: I don't think we actually need to use `jsutil.WrapInPromise` here. We're just returning some values and it doesn't look like we're doing anything blocking/asynchronous like sending network requests or reading a file. If we change these functions so that they don't return a promise, we should also rename them by removing the `Async` suffix.

browser/ts/encoding.ts
```diff
@@ -0,0 +1,291 @@
+import { SignedOrder } from '@0x/order-utils';
```
albrow: How about renaming this file to __wrapper_conversions.ts__? It seems like it is focused on converting to and from the `Wrapper` types.

browser/conversion-tests/conversion_test.ts
jalextowle: I like that a lot more!

browser/go/conversion-test/conversion_test.go
jalextowle: * I am actually going to capitalize all of them because it makes more sense to leave the ContractEvent types as capitalized.

browser/conversion-tests/conversion_test.ts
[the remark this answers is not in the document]
jalextowle: After discussion offline, we have decided to monitor this situation and reevaluate the need for more advanced tools after we have a sense of how common regressions in the tests will be.

browser/conversion-tests/conversion_test.ts
```diff
@@ -0,0 +1,1003 @@
...
+let isWasmLoaded = false;
+const loadEventName = '0xmeshtest';
+window.addEventListener(loadEventName, () => {
+    isWasmLoaded = true;
+});
+
+// Start compiling the WebAssembly as soon as the script is loaded. This lets
+// us initialize as quickly as possible.
+const go = new Go();
+WebAssembly.instantiate(wasmBuffer, go.importObject)
```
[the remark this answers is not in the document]
jalextowle: I'm going to take a crack at this for this PR.

browser/conversion-tests/conversion_test.ts
```diff
@@ -81,16 +80,30 @@ window.addEventListener(loadEventName, () => {
 // Start compiling the WebAssembly as soon as the script is loaded. This lets
 // us initialize as quickly as possible.
 const go = new Go();
-WebAssembly.instantiate(wasmBuffer, go.importObject)
-    .then(module => {
-        go.run(module.instance);
+fetch('conversion_test.wasm')
+    .then(wasm => {
+        WebAssembly.instantiateStreaming(wasm, go.importObject)
```
albrow: The [Go WebAssembly Wiki](https://github.com/golang/go/wiki/WebAssembly#getting-started) suggests using `instantiateStreaming` this way:

```suggestion
        WebAssembly.instantiateStreaming(fetch('conversion_test.wasm'), go.importObject)
```

In addition to being less code, it might also perform better.

jalextowle: I see. That explains why `instantiateStreaming` can take either `Response` or `Promise<Response>` as its first argument.

PR #697, opened 2020-02-03
Addressed inconsistencies in the Typescript bindings in the browser

jalextowle:
Squashes some bugs in the browser Typescript bindings and in the Go to Javascript conversion functions.

### DIFF
--- a/browser/ts/index.ts
+++ b/browser/ts/index.ts
@@ -561,7 +561,7 @@ export interface ContractEvent {
     txHash: string;
     txIndex: number;
     logIndex: number;
-    isRemoved: string;
+    isRemoved: boolean;
     address: string;
     kind: ContractEventKind;
     parameters: ContractEventParameters;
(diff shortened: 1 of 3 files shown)


browser/ts/index.ts
```diff
@@ -271,13 +271,13 @@ export interface OrderInfo {
...
     snapshotID: string;
-    snapshotTimestamp: number;
+    snapshotTimestamp: string;
     ordersInfos: WrapperOrderInfo[];
 }
 
 export interface GetOrdersResponse {
     snapshotID: string;
-    snapshotTimestamp: number;
+    snapshotTimestamp: Date;
```
albrow: This is technically a breaking change. I'm okay with merging for now but let's update the changelog later before we release 9.0.0.

browser/ts/index.ts
```diff
@@ -480,13 +486,13 @@ export interface ExchangeCancelEvent {
 
 export interface ExchangeCancelUpToEvent {
     makerAddress: string;
-    senderAddress: string;
+    orderSenderAddress: string;
```
albrow: This is a breaking change also.

A31315: Hi

PR #698, opened 2020-02-03
Add UseBootstrapList flag to mesh-bootstrap config

opaolini:
About
===

This PR adds a `UseBootstrapList` flag to `mesh-bootstrap` config in order to have an easy way to prevent the `mesh-bootstrap` from connecting to the `DefaultBootstrapList`.

### DIFF
--- a/cmd/mesh-bootstrap/main.go
+++ b/cmd/mesh-bootstrap/main.go
@@ -277,16 +277,6 @@ func main() {
 		log.WithField("error", err).Fatal("could not bootstrap DHT")
 	}
 
-	bootstrapList := p2p.DefaultBootstrapList
-	if config.UseBootstrapList {
-		if config.BootstrapList != "" {
-			bootstrapList = strings.Split(config.BootstrapList, ",")
-		}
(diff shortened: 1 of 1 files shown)


cmd/mesh-bootstrap/main.go
```diff
@@ -290,18 +286,30 @@ func main() {
...
-	// from being banned.
-	bootstrapAddrInfos, err := p2p.BootstrapListToAddrInfos(bootstrapList)
-	if err != nil {
-		log.WithField("error", err).Fatal("could not parse bootstrap list")
-	}
-	for _, addrInfo := range bootstrapAddrInfos {
-		connManager.Protect(addrInfo.ID, "bootstrap-peer")
-		for _, addr := range addrInfo.Addrs {
-			_ = banner.ProtectIP(addr)
+	bootstrapList := p2p.DefaultBootstrapList
```
albrow: nit: If `bootstrapList` isn't used anywhere else we can scope it inside of the `if config.UseBootstrapList` block.

opaolini: 👍

PR #703, opened 2020-02-05
convertConfig Tests

jalextowle:
Added tests for the `convertConfig` function that use the new framework.

### DIFF
--- a/browser/conversion-tests/conversion_test.ts
+++ b/browser/conversion-tests/conversion_test.ts
@@ -147,10 +147,10 @@ WebAssembly.instantiateStreaming(fetch('conversion_test.wasm'), go.importObject)
     testValidationResults(validationResults);
 
     // Execute the Typescript --> Go tests
-    // tslint:disable:no-object-literal-type-assertion
     const ethereumRPCURL = 'http://localhost:8545';
 
     // Set up a Web3 Provider that uses the RPC endpoint
+    // tslint:disable:no-object-literal-type-assertion
     const provider = new Web3ProviderEngine();
     provider.addProvider(new RPCSubprovider(ethereumRPCURL));
     provider.start();
--- a/browser/conversion-tests/conversion_test.ts
+++ b/browser/conversion-tests/conversion_test.ts
@@ -1,3 +1,4 @@
+import { RPCSubprovider, Web3ProviderEngine } from '@0x/subproviders';
 import { BigNumber, hexUtils, logUtils } from '@0x/utils';
 
 import {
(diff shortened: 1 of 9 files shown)


browser/webpack.config.js
```diff
@@ -2,6 +2,9 @@ const path = require('path');
 
 module.exports = {
     entry: './conversion-tests/conversion_test.ts',
+    node: {
+      fs: 'empty'
+    },
```
fabioberger: Why was this necessary?

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -445,10 +444,11 @@ blockchainTests.resets('WSClient', env => {
                                     }
                                 `;
                                 connection.sendUTF(response);
-                                numMessages++;
-                                return;
                             }
-                            numMessages++;
+                            if (++numMessages > 2) {
+                                // NOTE(jalextowle): This log is here to help debug intermittent failures in CI.
+                                console.log('extra message', message); // tslint:disable-line:no-console
```
fabioberger: Should be resolved in: https://github.com/0xProject/0x-mesh/pull/705

browser/conversion-tests/conversion_test.ts
```diff
@@ -139,6 +146,51 @@ WebAssembly.instantiateStreaming(fetch('conversion_test.wasm'), go.importObject)
...
     testValidationResults(validationResults);
 
+    // Execute the Typescript --> Go tests
+    const ethereumRPCURL = 'http://localhost:8545';
+
+    // Set up a Web3 Provider that uses the RPC endpoint
+    // tslint:disable:no-object-literal-type-assertion
+    const provider = new Web3ProviderEngine();
+    provider.addProvider(new RPCSubprovider(ethereumRPCURL));
+    provider.start();
```
albrow: I don't think you need to call `provider.start()` here.

```suggestion
```

browser/go/conversion-test/conversion_test.go
```diff
@@ -41,7 +41,7 @@ func TestBrowserConversions(t *testing.T) {
```
albrow: `TestBrowserConversions` took 70 seconds for me the first time I ran it because I needed to download new packages from NPM. Can you change the context timeout to 120 seconds?

albrow:
Just some small suggestions. Looks pretty good overall.

browser/go/conversion-test/conversion_test.go
jalextowle: Yep

browser/webpack.config.js
jalextowle: For some reason, we get an error that is close to "cannot resolve \"fs\" in @0x/dev-utils". After doing some reading, this is actually a bug in webpack and is fixed by adding this line. This is the issue I referred to find this solution: https://github.com/webpack-contrib/css-loader/issues/447. After reading the issue again, I'm going dig into this a bit more on a low-priority cycle, because it seems like there is a better way to fix this.

PR #707, opened 2020-02-06
Streaming Wasm Loading

jalextowle:
This PR creates a new package called `@0x/mesh-browser-lite` which does not bundle the `main.wasm` file with the mesh typescript browser bindings. In the process of creating this package, a Typescript monorepo was created in the mesh codebase, which should reduce technical debt and allow Typescript code to be managed more easily.


typescript/packages/mesh-browser-lib/src/wrapper_conversion.ts
```diff
@@ -1,5 +1,6 @@
 import { SignedOrder } from '@0x/order-utils';
-import { BigNumber, providerUtils } from '@0x/utils';
+import { providerUtils } from '@0x/utils';
+import { BigNumber } from 'bignumber.js';
```
jalextowle: This is an awful hack that I needed to implement so that the `BigNumber` types from `@0x/utils` and `@0x/types` don't conflict.

cmd/cut-release/main.go
```diff
@@ -72,10 +72,12 @@ func generateTypescriptClientDocs() {
 	}
 }
 
+// TODO(jalextowle): Docs should also be built for `@0x/mesh-browser-lite` or the
```
jalextowle: @fabioberger I can add this in this PR. What docs solution do you prefer here?

fabioberger: @jalextowle let's remove this version update stages from `cut-release`. We will not be using `lerna publish` which does both the `yarn install` and version updates for us.

typescript/packages/rpc-client/ganache.log
```diff
@@ -0,0 +1,288 @@
+Saved snapshot #1
```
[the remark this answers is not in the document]
jalextowle: Ah must have missed this. =

typescript/packages/browser-lib/src/wrapper_conversion.ts
```diff
@@ -1,5 +1,8 @@
 import { SignedOrder } from '@0x/order-utils';
-import { BigNumber, providerUtils } from '@0x/utils';
+import { providerUtils } from '@0x/utils';
+// HACK(jalextowle): We need to use this BigNumber so that `yarn` doesn't complain
+// about the difference in the `@0x/types` and `@0x/utils` `BigNumber` type.
```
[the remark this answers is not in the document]
jalextowle: They _should be_ and in fact it seems that they even _are_ using the same version. Typescript isn't smart enough to figure that out though.

typescript/packages/browser-lib/src/wrapper_conversion.d.ts
```diff
@@ -0,0 +1,15 @@
...
+export declare function signedOrderToWrapperSignedOrder(signedOrder: SignedOrder): WrapperSignedOrder;
+export declare function wrapperOrderEventToOrderEvent(wrapperOrderEvent: WrapperOrderEvent): OrderEvent;
+export declare function orderEventsHandlerToWrapperOrderEventsHandler(orderEventsHandler: (events: OrderEvent[]) => void): (events: WrapperOrderEvent[]) => void;
+export declare function wrapperValidationResultsToValidationResults(wrapperValidationResults: WrapperValidationResults): ValidationResults;
+export declare function wrapperAcceptedOrderInfoToAcceptedOrderInfo(wrapperAcceptedOrderInfo: WrapperAcceptedOrderInfo): AcceptedOrderInfo;
+export declare function wrapperRejectedOrderInfoToRejectedOrderInfo(wrapperRejectedOrderInfo: WrapperRejectedOrderInfo): RejectedOrderInfo;
+export declare function wrapperStatsToStats(wrapperStats: WrapperStats): Stats;
+export declare function wrapperGetOrdersResponseToGetOrdersResponse(wrapperGetOrdersResponse: WrapperGetOrdersResponse): GetOrdersResponse;
+export declare function wrapperOrderInfoToOrderInfo(wrapperOrderInfo: WrapperOrderInfo): OrderInfo;
+//# sourceMappingURL=wrapper_conversion.d.ts.map
```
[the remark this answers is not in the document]
jalextowle: No, I didn't write these by hand. This occurred because I am explicitly importing from `@0x/mesh-browser-lib/src/wasm_exec`, `@0x/mesh-browser-lib/src/types`, and `@0x/mesh-browser-lib/src/wrapper_conversion` in other packages. To avoid the creation of these files, I could export more in `@0x/mesh-browser-lib/src` but I wasn't a huge fan of that.

Thoughts?

cmd/cut-release/main.go
```diff
@@ -95,17 +97,29 @@ func generateTypescriptBrowserDocs() {
...
+	// Update `typescript/rpc-client/package.json`
+	tsClientPackageJSONPath := "typescript/rpc-client/package.json"
 	newVersionString := fmt.Sprintf(`"version": "%s"`, version)
 	regex := `"version": "(.*)"`
 	updateFileWithRegex(tsClientPackageJSONPath, regex, newVersionString)
 
-	// Update `browser/package.json`
-	browserPackageJSONPath := "browser/package.json"
+	// Update `mesh-browser/package.json`
+	meshBrowserPackageJSONPath := "typescript/packages/mesh-browser/package.json"
```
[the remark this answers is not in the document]
jalextowle: If you're talking about the package name, I just used the original names. I'd prefer it if we remove all prefixes. If you're talking about directories, I think this was an inconsistency, as I've already changed the names of all of the directories to exclude the `mesh-` prefix.

cmd/cut-release/main.go
jalextowle: Just to clarify, does lerna publish also update the version strings in the generated docs?

jalextowle: Also, does lerna handle docs generation? It seems like this script should still be used to generate docs.

typescript/packages/browser-lib/src/wrapper_conversion.ts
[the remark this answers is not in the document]
jalextowle: I'll try to figure this out today.

typescript/package.json
```diff
@@ -0,0 +1,27 @@
...
+        "packages/*"
+    ],
+    "scripts": {
+        "lerna": "lerna",
+        "wsrun": "wsrun",
+        "build": "lerna link && lerna bootstrap && wsrun --stages --fast-exit build $PKG ",
+        "lint": "wsrun lint $PKG"
+    },
+    "devDependencies": {
+        "@0x/ts-doc-gen": "^0.0.16",
```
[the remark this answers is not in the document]
jalextowle: I'd be curious to hear you elaborate on why "we want to make it easy to move it out in the future" if you've got time. Makes sense though.

typescript/packages/browser-lite/README.md
```diff
@@ -0,0 +1,45 @@
+## @0x/mesh-browser-lite
+
+TypeScript and JavaScript bindings for running Mesh directly in the browser that have a small bundle size.
```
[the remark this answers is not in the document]
jalextowle: Yeah I'll improve this a bit. It's pretty much the same as the `@0x/mesh-browser` description, which I can also improve.

typescript/packages/browser-lib/src/wrapper_conversion.ts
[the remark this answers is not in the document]
jalextowle: Update: I was able fix this by explicitly using the `@0x/types` package in `@0x/mesh-browser-lib`. Goodbye `bignumber.js` 👋

cmd/cut-release/main.go
fabioberger: Ah sorry, realized I replied to the wrong comment here. Lerna updates versions in `package.json` but _does not_ generate docs. That part of `cut-release` must be kept.

typescript/package.json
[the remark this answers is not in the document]
fabioberger: That is simply a good guide to making these packages look and act as closely to standalone packages as possible. It's less about actually moving them out, so much as keeping their structure easily understandable to devs not familiar with yarn workspace monorepos.

typescript/packages/browser-lib/src/wrapper_conversion.d.ts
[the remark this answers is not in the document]
fabioberger: Can you try importing from `@0x/mesh-browser-lib/lib/src/types`, etc... so point to the actual `lib` dir containing the JS, instead of the `src` dir where the TS lives? I have a hunch that might solve this.

Otherwise, I'd be in favor of exporting these from `mesh-browser-lib`. It's a support package, and as such, these things we need from it are part of the interface other packages are relying on, so whether we want or not, they are part of it's "public" interface.

typescript/packages/browser-lite/package.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "@0x/mesh-browser-lite",
+    "version": "1.0.0",
```
fabioberger: We need to set the version of this package to `8.2.0`

typescript/packages/browser/package.json
```diff
@@ -5,37 +5,28 @@
     "main": "./lib/index.js",
```
fabioberger: We need to set the version of this package to `8.2.0`

typescript/packages/rpc-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/mesh-rpc-client",
-    "version": "development",
+    "version": "1.0.0",
```
fabioberger: We need to set the version of this package to `8.2.0`

typescript/packages/browser-lite/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "scripts": {
+        "build": "node --max_old_space_size=4096 ./node_modules/.bin/tsc -b",
+        "docs:md": "ts-doc-gen --sourceDir=./src --output=${npm_package_config_docsPath}",
+        "lint": "tslint --format stylish --project ."
+    },
+    "config": {
+        "docsPath": "../docs/browser"
+    },
+    "dependencies": {
+        "@0x/mesh-browser-lib": "^1.0.0",
```
fabioberger: We need to set the version of this dep to `8.2.0`

typescript/packages/browser/package.json
```diff
@@ -5,37 +5,28 @@
...
-        "ts-loader": "^6.2.1",
-        "tslint": "5.11.0",
-        "typescript": "^3.5.3",
-        "webpack": "^4.41.5",
-        "webpack-cli": "^3.3.10"
-    },
     "dependencies": {
-        "@0x/order-utils": "^10.0.1",
-        "@0x/utils": "^5.1.2",
+        "@0x/mesh-browser-lib": "^1.0.0",
```
fabioberger: We need to set the version of this dep to `8.2.0`

typescript/packages/integration-tests/package.json
```diff
@@ -5,20 +5,22 @@
...
         "rimraf": "^3.0.0",
         "ts-loader": "^6.0.4",
         "typescript": "^3.5.3",
         "webpack": "^4.39.2",
         "webpack-cli": "^3.3.7"
     },
     "dependencies": {
-        "@0x/mesh-browser": "./../../browser/",
-        "@0x/utils": "^5.1.1",
+        "@0x/mesh-browser": "^1.0.0",
```
fabioberger: We need to set the version of this dep to `8.2.0`

typescript/packages/webpack-example/package.json
```diff
@@ -5,17 +5,19 @@
...
+        "@0x/ts-doc-gen": "^0.0.16",
         "rimraf": "^3.0.0",
         "ts-loader": "^6.0.4",
         "typescript": "^3.5.3",
         "webpack": "^4.39.2",
         "webpack-cli": "^3.3.7"
     },
     "dependencies": {
-        "@0x/mesh-browser": "./../../browser/"
+        "@0x/mesh-browser": "^1.0.0"
```
fabioberger: We need to set the version of this dep to `8.2.0`

typescript/packages/browser-lib/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@0x/mesh-browser-lib",
+    "version": "1.0.0",
```
fabioberger: We need to set the version of this package to `8.2.0`

typescript/package.json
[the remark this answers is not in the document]
jalextowle: Okay, that makes a lot of sense. Thanks for explaining!

typescript/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "private": true,
+    "workspaces": [
+        "packages/*"
+    ],
+    "scripts": {
+        "lerna": "lerna",
+        "wsrun": "wsrun",
+        "build": "lerna link && lerna bootstrap && wsrun --stages --fast-exit  --exclude-missing build $PKG ",
+        "lint": "wsrun lint $PKG",
+        "test": "lerna link && lerna bootstrap && wsrun --fast-exit --exclude-missing test $PKG "
```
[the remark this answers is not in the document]
jalextowle: This is the top level `tsconfig.json` file that is currently extended from in every sub-package:

```
{
    "compilerOptions": {
        "module": "commonjs",
        "target": "es5",
        "lib": ["es2017", "dom"],
        "experimentalDecorators": true,
        "downlevelIteration": true,
        "noImplicitReturns": true,
        "pretty": true,
        "skipLibCheck": true,
        "typeRoots": ["node_modules/@0x/typescript-typings/types", "node_modules/@types"],
        "strict": true,
        // These settings are required for TypeScript project references
        "composite": true,
        "declaration": true,
        "declarationMap": true,
        "sourceMap": true
    },
    // The root of the project is just a list of references and does not contain
    // any top-level TypeScript code.
    "include": [],
    "references": [
        { "path": "./packages/browser" },
        { "path": "./packages/browser-lib" },
        { "path": "./packages/browser-lite" },
        { "path": "./packages/webpack" },
        { "path": "./packages/rpc-client" },
        { "path": "./packages/integration-tests" }
    ]
}
```

Unless I'm missing something, I think we already have this in the `tsconfig`

typescript/packages/browser/package.json
```diff
@@ -1,41 +1,34 @@
...
     "license": "Apache-2.0",
     "scripts": {
-        "prettier": "prettier --write '**/*.{ts,tsx,json,md}' --config .prettierrc",
         "build": "yarn build:go && yarn build:generate && yarn build:ts && yarn build:bundle",
-        "build:bundle": "node --max_old_space_size=3072 ./node_modules/webpack/bin/webpack.js --mode=development",
-        "build:ts": "node --max_old_space_size=4096 ./node_modules/.bin/tsc -b",
-        "build:generate": "INPUT_PATH=./wasm/main.wasm OUTPUT_PATH=./ts/generated/wasm_buffer.ts go run ./scripts/generate_wasm_buffer.go",
+        "build:bundle": "node --max_old_space_size=3072 ./node_modules/.bin/webpack --mode=development",
+        "build:ts": "tsc -b",
+        "watch:ts": "tsc -b -w",
```
fabioberger: Hm, in the monorepo we only have these commands in the top-level package.json since they partially build/watch all packages. Does it work to call them from within a specific package's dir? If so, we _could_ keep it here, but we should def. have them in the monorepo root package.json.

jalextowle: It actually does work when calling from a single package. That actually sounds kind of useful, but I'll add a top level command that doesn't use `wsrun` to make sure that it actually watches the whole repo.

.gitignore
```diff
@@ -11,11 +11,14 @@ rpc/clients/typescript/lib
...
 *.wasm
-browser/lib
-browser/ts/generated
-browser/dist/bundle.js
-examples/browser/dist/bundle.js
-integration-tests/browser/dist/bundle.js
+typescript/**/tsconfig.tsbuildinfo
+typescript/**/lib/**/*
+*/**/yarn-error.log*
+*/**/lerna-debug.log*
```
albrow: If we removed Lerna we probably don't need this anymore right?

package.json
```diff
@@ -11,6 +11,7 @@
     "devDependencies": {
         "es6-promise": "^4.2.6",
         "isomorphic-fetch": "^2.2.1",
+        "lerna": "^3.20.2",
```
albrow: I think this can also be removed.

typescript/packages/browser-lite/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@0x/mesh-browser-lite",
+    "version": "8.2.0",
+    "description": "TypeScript and JavaScript bindings for running Mesh directly in the browser.",
```
albrow: We should explain the difference between the two packages in the description. Something like:

```suggestion
    "description": "TypeScript and JavaScript bindings for running Mesh directly in the browser. To use this package, you must serve your own copy of the Mesh WebAssembly bytecode.",
```

typescript/packages/browser-lite/src/index.ts
```diff
@@ -0,0 +1,20 @@
+export * from './mesh';
+
+// Start compiling the WebAssembly as soon as the script is loaded. This lets
+// us initialize as quickly as possible.
+const go = new Go();
+WebAssembly.instantiateStreaming(fetch('main.wasm'), go.importObject)
```
albrow: I think requiring users to serve the WebAssembly bytecode at `"main.wasm"` is too strict. It prevents people from doing things like pulling in the bytecode from IPFS or using a CDN that is separate from their main URL. Can we re-architect this so that you pass in the URL or Promise in order to get the bytecode?

typescript/packages/browser/README.md
```diff
@@ -1,6 +1,11 @@
 ## @0x/mesh-browser
 
-TypeScript and JavaScript bindings for running Mesh directly in the browser.
+This package provides an easy way to run a browser-based mesh node. Specifically, it
+provides Typescript and Javascript bindings that can be used to interact with a mesh
```
albrow: nit: Mesh should be capitalized unless we are referring specifically to something that is not capitalized (e.g. the name of a Docker image or executable).

```suggestion
provides Typescript and Javascript bindings that can be used to interact with a Mesh
```

typescript/packages/browser/go/browserutil/browserutil.go
```diff
@@ -10,10 +10,10 @@ import (
 	"syscall/js"
 	"time"
 
-	"github.com/0xProject/0x-mesh/browser/go/jsutil"
-	"github.com/0xProject/0x-mesh/browser/go/providerwrapper"
 	"github.com/0xProject/0x-mesh/core"
 	"github.com/0xProject/0x-mesh/orderfilter"
+	"github.com/0xProject/0x-mesh/typescript/packages/browser/go/jsutil"
```
albrow: Importing from __typescript/something/go__ is a little awkward. If the __typescript/__ directory doesn't just contain TypeScript code, maybe we should rename it? Would you be comfortable just having a top-level __packages/__ directory?

typescript/packages/browser/package.json
```diff
@@ -1,23 +1,23 @@
 {
     "name": "@0x/mesh-browser",
-    "version": "1.0.0",
+    "version": "8.2.0",
```
albrow: In the Git Flow model, code in the `development` is not associated with a specific version. Can we change this to "development" and then add in the correct version when we cut a new release?

```suggestion
    "version": "development",
```

typescript/packages/rpc-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/mesh-rpc-client",
-    "version": "development",
+    "version": "8.2.0",
```
albrow: We should change this back to "development".

typescript/packages/webpack-example/package.json
```diff
@@ -16,6 +17,6 @@
         "webpack-cli": "^3.3.7"
     },
     "dependencies": {
-        "@0x/mesh-browser": "./../../browser/"
+        "@0x/mesh-browser": "^8.2.0"
```
albrow: We were using relative imports here for a specific reason. It let's us test whether changes we are making would break the example. Is there a reason you want to change it to `^8.2.0`?

jalextowle: Good point. Now that we are working in a yarn workspace, this needs to be `file:../browser` instead of the previous `../../browser` so that the relative path resolves in both the root package and the `@0x/mesh-webpack` package.

typescript/packages/browser/package.json
jalextowle: I'm going to keep playing with this, but the current answer I'd give is "no" because it seems to break `yarn install` at the monorepo level.

typescript/packages/browser/go/browserutil/browserutil.go
jalextowle: That works for me!

typescript/packages/browser/package.json
fabioberger: If not `development`, perhaps just always keep it as `1.0.0`?

albrow: Agree `1.0.0` is the next best thing.

packages/integration-tests/package.json
```diff
@@ -5,8 +5,10 @@
     "license": "Apache-2.0",
     "private": true,
     "scripts": {
-        "build": "node --max_old_space_size=3072 ./node_modules/webpack/bin/webpack.js --mode=development",
-        "postinstall": "yarn rimraf ./node_modules/@0x/mesh-browser/go ./node_modules/@0x/mesh-browser/scripts"
+        "build": "node --max_old_space_size=3072 ./node_modules/.bin/webpack --mode=development",
+        "build:ts": "tsc -b",
+        "postinstall": "yarn rimraf ./node_modules/@0x/mesh-browser/go ./node_modules/@0x/mesh-browser/scripts",
```
fabioberger: Can you add a comment to why this is necessary? Since JSON doesn't support comments, our convention is to add a "script" comment like so:

```
"postinstall:comment": "YOUR COMMENT HERE",
"postinstall": "yarn rimraf ./node_modules/@0x/mesh-browser/go ./node_modules/@0x/mesh-browser/scripts",
```

packages/rpc-client/package.json
```diff
@@ -41,6 +42,8 @@
         "@0x/types": "^3.1.1",
         "@0x/typescript-typings": "^5.0.1",
         "@0x/utils": "^5.1.1",
+        "@types/chai": "^4.2.8",
+        "@types/chai-as-promised": "^7.1.2",
```
fabioberger: Types are typically `devDependencies`. What am I missing?

packages/webpack-example/package.json
```diff
@@ -5,8 +5,9 @@
     "license": "Apache-2.0",
     "private": true,
     "scripts": {
-        "build": "node --max_old_space_size=3072 ./node_modules/webpack/bin/webpack.js --mode=development",
-        "postinstall": "yarn rimraf ./node_modules/@0x/mesh-browser/go ./node_modules/@0x/mesh-browser/scripts"
+        "build": "node --max_old_space_size=3072 ./node_modules/.bin/webpack --mode=development",
+        "postinstall": "yarn rimraf ./node_modules/@0x/mesh-browser/go ./node_modules/@0x/mesh-browser/scripts",
```
fabioberger: Ditto comment here.

packages/rpc-client/package.json
jalextowle: Nothing, I missed this.

.circleci/config.yml
```diff
@@ -38,11 +38,11 @@ jobs:
           name: Run linters
           command: make lint
       - run:
-          name: Build browser package
-          command: cd browser && yarn install && yarn build
+          name: Build the typescript monorepo
+          command: yarn install --force && yarn build
```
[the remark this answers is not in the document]
jalextowle: I used `--force` here because of the case where we used it in `examples/browser/`. I'll try to remove it, and see if we can remove.

PR #709, opened 2020-02-12
Increase timeout for conversion tests

albrow:
The tests often fail on the first run because it takes a long time to install some dependencies. The exact time depends on your internet speed and other factors, but 180s seems safe.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ test-go: test-go-parallel test-go-serial
 
 .PHONY: test-go-parallel
 test-go-parallel:
-	go test ./... -race -timeout 30s
+	go test ./... -race -timeout 180s
 
 .PHONY: test-go-serial
 test-go-serial:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -42,7 +42,7 @@ test-go: test-go-parallel test-go-serial
 
 .PHONY: test-go-parallel
 test-go-parallel:
-	go test ./... -race -timeout 30s
+	go test ./... -race -timeout 180s
```
jalextowle: Do we want to increase the lifetime of all of these tests? In my `convertConfig` tests PR, I added a new command to the Makefile. You might check that out and see which you prefer.

albrow: It shouldn't _really_ matter much, but I just approved https://github.com/0xProject/0x-mesh/pull/703 so let's go with that instead. I'll close this one.

PR #711, opened 2020-02-14
Delete dead code in go_js_wasm_exec

jalextowle:
Deletes some dead code in go_js_wasm_exec.

### DIFF
--- a/test-wasm/go_js_wasm_exec
+++ b/test-wasm/go_js_wasm_exec
@@ -11,13 +11,5 @@ then
 	exit 1
 fi
 
-SOURCE="${BASH_SOURCE[0]}"
-while [ -h "$SOURCE" ]; do
-	DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-	SOURCE="$(readlink "$SOURCE")"
-	[[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
-done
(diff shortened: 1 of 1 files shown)


test-wasm/go_js_wasm_exec
```diff
@@ -11,13 +11,5 @@ then
 	exit 1
 fi
 
-SOURCE="${BASH_SOURCE[0]}"
```
albrow: Can we really remove this? We copied it over from https://github.com/golang/go/blob/master/misc/wasm/go_js_wasm_exec. I must admit I'm not 100% sure what this code is supposed to be doing.

jalextowle: I definitely understand that concern. I'm currently verifying that myself, but I'm pretty sure that we can.

jalextowle: After doing some printing for myself, I found an article that explains exactly what this script does. I was pleasantly surprised that the code I deleted can be found in this stack overflow answer: https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself. From this answer, it is clear that the code that I deleted is simply used to find the path to the implementation of the bash script in such a way that it respects symlinks.

Basically, they just used this script to find the implementation of the script, which in conventional use happened to be in the same directory as `wasm_exec.js`.

PR #712, opened 2020-02-18
Add CLI for extracting public key from peer ID

fabioberger:

### DIFF
--- a/cmd/peer-id-to-pub-key/main.go
+++ b/cmd/peer-id-to-pub-key/main.go
@@ -6,7 +6,7 @@ import (
 	"log"
 	"os"
 
-	peer "github.com/libp2p/go-libp2p-peer"
+	peer "github.com/libp2p/go-libp2p-core/peer"
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


cmd/peer-id-to-pub-key/main.go
```diff
@@ -0,0 +1,30 @@
+package main
+
+import (
+	"encoding/hex"
+	"fmt"
+	"log"
+	"os"
+
+	peer "github.com/libp2p/go-libp2p-peer"
```
albrow: "github.com/libp2p/go-libp2p-peer" is deprecated. Use this instead:

```suggestion
	peer "github.com/libp2p/go-libp2p-core/peer"
```

go.mod
```diff
@@ -45,6 +45,7 @@ require (
 	github.com/libp2p/go-libp2p-core v0.3.0
 	github.com/libp2p/go-libp2p-discovery v0.2.0
 	github.com/libp2p/go-libp2p-kad-dht v0.5.0
+	github.com/libp2p/go-libp2p-peer v0.2.0
```
albrow: This should not be here. Is some other file importing this package?

fabioberger: Removed.

PR #713, opened 2020-02-19
blockwatch: Demote blockwatcher recoverable errors to warnings

fabioberger:
"I think it would be amazing if every log that has an error level indicates a bug that we need to fix."
-- @jalextowle 

Towards that aim, this PR demotes the recoverable errors returned from blockwatch to warnings. Most are unavoidable and not things we can fix so it doesn't make sense to consider them errors given the above definition. To reduce the clutter, it makes sense to demote them.

### DIFF
--- a/ethereum/blockwatch/block_watcher.go
+++ b/ethereum/blockwatch/block_watcher.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ethereum/blockwatch/block_watcher.go
```diff
@@ -171,7 +171,7 @@ func (w *Watcher) Watch(ctx context.Context) error {
 			// return an error.
 			return err
 		}
-		log.WithError(err).Error("blockwatch.Watcher error encountered")
+		log.WithError(err).Warn("blockwatch.Watcher error encountered")
```
albrow: Is it really okay to demote all block watcher errors to warnings? Aren't these sometimes an indication of a bug that we need to address?

fabioberger: The errors we see in Kibana are:
- `unknown block`
- `not found`
- `couldn't parse parameters: blockhash`
- `Post https://...: context deadline exceeded`

The only one that _could_ indicate a bug or issue is `couldn't parse parameters: blockhash`. I'll special case the other errors instead of blanked using `Warn`.

PR #716, opened 2020-02-20
Remove MiniHeaders above retention limit on startup & after processing block events

albrow:
Fixes: #667 

It also makes sure we delete miniHeaders from the DB after processing new block events that could cause new miniHeaders to be added to the DB, such that the number of miniHeaders stored does not exceed the retention limit.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -308,9 +308,13 @@ func New(config Config) (*App, error) {
 	}
 
 	// Remove any old mini headers that might be lingering in the database.
-	// HACK(albrow): This is a workaround for an issue where miniheaders are
-	// not removed from the database. This issue has been fixed and this hack
-	// should be removed after the next release.
+	// See https://github.com/0xProject/0x-mesh/issues/667 and https://github.com/0xProject/0x-mesh/pull/716
+	// We need to leave this in place becuase:
+	//
+	// 1. It is still necessary for anyone upgrading from older versions to >= 9.0.1 in the future.
+	// 2. There's still a chance there are old MiniHeaders in the database (e.g. due to a sudden
+	//    unexpected shut down).
+	//
 	err = meshDB.PruneMiniHeadersAboveRetentionLimit()
 	if err != nil {
 		return nil, err
--- a/core/core.go
+++ b/core/core.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math/big"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -48,7 +47,6 @@ import (
 )
 
 const (
-	miniHeaderRetentionLimit      = 20
 	ethereumRPCRequestTimeout     = 30 * time.Second
 	peerConnectTimeout            = 60 * time.Second
 	checkNewAddrInterval          = 20 * time.Second
@@ -313,23 +311,9 @@ func New(config Config) (*App, error) {
 	// HACK(albrow): This is a workaround for an issue where miniheaders are
 	// not removed from the database. This issue has been fixed and this hack
 	// should be removed after the next release.
-	if totalMiniHeaders, err := meshDB.MiniHeaders.Count(); err != nil {
+	err = meshDB.PruneMiniHeadersAboveRetentionLimit()
+	if err != nil {
 		return nil, err
-	} else if totalMiniHeaders > miniHeaderRetentionLimit {
-		miniHeadersToRemove := totalMiniHeaders - miniHeaderRetentionLimit
-		log.WithFields(log.Fields{
-			"numHeadersToRemove": miniHeadersToRemove,
-			"totalHeadersStored": totalMiniHeaders,
-		}).Warn("Removing outdated block headers in database (this can take a while)")
-		latestMiniHeader, err := meshDB.FindLatestMiniHeader()
-		if err != nil {
-			return nil, err
-		} else if latestMiniHeader != nil {
-			minBlockNumber := big.NewInt(0).Sub(latestMiniHeader.Number, big.NewInt(miniHeaderRetentionLimit-1))
-			if err := meshDB.ClearOldMiniHeaders(minBlockNumber); err != nil {
-				return nil, err
-			}
-		}
 	}
 
 	topics := orderwatch.GetRelevantTopics()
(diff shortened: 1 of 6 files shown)


meshdb/meshdb.go
```diff
@@ -16,6 +16,13 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	// The default miniHeaderRetentionLimit used by Mesh. This default only gets overwritten in tests.
+	defaultMiniHeaderRetentionLimit = 20
+	// The maximum MiniHeaders to query per page when deleting MiniHeaders
+	miniHeadersMaxPerPage = 5000
```
jalextowle: As mentioned yesterday offline, we should change this to `1000`. The memory usage is quite a bit lower and the execution time is almost identical (and actually a bit quicker on average).

meshdb/meshdb.go
```diff
@@ -318,26 +328,93 @@ func (m *MeshDB) FindMiniHeaderByBlockNumber(blockNumber *big.Int) (*miniheader.
...
-	for _, header := range storedHeaders {
-		if err := txn.Delete(header.ID()); err != nil {
-			return err
+	log.WithFields(log.Fields{
+		"page":    page,
+		"perPage": miniHeadersMaxPerPage,
+	}).Trace("Removing MiniHeaders")
+	offset := page * miniHeadersMaxPerPage
+	var miniHeaders []*miniheader.MiniHeader
+	if err := m.MiniHeaders.NewQuery(filter).Offset(offset).Max(miniHeadersMaxPerPage).Run(&miniHeaders); err != nil {
```
jalextowle: We need to remove `Offset` unless I'm missing something. Since we're deleting miniheaders, the offset should always be `0`, or else we'll end up with almost 70000 miniheaders.

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -34,7 +34,7 @@ import (
 
 const (
 	ethereumRPCRequestTimeout   = 30 * time.Second
-	blockWatcherRetentionLimit  = 20
+	miniHeaderRetentionLimit    = 2
```
jalextowle: Nice!

meshdb/meshdb.go
fabioberger: @albrow can you confirm? I didn't actually write this code.

albrow: Yes that's correct. I already mentioned it in Slack. This was a bug that I accidentally introduced in subsequent commits within this PR when I changed from using a single transaction to remove all the MiniHeaders to multiple transactions.

core/core.go
```diff
@@ -307,12 +306,22 @@ func New(config Config) (*App, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Remove any old mini headers that might be lingering in the database.
+	// HACK(albrow): This is a workaround for an issue where miniheaders are
```
albrow: We should update this comment. I realized we should actually leave this in place for longer (maybe indefinitely) since

1. It is still necessary for anyone upgrading from older versions to >= 9.0.1 in the future.
2. There's still a chance there are old MiniHeaders in the database (e.g. due to a sudden unexpected shut down).

meshdb/meshdb.go
albrow: @fabioberger you resolved the conversation but this is still set to `5000`. Can you update it to `1000`?

core/core.go
```diff
@@ -307,12 +306,22 @@ func New(config Config) (*App, error) {
...
+		return nil, err
+	}
+
 	topics := orderwatch.GetRelevantTopics()
 	miniHeaders, err := meshDB.FindAllMiniHeadersSortedByNumber()
 	if err != nil {
 		return nil, err
 	}
-	stack := simplestack.New(blockWatcherRetentionLimit, miniHeaders)
+	stack := simplestack.New(meshDB.MiniHeaderRetentionLimit, miniHeaders)
```
albrow: I think it's cleaner if `MiniHeaderRetentionLimit` is declared in `core` and passed through to both `meshdb` and `simplestack`. The `core` package should be the glue that connects everything together.

meshdb/meshdb.go
```diff
@@ -318,26 +328,93 @@ func (m *MeshDB) FindMiniHeaderByBlockNumber(blockNumber *big.Int) (*miniheader.
...
 
+// UpdateMiniHeaderRetentionLimit updates the MiniHeaderRetentionLimit. This is only used by tests in order
+// to set the retention limit to a smaller size, making the tests shorter in length
+func (m *MeshDB) UpdateMiniHeaderRetentionLimit(limit int) error {
+	m.MiniHeaderRetentionLimit = limit
+	return m.PruneMiniHeadersAboveRetentionLimit()
+}
+
+// PruneMiniHeadersAboveRetentionLimit prunes miniHeaders from the DB that are above the retention limit
+func (m *MeshDB) PruneMiniHeadersAboveRetentionLimit() error {
```
albrow: Why don't we just pass in `retemtionLimit` to `PruneMiniHeadersAboveRetentionLimit`? It would let us remove a lot of code and have better separation of concerns.

meshdb/meshdb.go
albrow: @fabioberger can you also improve the test so that it would catch this bug? We need to add more than `miniHeadersMaxPerPage` in order to test that pagination works correctly and all the MiniHeaders are removed. To make testing this case easier/faster, you can unit test `ClearOldMiniHeaders` in parallel without setting up block watcher or order watcher.

albrow: We can also refactor this a bit since we no longer need to pass in the `page` number. And then, we should probably rename `clearMiniHeadersForPage`.

meshdb/meshdb.go
albrow: Decided to just take this on as technical debt for now. We should address it in the future, possibly at the same time as cleaning up some other parts of the MiniHeaders collection.

meshdb/meshdb.go
albrow: I added this test and applied the other recommended changes.

core/core.go
albrow: Decided to just take this on as technical debt for now. We should address it in the future, possibly at the same time as cleaning up some other parts of the MiniHeaders collection.

PR #717, opened 2020-02-20
Fixed problem where blockchainTests.resets suite would not reset

jalextowle:
Fixed an inconsistency in the rpc clients test.


rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -49,6 +50,13 @@ blockchainTests.resets('WSClient', env => {
             };
             provider = web3Factory.getRpcProvider(providerConfigs);
 
+            // HACK(jalextowle): We can't currently specify an out of process provider for a blockchainTests
+            // suit, so we need to update env.blockchainLifecycle so that the resets suite works as expected.
```
fabioberger:
```suggestion
            // suite, so we need to update env.blockchainLifecycle so that the `resets` suite works as expected.
```

rpc/clients/typescript/test/ws_client_test.ts
```diff
@@ -49,6 +50,13 @@ blockchainTests.resets('WSClient', env => {
...
+            // Additionally, `env.blockchainLifecycle.startAsync` will have already been called at this point,
+            // so we call `startAsync` here to avoid an unbalanced snapshot stack.
```
fabioberger: Can you elaborate on what you mean by this? If `startAsync` was already called, shouldn't we be calling `stopAsync` to avoid an unbalanced stack?

jalextowle: This is pretty hacky, which makes the situation a little bit complicated to understand. Essentially, blockchainTests.resets internally has “beforeEach” hooks that run before every test iteration. From the test results, we were trying to revert a snapshot that didn’t exist at the end of the test suite. This is because blockchainTests.resets has a different “env.blockchainLifecycle” which gets the first call to “startAsync” and is then replaced. So we need some way to replace this first snapshot, so that it can be safely reverted at the end.

PR #720, opened 2020-02-21
Release version 9.0.1

albrow:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -95,7 +95,7 @@ steps:
       api_key:
         from_secret: github_public_repo
       files:
-        - /
+        - *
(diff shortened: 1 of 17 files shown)


.drone.yml
```diff
@@ -95,7 +95,7 @@ steps:
       api_key:
         from_secret: github_public_repo
       files:
-        - /
+        - *
```
jalextowle: What did this fix?

albrow: I posted about it in Slack. See https://drone.spaceship.0x.org/0xProject/0x-mesh/748/3/1.

albrow: Also https://github.com/0xProject/0x-mesh/pull/715

PR #723, opened 2020-02-21
cmd: Network bridge

fabioberger:
This PR implements a network bridge that can be used to bridge two Mesh networks after introducing a network-level breaking change.


cmd/mesh_bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+		panic(err)
+	}
+
+	if env.Verbose {
+		log.SetLevel(log.Level(4))
+	} else {
+		log.SetLevel(log.Level(0))
+	}
+
+	firstClient, err := rpc.NewClient(env.SecondWSRPCAddress)
```
jalextowle: I think this would be clearer if the `firstClient` used the `FirstWSRPCAddress` and the `secondClient` used the `SecondWSRPCAddress`. Anything I'm missing?

jalextowle: Even better would be `Version8WSRPCAddress`, `v8Client`, etc.

cmd/mesh_bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		pipeOrders(secondClient, firstClient, "v8", "v9")
+	}()
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		go pipeOrders(firstClient, secondClient, "v9", "v8")
```
jalextowle: Should this be a go func? If it should, should the other `pipeOrders` be called from a go func?

cmd/mesh_bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+		select {
+		case orderEvents := <-orderEventsChan:
+			for _, orderEvent := range orderEvents {
+				if orderEvent.EndState != zeroex.ESOrderAdded {
+					continue
+				}
+				log.WithField("orderHash", orderEvent.OrderHash.Hex()).Info(fmt.Sprintf("Found new order %s -> %s", inLabel, outLabel))
+				signedOrdersCache = append(signedOrdersCache, orderEvent.SignedOrder)
+				// If more than 10 orders discovered, pipe them to the outClient
+				if len(signedOrdersCache) >= 10 {
```
jalextowle: Nit: Let's make this magic number a constant.

cmd/mesh_bridge/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN apk update && apk add ca-certificates nodejs-current npm make git dep gcc build-base musl linux-headers
+
+WORKDIR /0x-mesh
+
+ADD . ./
+
+RUN go build ./cmd/mesh_bridge
+
+# Final Image
+FROM alpine:3.9
```
jalextowle: Should this be `alpine:3.9`? We are using `golang:1.13.4-alpine3.10` above.

cmd/mesh_bridge/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Note: this must be built from the root of the project with:
```
albrow: @fabioberger can you move this to __/dockerfiles/mesh-bridge/Dockerfile__?

cmd/mesh_bridge/main.go
```diff
@@ -0,0 +1,106 @@
+// +build !js
+
+// mesh_bridge is a short program that bridges two Mesh nodes. This is useful in cases where
```
albrow: Can we rename this to `mesh-bridge` to match the name of our other executables (e.g. `mesh-bootstrap`, `peer-id-to-pub-key`)? This also seems to be the convention in the Go ecosystem (e.g. `golang-ci-lint`), though it isn't used by quite everyone.

cmd/mesh_bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/0xProject/0x-mesh/rpc"
+	"github.com/0xProject/0x-mesh/zeroex"
+	"github.com/plaid/go-envvar/envvar"
+	log "github.com/sirupsen/logrus"
+)
+
+type clientEnvVars struct {
+	FirstWSRPCAddress  string `envvar:"FIRST_WS_RPC_ADDRESS"`
+	SecondWSRPCAddress string `envvar:"SECOND_WS_RPC_ADDRESS"`
+	Verbose            bool   `envvar:"VERBOSE"`
```
albrow: Why not let this be an `int` and call it `Verbosity` like we do for our other executables? It gives users more control and doesn't require any extra lines of code to support.

cmd/mesh_bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.WithField("stats", stats).Info("Spun up second client")
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		pipeOrders(secondClient, firstClient, "v8", "v9")
```
albrow: I know this bridge was written with versions 8 and 9 in mind, but we could easily use it for future versions or for poor man's cross-topic forwarding without making any additional changes. I don't think these "v8" and "v9" labels should be hard-coded. There's also nothing to indicate which of the two addresses should be "v8" and which should be "v9" and it isn't enforced in any way. Maybe we should just remove the labels? Alternatively we could pass them in as environment variables or use the existing `FirstWSRPCAddress` and `SecondWSRPCAddress` as labels.

cmd/mesh_bridge/main.go
albrow: We're already in a goroutine so we don't need to start another one.

```suggestion
		pipeOrders(firstClient, secondClient, "v9", "v8")
```

cmd/mesh_bridge/main.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		log.WithError(err).Fatal("could not create client")
+	}
+	stats, err = secondClient.GetStats()
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.WithField("stats", stats).Info("Spun up second client")
+
+	wg := &sync.WaitGroup{}
```
albrow: You've set up this `WaitGroup` so that we can wait for `pipeOrders` to return, but because of the way the function is written, it never does. We're waiting for nothing. We should either:

1.  Make `pipeOrders` return an error if something goes wrong, and then handle that error by e.g. canceling a context. Look at [how we do this in the `core` package](https://github.com/0xProject/0x-mesh/blob/1a4a775b4e6c6f726d004a1e897ffc88bc197180/core/core.go#L470).
2. Remove this `WaitGroup` and just exit the program if something goes wrong in `pipeOrders` with `log.Fatal`.

cmd/mesh_bridge/main.go
albrow: I think we need to rearchitect this or just remove it. As it's currently written, we could receive up to 9 orders and wait indefinitely to share them. This kind of delay is important and can have an impact on market makers.

The best way to do this is similar to [what we do in the `p2p` package when batching incoming messages](https://github.com/0xProject/0x-mesh/blob/1a4a775b4e6c6f726d004a1e897ffc88bc197180/p2p/node.go#L565). We wait for up to `maxReceiveBatch` messages or `receiveTimeout` seconds between messages, whichever comes first. 

(Though looking at the `p2p` code again I think there is a subtle bug where the data associated with a message could have been sent over the network but we canceled the `context` before processing it. Need to take a closer look here and maybe tweak it. Still the same principle applies.)

fabioberger: Agreed. Using that type of logic now.

cmd/mesh_bridge/Dockerfile
jalextowle: Okay, gotcha. I would think that we should update all 3, but maybe @albrow can comment on this?

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+	ethrpc "github.com/ethereum/go-ethereum/rpc"
+	"github.com/plaid/go-envvar/envvar"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	firstWSRPCAddressLabel  = "FirstWSRPCAddress"
+	secondWSRPCAddressLabel = "SecondWSRPCAddress"
+	maxReceiveBatch         = 10
+	receiveTimeout          = 10 * time.Second
```
jalextowle: Optional Suggestion: Considering @albrow's original comment about wanting to make sure that orders are forwarded quickly after being received, I could see a case for making `receiveTimeout` an `envvar` so that it can be configured without changing the source code.

cmd/mesh_bridge/Dockerfile
albrow: We should update them all, but it also likely doesn't matter much. We only switched to `alpine3.10` for the intermediate container that builds Mesh because `golang:1.13.4-alpine3.9` does not exist.

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/0xProject/0x-mesh/zeroex"
+	ethrpc "github.com/ethereum/go-ethereum/rpc"
+	"github.com/plaid/go-envvar/envvar"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	firstWSRPCAddressLabel  = "FirstWSRPCAddress"
+	secondWSRPCAddressLabel = "SecondWSRPCAddress"
+	maxReceiveBatch         = 10
```
albrow: I think it makes more sense to increase `maxReceiveBatch` and decrease `receiveTimeout`. From what we have measured so far, it looks like Mesh performs best when receiving batches of ~100 orders and usually takes less than 1 second to validate them all. How about:

```
maxReceiveBatch = 100
receiveTimeout  = 1 * time.Second
```

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+	receiveTimeout          = 10 * time.Second
+)
+
+type clientEnvVars struct {
+	FirstWSRPCAddress  string `envvar:"FIRST_WS_RPC_ADDRESS"`
+	SecondWSRPCAddress string `envvar:"SECOND_WS_RPC_ADDRESS"`
+	Verbosity          int    `envvar:"VERBOSE"`
```
albrow:
```suggestion
	Verbosity          int    `envvar:"VERBOSITY"`
```

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+	go func() {
+		pipeOrders(secondClient, firstClient, secondWSRPCAddressLabel, firstWSRPCAddressLabel)
+	}()
+	pipeOrders(firstClient, secondClient, firstWSRPCAddressLabel, secondWSRPCAddressLabel)
+}
+
+func pipeOrders(inClient, outClient *rpc.Client, inLabel, outLabel string) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	orderEventsChan := make(chan []*zeroex.OrderEvent, 8000)
```
albrow: Why 8000? Can we make this a constant?

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Info(fmt.Sprintf("Sent %d orders from %s -> %s. Accepted: %d Rejected: %d", len(incomingSignedOrders), inLabel, outLabel, len(validationResults.Accepted), len(validationResults.Rejected)))
+	}
+}
+
+func receiveBatch(inChan chan []*zeroex.OrderEvent, subscription *ethrpc.ClientSubscription, inLabel, outLabel string) ([]*zeroex.SignedOrder, error) {
+	signedOrdersCache := []*zeroex.SignedOrder{}
+	ticker := time.NewTicker(receiveTimeout)
```
albrow: We only need the ticker to go off once right? How about using [`time.After`](https://golang.org/pkg/time/#After) instead?

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+		validationResults, err := outClient.AddOrders(incomingSignedOrders)
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Info(fmt.Sprintf("Sent %d orders from %s -> %s. Accepted: %d Rejected: %d", len(incomingSignedOrders), inLabel, outLabel, len(validationResults.Accepted), len(validationResults.Rejected)))
+	}
+}
+
+func receiveBatch(inChan chan []*zeroex.OrderEvent, subscription *ethrpc.ClientSubscription, inLabel, outLabel string) ([]*zeroex.SignedOrder, error) {
+	signedOrdersCache := []*zeroex.SignedOrder{}
```
albrow: Maybe subjective, but I wouldn't call this a cache. Typically I think of a cache as something with contents that can change over time and that has an eviction strategy, neither of which apply here.

```suggestion
	signedOrders := []*zeroex.SignedOrder{}
```

dockerfiles/mesh-bridge/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Note: this must be built from the root of the project with:
+#
+#     docker build . -f ./cmd/mesh-bridge/Dockerfile
```
albrow:
```suggestion
#     docker build . -f ./dockerfiles/mesh-bridge/Dockerfile
```

Can you also add a command to the Makefile? I forgot that we did this for our other dockerfiles.

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+	for {
+		incomingSignedOrders, err := receiveBatch(orderEventsChan, clientSubscription, inLabel, outLabel)
+		if err != nil {
+			log.Fatal(err)
+		}
+		validationResults, err := outClient.AddOrders(incomingSignedOrders)
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Info(fmt.Sprintf("Sent %d orders from %s -> %s. Accepted: %d Rejected: %d", len(incomingSignedOrders), inLabel, outLabel, len(validationResults.Accepted), len(validationResults.Rejected)))
```
albrow: If we ever want to track these logs in Kibana, it's really important that the message stays the same each time. the fields can change, but the message should be the same.

```suggestion
		log.WithFields(log.Fields{
		    "from": inLabel,
		    "to": outLabel,
		    "numSent": len(incomingSignedOrders),
		    "numAccepted": len(validationResults.Accepted),
		    "numRejected": len(validationResults.Rejected),
		}).Info("Finished sending orders")
```

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		select {
+		case <-ticker.C:
+			return signedOrdersCache, nil
+		case orderEvents := <-inChan:
+			for _, orderEvent := range orderEvents {
+				if orderEvent.EndState != zeroex.ESOrderAdded {
+					continue
+				}
+				log.WithField("orderHash", orderEvent.OrderHash.Hex()).Info(fmt.Sprintf("Found new order %s -> %s", inLabel, outLabel))
```
albrow: Same comment here regarding logging.

cmd/mesh-bridge/main.go
```diff
@@ -0,0 +1,112 @@
...
+	Verbosity          int    `envvar:"VERBOSE"`
+}
+
+func main() {
+	env := clientEnvVars{}
+	if err := envvar.Parse(&env); err != nil {
+		panic(err)
+	}
+
+	log.SetLevel(log.Level(env.Verbosity))
```
albrow: If we do want these logs to show up in Kibana, we also need to add these lines in order to format them properly.

```suggestion
	log.SetLevel(log.Level(env.Verbosity))
	log.SetFormatter(&log.JSONFormatter{})
	log.AddHook(loghooks.NewKeySuffixHook())
```

cmd/mesh_bridge/Dockerfile
fabioberger: Upgraded.

cmd/mesh-bridge/main.go
fabioberger: This was copied over from `subscribe-to-orders` example. Moved to constant.

PR #724, opened 2020-02-25
Localize contract addresses state

jalextowle:
Localizes the state for contract addresses to `App`.


core/core_test.go
```diff
@@ -101,6 +101,33 @@ func init() {
 	}
 }
 
+func TestIdempotentAppInitialization(t *testing.T) {
```
albrow: nit: We don't want `New` to be idempotent and that's not really what this PR does. Idempotent means that subsequent calls are effectively a no-op. What we want is for subsequent calls to return a new app instance (and for closing the old instance to not leave any dangling goroutines).

core/core.go
```diff
@@ -217,8 +218,9 @@ func New(config Config) (*App, error) {
 	})
 
 	// Add custom contract addresses if needed.
+	chainIDToContractAddresses := ethereum.NewChainIDToContractAddresses()
 	if config.CustomContractAddresses != "" {
-		if err := parseAndAddCustomContractAddresses(config.EthereumChainID, config.CustomContractAddresses); err != nil {
+		if err := parseAndAddCustomContractAddresses(config.EthereumChainID, chainIDToContractAddresses, config.CustomContractAddresses); err != nil {
```
albrow: I think this is making things more complicated than they need to be. Instead of passing around `chainID` and `chainIDToContractAddresses`, why don't we just pass in `contractAddresses`? We don't need to keep looking up `chainIDToContractAddresses[chainID]` everywhere and there is no case I can think of where we would need to look at the contract addresses for more than one chain ID within a single `core.App` instance. This would save us memory, computation, and lines of code.

core/core_test.go
jalextowle: Yeah, I was thinking about that this morning. In general, idempotent means that the result of applying a function is the same after the first application. Admittedly, using that terminology here is a stretch.

core/core.go
jalextowle: Great, this was the feedback I was looking for on the draft. I wasn’t sure how flexible we wanted to leave this (a bunch of database callbacks actually read the chainID from orders, so i wondered if these could be different chainIDs from the app’s chainID in the future. It sounds like that isn’t the case)

core/core.go
```diff
@@ -217,10 +218,15 @@ func New(config Config) (*App, error) {
 	})
 
 	// Add custom contract addresses if needed.
+	var contractAddresses ethereum.ContractAddresses
+	var err error
```
jalextowle: Is there a way to avoid doing this? The two `vars` in a row are pretty gross 🤢

albrow: @jalextowle you can do something like this:

```
	// Add custom contract addresses if needed.
	var contractAddresses ethereum.ContractAddresses
	if config.CustomContractAddresses != "" {
		addresses, err := parseAndValidateCustomContractAddresses(config.EthereumChainID, config.CustomContractAddresses)
		if err != nil {
			return nil, err
		}
		contractAddresses = addresses
	} else {
		addresses, err := ethereum.NewContractAddressesForChainID(config.EthereumChainID)
		if err != nil {
			return nil, err
		}
		contractAddresses = addresses
	}
```

But honestly I prefer the way it's currently written. I don't see any problem with having two `var` declarations.

core/core_test.go
```diff
@@ -33,7 +33,9 @@ const (
 )
 
 func TestEthereumChainDetection(t *testing.T) {
-	meshDB, err := meshdb.New("/tmp/meshdb_testing/" + uuid.New().String())
+	contractAddresses, err := ethereum.NewContractAddressesForChainID(constants.TestChainID)
```
albrow: I see this pattern a lot in the tests:

```
	contractAddresses, err := ethereum.NewContractAddressesForChainID(constants.TestChainID)
	require.NoError(t, err)
```

We don't really need to go through the hoops of looking up the contract addresses for `constants.TestChainID` since we know that the addresses are defined. Can you make a helper method in __ethereum/contract_addresses__ that just returns the addresses for Ganache and doesn't return an error?

core/core.go
jalextowle: Yeah, I'll just leave it. It's not too bad. I just wanted to see if there was some way that I could use `:=` here, but the scoping doesn't seem to work.

core/core_test.go
jalextowle: Good catch!

ethereum/contract_addresses.go
```diff
@@ -22,6 +22,22 @@ type ContractAddresses struct {
 	ChaiToken           common.Address `json:"chaiToken"`
 }
 
+func GanacheAddresses() ContractAddresses {
```
albrow: Sorry for not catching this last time, but after seeing this I realized we can simplify this even further. We don't really need this to be function now, it can just be a `var`.

```suggestion
var GanacheAddresses = ContractAddresses {
```

jalextowle: Yeah I also thought of that. Is there a way of making variables readonly in Golang? I think that's really what we want here

PR #729, opened 2020-02-27
Improve peer discovery

### LINKED ISSUE #582: Connect to peers at the rendezvous point before accepting new connections
With the introduction of custom order filters in #563, we will also start using a different rendezvous point for each topic. Because of the fact that incoming connections are always accepted by default in libp2p it is possible that we could hit the high mark for number of peers before even attempting to dial any peers on the same rendezvous point. This means that it would take longer to find a peer that is using the same topic, especially as the total number of peers in the network grows.

One solution to this problem would be to connect to peers at the rendezvous point _before_ advertising ourselves on the DHT. This means that it is unlikely that other peers will try to connect to us early on and increases the chances of finding peers that are using the same topic.

albrow:
Fixes #582.

This PR includes a handful of related changes which improve peer discovery, especially for those using custom order filters. In general this will make the process of finding new peers, and specifically finding the _right_ peers, much faster.

1. Created separate rendezvous points for peers using custom order filters.
2. Made peer discovery parallel so it doesn't block message handling.
3. Connect to peers on the rendezvous points before advertising ourselves.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -561,8 +561,6 @@ func (n *Node) startMessageHandler(ctx context.Context) error {
 		if mathrand.Float64() <= chanceToCheckBandwidthUsage {
 			n.banner.CheckBandwidthUsage()
 		}
-
-		return nil
 	}
 }
 
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -46,6 +46,9 @@ const (
 	peerGraceDuration = 10 * time.Second
 	// peerDiscoveryInterval is how frequently to try to connect to new peers.
 	peerDiscoveryInterval = 5 * time.Second
+	// advertiseDelay is the amount of time to wait during startup before advertising
+	// ourselves on the DHT.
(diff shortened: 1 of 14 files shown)


p2p/node.go
```diff
@@ -390,10 +397,94 @@ func (n *Node) Start() error {
...
+		case <-innerCtx.Done():
+			// If the context was canceled, return immediately. Don't bother
+			// advertising ourselves.
+			return
+		case <-time.After(advertiseDelay):
+			// Otherwise, advertise ourselves on the DHT after a delay.
+			// The delay allows us to prioritize connecting to peers with a matching
+			// rendezvous point in order of preference.
+			//
+			// Note(albrow): Advertise doesn't return an error, so we have no
```
jalextowle: Hypothetically, we could wrap the `Advertise` function and actually test to verify that `Advertise` was successful if this in important.

albrow: How would that work exactly?

Issue #171, opened 2019-06-15
mesh: Make sure Ethereum RPC URL points to configured networkId


fabioberger:
Mesh operators can specify both a `ETHEREUM_RPC_URL` and `ETHEREUM_NETWORK_ID`. They could inadvertently specify an Ethereum RPC URL pointing to an Ethereum node running on a different network then the networkId specified by `ETHEREUM_NETWORK_ID`. 

We could either:
- Remove the `ETHEREUM_NETWORK_ID` config and on boot-up fetch the `networkId` from the Ethereum node.
OR
- Fetch the networkId from the Ethereum node, compare it to the supplied `ETHEREUM_NETWORK_ID` and simply check that they are the same.

albrow:
> Fetch the networkId from the Ethereum node, compare it to the supplied ETHEREUM_NETWORK_ID and simply check that they are the same.

This is my preferred option because the other would incur additional start up latency which is unnecessary in most cases.

fabioberger:
Should we crash if it turns out it isn't correct? Or simply print a warning to the console?

jalextowle:
@albrow Can we close this now that #733 has been merged?

albrow:
There are still some issues with the approach taken in #733. We made the decision to do the chainId check in a separate goroutine so that we don't slow down the rest of Mesh when first starting up. However, that also means there is currently a race condition where we can start processing blocks or incoming orders before we realize there is a chainId mismatch.

I proposed a different solution in https://github.com/0xProject/0x-mesh/issues/866. We can close this issue if we decide to go that route. Otherwise we should leave it open and address the problems with the current implementation somehow.

PR #733, opened 2020-03-02
Compare chain id env variable with rpc on startup

kimpers:
Implements #171

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -580,32 +580,36 @@ func (app *App) Start(ctx context.Context) error {
 		orderWatcherErrChan <- app.orderWatcher.Watch(innerCtx)
 	}()
 
-	// Ensure that ETHEREUM_CHAIN_ID matches chain id of the RPCClient
+	// Ensure that RPC client is on the same ChainID as is configured with ETHEREUM_CHAIN_ID
+	chainIdCheckerWg := &sync.WaitGroup{}
 	chainIdMismatchErrChan := make(chan error, 1)
 	wg.Add(1)
+	chainIdCheckerWg.Add(1)
 	go func() {
 		defer wg.Done()
+		defer chainIdCheckerWg.Done()
 		ctx, cancel := context.WithTimeout(context.Background(), ethereumRPCRequestTimeout)
 		defer cancel()
 
-		var chainID string
+		var chainIDRaw string
 
-		err := app.ethRPCClient.CallContext(ctx, &chainID, "eth_chainId")
+		err := app.ethRPCClient.CallContext(ctx, &chainIDRaw, "eth_chainId")
 		if err != nil {
 			chainIdMismatchErrChan <- err
 			return
 		}
 
-		cleanedChainId := strings.Replace(chainID, "0x", "", -1)
-		chainIDDecimal, err := strconv.ParseInt(cleanedChainId, 16, 64)
+		// Value in RPC response is a string in hexadecimal form e.g. "0x539"
+		// Need to remove the "0x" prefix before parsing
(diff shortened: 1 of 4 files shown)


core/core.go
```diff
@@ -711,6 +745,14 @@ func (app *App) Start(ctx context.Context) error {
 		}
 	}
 
+	// Need to wait for the async ETH RPC call to finish before we know if the chain id is correctly configured
+	chainIdCheckerWg.Wait()
+	if err := <-chainIdMismatchErrChan; err != nil {
```
fabioberger: Move this into a `case` statement in the above select block.

core/core.go
```diff
@@ -552,6 +552,40 @@ func (app *App) Start(ctx context.Context) error {
...
 
+	// Ensure that RPC client is on the same ChainID as is configured with ETHEREUM_CHAIN_ID
+	chainIdCheckerWg := &sync.WaitGroup{}
+	chainIdMismatchErrChan := make(chan error, 1)
+	wg.Add(1)
+	chainIdCheckerWg.Add(1)
+	go func() {
+		defer wg.Done()
+		defer chainIdCheckerWg.Done()
+		ctx, cancel := context.WithTimeout(context.Background(), ethereumRPCRequestTimeout)
```
fabioberger: Can we move the actual logic in this go routine into a helper method with a good name? Having it in-lined here feels like the wrong level of abstraction when compared with the rest of it's parent's function logic.

fabioberger: Use `innerCtx` instead of `context.Background()` here.

albrow:
@kimpers @fabioberger do you mind if I take over this PR to push it over the finish line? This check would actually be really useful and solve a common problem, especially in the browser.

kimpers:
> @kimpers @fabioberger do you mind if I take over this PR to push it over the finish line? This check would actually be really useful and solve a common problem, especially in the browser.

Feel free 👍 Beware that there seems to be a strange race condition with the test, locally on my machine it passes some times and fails other times. 

When you figure out what the underlying issue is I'd be very interested to know what causes it.

albrow:
@kimpers the race condition here was definitely a tricky one. It doesn't directly stem from the code you added, moreso this new code made an existing bug more apparent.

Inside of `core.App.Start` there is a `select` statement where we wait for several different goroutines to send an error through their respective error channels (e.g. `p2pErrChan`, `orderWatcherErrChan`, or the new `chainIDMismatchErrChan` which you added in this PR):

```go
select {
		case err := <-p2pErrChan:
			if err != nil {
				log.WithError(err).Error("p2p node exited with error")
				cancel()
				return err
			}
		case err := <-orderWatcherErrChan:
			if err != nil {
				log.WithError(err).Error("order watcher exited with error")
				cancel()
				return err
			}

		// etc...
}
```

There was an implicit assumption here that fi one of the goroutines sent an error through their respective channels, it was time to shut the app down. But that assumption was wrong. There are some cases where a goroutine may send `nil` through the channel to indicate that it has exited _without_ an error, and in this case we don't shut down the application. The most straightforward example of this is that the ETH RPC rate limiter sends `nil` through `ethRPCRateLimiterErrChan` immediately if ETH RPC rate limiting has been disabled. In the old code this would cause us to exit the `select` statement, and therefore not receive errors from any of the other (still running) goroutines.

I fixed this by wrapping the `select` statement in a `for` loop. This way if one of the goroutines exits and sends `nil` through its respective error channel, we keep waiting to receive errors on the other channels and then handle them appropriately.

albrow:
@fabio This PR is now ready for review. I changed it to no longer be a draft.

PR #735, opened 2020-03-02
Update project status in README

albrow:


README.md
```diff
@@ -12,7 +12,7 @@
...
+- Order sharing: We have recently made significant improvements to our order sharing algorithm, including reducing bandwidth usage and CPU usage by at least an order of magnitude on average. See https://github.com/0xProject/0x-mesh/pull/692 and https://github.com/0xProject/0x-mesh/pull/732. However, we are still working on accurately testing and measuring the speed at which orders propagate through the network with different network sizes and topologies. In some circumstances, it may take longer than we would like for orders to reach the majority of nodes in the network. This is an area we will continue to focus on and improve.
```
jalextowle: This seems fine to me. My only thought is that it might be better to avoid using words like "recently" in the README.md since it will feel out of date if we don't update it. That said, we do update this pretty regularly.

albrow: I kind of see where you're coming from but it's hard to avoid words like "recently" while at the same time providing an accurate picture of the current development status. Let's just try and keep it updated until we can remove this section entirely.

PR #737, opened 2020-03-02
Added better documentation and an example for `@0x/mesh-browser-lite`

jalextowle:


packages/webpack-example-lite/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    Mesh,
+    OrderEvent,
+    SignedOrder,
+    SupportedProvider,
+} from '@0x/mesh-browser-lite';
+
+// tslint:disable:no-console
+(async () => {
+    // Load the WebAssembly bytecode.
+    loadMeshStreamingWithURL('main.wasm');
```
albrow: I didn't catch this until actually seeing the usage example, but `loadMeshStreamingWithURL` and `loadMeshStreaming` should probably be async functions that return a Promise. Instead, what we're currently doing is:

```
         setImmediate(() => {
                throw err;
            });
```

This can lead to confusing behavior if the WebAssembly bytecode fails to compile for any reason (e.g. the request times out).


We only do this inside of `@0x/mesh-browser` because (1) We want to the WebAssembly to start loading immediately and (2) We're the ones responsible for bundling __main.wasm__ and we bundle it without sending any network requests, which drastically decreases the chances of something going wrong.

packages/webpack-example-lite/README.md
```diff
@@ -0,0 +1,29 @@
+## 0x Mesh Browser Example
```
albrow: This page is full of out-of-date links. We should also add a line near the top explaining the difference between this example and __../webpack-example__.

docs/browser.md
```diff
@@ -44,13 +48,51 @@ yarn add @0x/mesh-browser
...
-## Documentation
+### Documentation
 
 Documentation for the `@0x/mesh-browser` package is available at
 [0x-org.gitbook.io/mesh/browser](https://0x-org.gitbook.io/mesh/browser).
 
-## Example usage
+### Example usage
 
 [The examples/browser directory](../examples/browser) includes a bare-bones
```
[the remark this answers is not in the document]
jalextowle: Yep, good catch.

docs/browser.md
```diff
@@ -44,13 +48,51 @@ yarn add @0x/mesh-browser
...
+Or if you are using `yarn`:
+
+```
+yarn add @0x/mesh-browser-lite
+```
+
+### Documentation
+
+Documentation for the `@0x/mesh-browser-lite` package is available at
+[0x-org.gitbook.io/mesh/browser-lite](https://0x-org.gitbook.io/mesh/browser-lite).
```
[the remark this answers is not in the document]
jalextowle: Okay, I could have misunderstood what this is. I thought this would point to the documentation for `@0x/mesh-browser-lite`.

packages/webpack-example-lite/src/index.ts
jalextowle: I'm going to address this in a separate PR.

docs/browser-bindings/browser/README.md
```diff
@@ -1,6 +1,13 @@
+# @0x/mesh-browser - v1.0.0
```
albrow: Is there a specific reason to add `v1.0.0` here? AFAIK we only need to do that in __package.json__ because Yarn requires it. I would prefer not to include version numbers anywhere in the `development` branch. If we must include them, it should either be "development" or the commit hash.

jalextowle: Since the version are `1.0.0`, `@0x/ts-docs-gen` (which uses `typedoc` under-the-hood) will use that in the title. We could add a script to doc generation if you think it makes sense.

As far as using `1.0.0` in the `package.json`, that is what I needed to do to use `yarn install`, but I will double-check this.

packages/browser-lite/src/index.ts
```diff
@@ -15,5 +15,7 @@ export async function loadMeshStreamingWithURLAsync(url: string): Promise<void>
 export async function loadMeshStreamingAsync(response: Response | Promise<Response>): Promise<void> {
     const go = new Go();
     const module = await WebAssembly.instantiateStreaming(response, go.importObject);
-    return go.run(module.instance);
+    setImmediate(() => {
```
albrow: Can you add a brief comment about why `setImmediate` is needed here?

docs/browser.md
```diff
@@ -20,16 +20,34 @@ approach comes with a lot of advantages, but also has some drawbacks:
...
 
 For your convenience, we've published an NPM package called `@0x/mesh-browser`
-which includes the WebAssembly bytecode and a lightweight wrapper around it. You
-install this package in exactly the same way as any other NPM package and using
-it feels exactly like using a native TypeScript/JavaScript library.
+which includes the WebAssembly bytecode and a lightweight wrapper around it. To
+find more information on optimizing the startup time of using Mesh in the browser,
+refer to the next section on the `@0x/mesh-browser-lite` package. This package is
+used exactly the same way as any other NPM package and using it feels exactly
+like using a native TypeScript/JavaScript library.
```
albrow: IMO this still reads like `@0x/mesh-browser-lite` is the one that "feels exactly like using a native TypeScript/JavaScript library". If anything it's the opposite. `@0x/mesh-browser` is more like a typical TypeScript/JavaScript library since it bundles everything together for you (at the cost of increasing load times) and does not require you to separately load the Wasm bytecode.

Can we just rearrange the sentences to make it more clear?

```suggestion
which includes the WebAssembly bytecode and a lightweight wrapper around it. This
package is used exactly the same way as any other NPM package and using it feels exactly
like using a native TypeScript/JavaScript library. To find more information on optimizing the
startup time of using Mesh in the browser, refer to the next section on the
`@0x/mesh-browser-lite` package.
```

docs/browser.md
```diff
@@ -20,16 +20,34 @@ approach comes with a lot of advantages, but also has some drawbacks:
...
+`@0x/mesh-browser-lite` package that provides an identical abstraction around a
+browser-based Mesh node without requiring that Wasm bytecode be bundled with the
+rest of the webpage's code. Additionally, this package makes use of the
+WebAssembly's streaming functionality, which could provide a speedup to load
+times for users of browser-based Mesh.
+
+Using this package is a bit more complicated than using the `@0x/mesh-browser` package.
+This package provides an API that will create a Mesh wrapper using a provided URL
+or Response object that serves the WebAssembly bytecode that will be used to run a
+Mesh node.
```
albrow: Can we be more specific here? Something like:

It requires you to obtain a copy of the WebAssembly bytecode from [the releases page](https://github.com/0xProject/0x-mesh/releases) and serve it on your own server or a CDN.  It also requires you to make one additional function call to either `loadMeshStreamingWithURLAsync` or `loadMeshStreamingAsync` in your code before initializing Mesh.

docs/browser.md
```diff
@@ -41,16 +59,35 @@ Or if you are using `yarn`:
...
-[The examples/browser directory](../examples/browser) includes a bare-bones
-example of how to use the `@0x/mesh-browser` and bundle everything together with
-Webpack.
+[The packages/webpack-example-lite directory](../packages/webpack-example)
+includes a bare-bones example of how to use the `@0x/mesh-browser-lite` and bundle
+everything together with Webpack. This is slightly different than the other example
+because it requires that WebAssembly bytecode be provided by the user of the package.
+WebAssembly binaries for each version of 0x-mesh that has an associated
+`@0x/mesh-browser-lite` package can be found in the
+[0x-mesh release notes](https://github.com/0xProject/0x-mesh/releases).
```
albrow: Some of this is redundant and I think it's better to go into more detail above in the `@0x/mesh-browser-lite` section. Putting the information all in one place will make it easier for readers to understand.

```suggestion
everything together with Webpack.
```

docs/browser.md
```diff
@@ -25,25 +25,29 @@ approach comes with a lot of advantages, but also has some drawbacks:
...
 browser-based Mesh node without requiring that Wasm bytecode be bundled with the
 rest of the webpage's code. Additionally, this package makes use of the
 WebAssembly's streaming functionality, which could provide a speedup to load
 times for users of browser-based Mesh.
 
 Using this package is a bit more complicated than using the `@0x/mesh-browser` package.
-This package provides an API that will create a Mesh wrapper using a provided URL
-or Response object that serves the WebAssembly bytecode that will be used to run a
-Mesh node.
+WebAssembly binaries for each version of 0x-mesh that has an associated `@0x/mesh-browser-lite`
```
albrow: 👌 ✨

yarn-error.log
```diff
@@ -0,0 +1,9392 @@
+Arguments:
```
fabioberger: @jalextowle can you add this to the `.gitignore`?

jalextowle: Hmm, I already put `*/**/yarn-error.log` in the .gitignore 😕

PR #740, opened 2020-03-02
Promisified the `@0x/mesh-browser-lite` API

jalextowle:

### DIFF
--- a/packages/browser-lite/src/index.ts
+++ b/packages/browser-lite/src/index.ts
@@ -13,12 +13,7 @@ export async function loadMeshStreamingWithURLAsync(url: string): Promise<void> 
  * @param response The Wasm response that supplies the Wasm binary.
  */
 export async function loadMeshStreamingAsync(response: Response | Promise<Response>): Promise<void> {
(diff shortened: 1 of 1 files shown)


packages/browser-lite/src/index.ts
```diff
@@ -4,29 +4,21 @@ export * from './mesh';
...
-            // If the Wasm bytecode didn't compile, Mesh won't work. We have no
-            // choice but to throw an error.
-            setImmediate(() => {
-                throw err;
-            });
-        });
+export async function loadMeshStreamingAsync(response: Response | Promise<Response>): Promise<void> {
+    return new Promise((resolve, reject) => {
+        const go = new Go();
+        WebAssembly.instantiateStreaming(response, go.importObject)
```
albrow: Since we changed this to an `async` function, can't we use `await` here instead of `then` and `catch`?

albrow:
Looks good. @jalextowle don't forget to update the example on https://github.com/0xProject/0x-mesh/pull/737.

PR #741, opened 2020-03-04
Improved RPC Server Initialization Log 

jalextowle:
The `rpc_integration_test` would intermittently fail because the HTTP and WebSocket RPC Servers would log the same message when they started. The tests were designed with only one RPC server in mind, and this meant that there was a race-condition in the tests.

### DIFF
--- a/cmd/mesh/main.go
+++ b/cmd/mesh/main.go
@@ -66,7 +66,7 @@ func main() {
 	go func() {
 		defer wg.Done()
 		log.WithField("ws_rpc_addr", config.WSRPCAddr).Info("starting WS RPC server")
-		rpcServer := instantiateServer(ctx, app, config.WSRPCAddr)
+		rpcServer := instantiateServer(ctx, app, config.WSRPCAddr, "WS")
 		if err := rpcServer.Listen(ctx, rpc.WSHandler); err != nil {
 			wsRPCErrChan <- err
 		}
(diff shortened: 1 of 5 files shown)


cmd/mesh/main.go
```diff
@@ -66,7 +66,7 @@ func main() {
 	go func() {
 		defer wg.Done()
 		log.WithField("ws_rpc_addr", config.WSRPCAddr).Info("starting WS RPC server")
-		rpcServer := instantiateServer(ctx, app, config.WSRPCAddr)
+		rpcServer := instantiateServer(ctx, app, config.WSRPCAddr, "WS")
```
albrow: I don't like passing around magic strings like this. As a reader and potential caller of `instantiateServer`, it is difficult to understand what the string does and which values are allowed/accepted. Can we either:

1. Infer the type of server from the given address.
2. Create two separate functions which share common functionality. (e.g. `instantiateHTTPServer` and `instantiateWSServer`).

jalextowle: Good point. I will keep this in mind in the future.

cmd/mesh/main.go
```diff
@@ -21,10 +21,10 @@ import (
 type standaloneConfig struct {
 	// WSRPCAddr is the interface and port to use for the JSON-RPC API over
 	// WebSockets. By default, 0x Mesh will listen on localhost and port 60557.
-	WSRPCAddr string `envvar:"WS_RPC_ADDR" default:"localhost:60557"`
+	WSRPCAddr string `envvar:"WS_RPC_ADDR" default:"ws://localhost:60557"`
 	// HTTPRPCAddr is the interface and port to use for the JSON-RPC API over
 	// HTTP. By default, 0x Mesh will listen on localhost and port 60556.
-	HTTPRPCAddr string `envvar:"HTTP_RPC_ADDR" default:"localhost:60556"`
+	HTTPRPCAddr string `envvar:"HTTP_RPC_ADDR" default:"http://localhost:60556"`
```
fabioberger: Why is this change strictly necessary? I'd be against it if it's optional since it introduces a breaking change to the config envs of Mesh that _many_ people rely on.

PR #744, opened 2020-03-04
[zeroex/ordervalidator]: Add StaticCall support for `checkGasPrice`

jalextowle:
## Description
The 0x V3 protocol supports several asset data types that Mesh does not currently support. Several use cases have recently surfaced that use StaticCall asset data in the `makerFeeAssetData` field of the order, so this pull request provides a foundation for supporting specific types of staticcalls.


zeroex/asset_data_decoder.go
```diff
@@ -55,10 +59,17 @@ type ERC1155AssetData struct {
 type ERC20BridgeAssetData struct {
 	TokenAddress  common.Address
 	BridgeAddress common.Address
-	Calldata      []byte
+	BridgeData    []byte
```
jalextowle: ABIs don't hash parameter names into function selectors so this slipped through the cracks. Changing it back to agree with the implementation.

ethereum/contract_addresses.go
```diff
@@ -138,3 +121,22 @@ func ValidateContractAddressesForChainID(chainID int, addresses ContractAddresse
...
+		ERC1155Proxy:        common.HexToAddress("0x6a4a62e5a7ed13c361b176a5f62c2ee620ac0df8"),
+		Exchange:            common.HexToAddress("0x48bacb9266a570d521063ef5dd96e61686dbe788"),
+		Coordinator:         common.HexToAddress("0x4d3d5c850dd5bd9d6f4adda3dd039a3c8054ca29"),
+		CoordinatorRegistry: common.HexToAddress("0xaa86dda78e9434aca114b6676fc742a18d15a1cc"),
+		DevUtils:            common.HexToAddress("0xa31e64ea55b9b6bbb9d6a676738e9a5b23149f84"),
+		WETH9:               common.HexToAddress("0x0b1ba0af832d7c05fd64161e0db78e85978e8082"),
+		ZRXToken:            common.HexToAddress("0x871dd7c2b4b25e1aa18728e9d5f2af4c4e431f5c"),
+		ChaiBridge:          common.HexToAddress("0x0000000000000000000000000000000000000000"),
+		ChaiToken:           common.HexToAddress("0x0000000000000000000000000000000000000000"),
+		MaximumGasPrice:     common.HexToAddress("0x0000000000000000000000000000000000000000"),
```
jalextowle: This is currently failing because `MaximumGasPrice` was not deployed on Ganache. I'm going to fix that in the monorepo, which should fix this.

ethereum/contract_addresses.go
```diff
@@ -138,3 +121,22 @@ func ValidateContractAddressesForChainID(chainID int, addresses ContractAddresse
...
+		Exchange:            common.HexToAddress("0x48bacb9266a570d521063ef5dd96e61686dbe788"),
+		Coordinator:         common.HexToAddress("0x4d3d5c850dd5bd9d6f4adda3dd039a3c8054ca29"),
+		CoordinatorRegistry: common.HexToAddress("0xaa86dda78e9434aca114b6676fc742a18d15a1cc"),
+		DevUtils:            common.HexToAddress("0xa31e64ea55b9b6bbb9d6a676738e9a5b23149f84"),
+		WETH9:               common.HexToAddress("0x0b1ba0af832d7c05fd64161e0db78e85978e8082"),
+		ZRXToken:            common.HexToAddress("0x871dd7c2b4b25e1aa18728e9d5f2af4c4e431f5c"),
+		ChaiBridge:          common.HexToAddress("0x0000000000000000000000000000000000000000"),
+		ChaiToken:           common.HexToAddress("0x0000000000000000000000000000000000000000"),
+		MaximumGasPrice:     common.HexToAddress("0x2c530e4ecc573f11bd72cf5fdf580d134d25f15f"),
+	}
```
fabioberger: 🙏

zeroex/asset_data_decoder.go
```diff
@@ -55,10 +68,21 @@ type ERC1155AssetData struct {
...
+}
+
+// StaticCallAssetData represents a StaticCallAssetData
+type StaticCallAssetData struct {
+	StaticCallTargetAddress common.Address
+	StaticCallData          []byte
```
fabioberger: Are there somehow the canonical field names? Otherwise, I'd be in favor of dropping the `StaticCall` prefix as it stutters.

zeroex/orderwatch/order_watcher.go
```diff
@@ -1730,6 +1730,14 @@ func (w *Watcher) addAssetDataAddressToEventDecoder(assetData []byte) error {
...
 		w.eventDecoder.AddKnownERC1155(decodedAssetData.Address)
 		w.contractAddressToSeenCount[decodedAssetData.Address] = w.contractAddressToSeenCount[decodedAssetData.Address] + 1
+	case "StaticCall":
+		var decodedAssetData zeroex.StaticCallAssetData
+		err := w.assetDataDecoder.Decode(assetData, &decodedAssetData)
+		if err != nil {
+			return err
+		}
+		// NOTE(jalextowle): Only one staticcall is supported right now,
+		// so we don't neeed to add any addresses to the event decoder
```
fabioberger: I think this comment should be adjusted to explain that the only currently supported static call is to a function that does not rely on on-chain state and therefore no additional contract events need to be watched to watch those orders.

zeroex/ordervalidator/order_validator.go
```diff
@@ -782,6 +789,28 @@ func (o *OrderValidator) isSupportedAssetData(assetData []byte) bool {
...
+	case "checkGasPrice":
+		var decodedStaticCallData zeroex.CheckGasPriceStaticCallData
+		err := o.assetDataDecoder.Decode(staticCallAssetData.StaticCallData, &decodedStaticCallData)
+		if err != nil {
+			return false
+		}
+		// We currently restrict the `checkGasPrice` staticcall to the known MaximumGasPrice contract.
+		if o.contractAddresses.MaximumGasPrice == constants.NullAddress || staticCallAssetData.StaticCallTargetAddress != o.contractAddresses.MaximumGasPrice {
+			return false
+		}
```
fabioberger: We only check the contract address here. Where do we validate that it's calling one of the two functions expected? If we aren't validating the function and only the contract, then let's remove the function ABI's from the ABI decoder.

zeroex/asset_data_decoder.go
jalextowle: These are the canonical field names from the contracts. I'd prefer to stick with those in these structures and in the ABIs so that the contracts can be the single source of truth for these names.

zeroex/ordervalidator/order_validator.go
jalextowle: We validate that by decoding the `staticCallData`. We first get the name of the staticcall that is encoded by the `staticCallData`. Then if it is `checkGasPrice` (the name of one of our two functions), we actually decode the data to ensure that it is a valid `checkGasPrice` staticcall.

It's only when we've confirmed that the `staticCallData` encodes a valid `checkGasPrice` staticcall that we check the contract address. Is there something that I'm missing about the ABI Decoder? My assumption was that I could use its `Decode` function to validate that a given `staticCallData` satisfies a given ABI.

fabioberger: Ah yep, that sounds good enough, I must have just misread. Thanks for clarifying.

zeroex/orderwatch/order_watcher.go
```diff
@@ -1730,6 +1730,16 @@ func (w *Watcher) addAssetDataAddressToEventDecoder(assetData []byte) error {
...
+	case "StaticCall":
+		var decodedAssetData zeroex.StaticCallAssetData
+		err := w.assetDataDecoder.Decode(assetData, &decodedAssetData)
+		if err != nil {
+			return err
+		}
+		// NOTE(jalextowle): The only staticcall that is currently supported
+		// only relies on transaction gas price. This means that we do not need
+		// to monitor any new contract addresses because the only supported
+		// staticcall doesn't rely on any blockchain state state.
```
albrow:
```suggestion
		// staticcall doesn't rely on any blockchain state.
```

zeroex/asset_data_decoder.go
```diff
@@ -55,10 +68,21 @@ type ERC1155AssetData struct {
 type ERC20BridgeAssetData struct {
 	TokenAddress  common.Address
 	BridgeAddress common.Address
-	Calldata      []byte
+	BridgeData    []byte
+}
+
+// StaticCallAssetData represents a StaticCallAssetData
```
albrow: nit: I know this PR isn't the first time this type of comment showed up, but this doesn't add any useful information and IMO we'd be better off removing it. Alternatively we can replace it with a comment that provides some more context.

zeroex/ordervalidator/order_validator_test.go
```diff
@@ -210,6 +210,131 @@ func TestBatchValidateAValidOrder(t *testing.T) {
...
+	orderValidator, err := New(ethRPCClient, constants.TestChainID, constants.TestMaxContentLength, ganacheAddresses)
+	require.NoError(t, err)
+
+	accepted, rejected := orderValidator.BatchOffchainValidation(signedOrders)
+	assert.Len(t, accepted, 0)
+	require.Len(t, rejected, 1)
+	testSignedOrder.ResetHash()
+	expectedOrderHash, err := testSignedOrder.ComputeOrderHash()
+	require.NoError(t, err)
+	assert.Equal(t, expectedOrderHash, rejected[0].OrderHash)
```
albrow: As I mentioned in Slack, let's actually check the `RejectedOrderStatus` here instead of just checking that the order was rejected.

PR #745, opened 2020-03-04
Release version 9.1.0

albrow:


docs/browser-bindings/browser/README.md
```diff
@@ -1,6 +1,13 @@
+# @0x/mesh-browser - v9.1.0
```
jalextowle: It doesn't look like `docs/browser-bindings/browser/reference.md` was created. We will need this for the gitbook setup to work correctly.

jalextowle: I approved, but we should make sure this is addressed before merging

PR #753, opened 2020-03-09
Fix race condition in Typescript RPC tests

jalextowle:


jalextowle:
This didn't really fix the problem and this PR will be closed in favor of a later fix

PR #754, opened 2020-03-09
Decrease max bytes/s from 100 MiB to 1 MiB

albrow:
Since https://github.com/0xProject/0x-mesh/pull/692 was implemented we have been seeing vastly reduced bandwidth usage. We can now safely reduce the per-IP bandwidth limits.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -68,9 +68,7 @@ const (
 	chanceToCheckBandwidthUsage = 0.1
 	// defaultMaxBytesPerSecond is the maximum number of bytes per second that a
 	// peer is allowed to send before failing the bandwidth check.
-	// TODO(albrow): Reduce this limit once we have a better picture of what
-	// real world bandwidth should be.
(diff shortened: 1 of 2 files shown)


p2p/node.go
```diff
@@ -68,9 +68,7 @@ const (
 	chanceToCheckBandwidthUsage = 0.1
 	// defaultMaxBytesPerSecond is the maximum number of bytes per second that a
 	// peer is allowed to send before failing the bandwidth check.
-	// TODO(albrow): Reduce this limit once we have a better picture of what
-	// real world bandwidth should be.
-	defaultMaxBytesPerSecond = 104857600 // 100 MiB.
+	defaultMaxBytesPerSecond = 1048576 // 1 MiB.
```
jalextowle: Should we add a new comment back in here?

albrow: 👍Just updated the comments in both places.

PR #756, opened 2020-03-13
Support adding multiple orders to Watcher.add in a single batch.

albrow:
This PR solves one of the main performance bottlenecks for order propagation speed, especially when adding many orders at once. Prior to this PR, `orderwatch.Watcher.add` only accepted one order at a time. This PR updates it to accept multiple orders in an array, which means we can add all of them in a single database transaction.

One side-effect of this PR is that it is slightly more likely to end up with more than `MAX_ORDERS_IN_STORAGE` in storage temporarily. The chance of that happening is still slim, it would never exceed the maximum batch size, and would quickly be corrected. I think this trade-off is worth it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
+## v9.2.0
+
+### Features ✅
+
+- Greatly reduced latency for propagating orders, especially for browser nodes [#756](https://github.com/0xProject/0x-mesh/pull/756).
+
+
 ## v9.1.0
 
 ### Features ✅
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This changelog is a work in progress and may contain notes for versions which ha
 ### Features ✅
 
 - Greatly reduced latency for propagating orders, especially for browser nodes [#756](https://github.com/0xProject/0x-mesh/pull/756).
+- Added support for `checkGasPrice` StaticCall asset data [#744](https://github.com/0xProject/0x-mesh/pull/744)
 
 
 ## v9.1.0
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -929,59 +929,57 @@ func (w *Watcher) add(orderInfos []*ordervalidator.AcceptedOrderInfo, validation
 			IsRemoved:                false,
 			IsPinned:                 pinned,
 		}
-		err = txn.Insert(order)
-		if err != nil {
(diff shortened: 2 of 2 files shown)


zeroex/orderwatch/order_watcher.go
```diff
@@ -1392,35 +1397,24 @@ func (w *Watcher) ValidateAndStoreValidOrders(ctx context.Context, orders []*zer
 	results.Accepted = append(results.Accepted, zeroexResults.Accepted...)
 	results.Rejected = append(results.Rejected, zeroexResults.Rejected...)
 
-	// Store valid orders
-	allOrderEvents := []*zeroex.OrderEvent{}
-	for i, acceptedOrderInfo := range results.Accepted {
+	// Flter out only the new orders.
```
fabioberger:
```suggestion
	// Filter out only the new orders.
```

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 This changelog is a work in progress and may contain notes for versions which have not actually been released. Check the [Releases](https://github.com/0xProject/0x-mesh/releases) page to see full release notes and more information about the latest released versions.
 
+## v9.2.0
+
+### Features ✅
+
+- Greatly reduced latency for propagating orders, especially for browser nodes [#756](https://github.com/0xProject/0x-mesh/pull/756).
```
jalextowle: This doesn't have to do with your PR, but it looks like I didn't include a `CHANGELOG.md` entry in my PR. 

If you wouldn't mind, could we include that in this PR? I think something along the lines of the following would suffice: "Added support for `checkGasPrice` StaticCall asset data [#744](https://github.com/0xProject/0x-mesh/pull/744)"

zeroex/orderwatch/order_watcher.go
```diff
@@ -919,82 +919,87 @@ func (w *Watcher) add(orderInfo *ordervalidator.AcceptedOrderInfo, validationBlo
...
+		} else {
+			err = txn.Insert(order)
+			if err != nil {
+				if _, ok := err.(db.AlreadyExistsError); ok {
+					// If we're already watching the order, that's fine in this case. Don't
+					// return an error.
+					return orderEvents, nil
+				}
+				if _, ok := err.(db.ConflictingOperationsError); ok {
+					logger.WithFields(logger.Fields{
```
jalextowle: I understand that this code was taken from the previous implementation, but could you add a comment here explaining why we don't want to return an error?

albrow: @fabioberger wrote this. After looking at it again, I actually think we should return an error here. Any thoughts Fabio?

fabioberger: Yes, we should. This is legacy. At one point we had a bug causing these conflicting operations to happen, and since they were harmless, we did not want to return an error. Not that those are fixed, we should return the error.

albrow: 👍 changed it so we return the error.

PR #758, opened 2020-03-17
Release version 9.2.0

albrow:


zeroex/orderwatch/order_watcher.go
```diff
@@ -919,82 +921,87 @@ func (w *Watcher) add(orderInfo *ordervalidator.AcceptedOrderInfo, validationBlo
...
+					// If we're already watching the order, that's fine in this case. Don't
+					// return an error.
+					return orderEvents, nil
+				}
+				if _, ok := err.(db.ConflictingOperationsError); ok {
+					logger.WithFields(logger.Fields{
+						"error": err.Error(),
+						"order": order,
+					}).Error("Failed to insert order into DB")
+					return orderEvents, nil
```
jalextowle: @albrow I may have missed something on your last PR, but I thought we were going to return an error here.

zeroex/asset_data_decoder.go
```diff
@@ -55,10 +68,20 @@ type ERC1155AssetData struct {
...
+	StaticCallData          []byte
+	ExpectedReturnHashData  [32]byte
 }
 
-// MultiAssetData represents an MultiAssetData
+type CheckGasPriceStaticCallData struct {
+	MaxGasPrice *big.Int
+}
+
+// MultiAssetData represents a MultiAssetData
```
jalextowle: Non-blocking: This file has a lot of the unproductive comments that I mentioned. We should delete this at some point.

zeroex/asset_data_decoder.go
```diff
@@ -20,6 +20,16 @@ const ERC721AssetDataID = "02571792"
 // ERC1155AssetDataID is the assetDataId for ERC721 tokens
 const ERC1155AssetDataID = "a7cb5fb7"
 
+// StaticCallAssetDataID is the assetDataId for staticcalls
```
jalextowle: Non-blocking: I used `assetDataId` here to keep it consistent with the rest of the file; however, we refer to these as `assetProxyIds` in the monorepo. I'd be in favor of following that convention instead.

meshdb/meshdb.go
```diff
@@ -568,6 +568,19 @@ func parseContractAddressesAndTokenIdsFromAssetData(assetData []byte, contractAd
...
+	case "StaticCall":
+		var decodedAssetData zeroex.StaticCallAssetData
+		err := assetDataDecoder.Decode(assetData, &decodedAssetData)
+		if err != nil {
+			return nil, err
+		}
+		// NOTE(jalextowle): As of right now, none of the supported staticcalls
+		// have important information in the StaticCallData. This will probably
+		// change, which may require addtional information from this case.
+		a := singleAssetData{
```
jalextowle: Now that I look at this again, there really isn't a reason to add this for the current staticcalls. We aren't watching any events from that address, so this will just waste space. It also doesn't really make sense in the context of the callbacks registered in the `setupOrders` function. This may change with future staticcalls. 

I'd be in favor of fixing this before the release (I'll make a PR after this review).

ethereum/contract_addresses.go
```diff
@@ -20,22 +20,13 @@ type ContractAddresses struct {
...
-	Exchange:            common.HexToAddress("0x48bacb9266a570d521063ef5dd96e61686dbe788"),
-	Coordinator:         common.HexToAddress("0x4d3d5c850dd5bd9d6f4adda3dd039a3c8054ca29"),
-	CoordinatorRegistry: common.HexToAddress("0xaa86dda78e9434aca114b6676fc742a18d15a1cc"),
-	DevUtils:            common.HexToAddress("0xa31e64ea55b9b6bbb9d6a676738e9a5b23149f84"),
-	WETH9:               common.HexToAddress("0x0b1ba0af832d7c05fd64161e0db78e85978e8082"),
-	ZRXToken:            common.HexToAddress("0x871dd7c2b4b25e1aa18728e9d5f2af4c4e431f5c"),
-	ChaiBridge:          common.HexToAddress("0x0000000000000000000000000000000000000000"),
-	ChaiToken:           common.HexToAddress("0x0000000000000000000000000000000000000000"),
-}
+// GanacheAddresses The addresses that the 0x contracts were deployed to on the Ganache snapshot (chainID = 1337).
```
jalextowle: Non-blocking: This is technically inaccurate because not all of these contracts are 0x contracts.

jalextowle: I suspect that even more of them will by non-0x contracts in the future.

zeroex/orderwatch/order_watcher.go
albrow: Sorry I made that change in a different spot here: https://github.com/0xProject/0x-mesh/pull/756/commits/536804815b797f5e63d0598b9e4f9a9e615e0542. We should probably return an error everywhere we log one in this function, but I don't think it should block the PR.

cmd/cut-release/main.go
```diff
@@ -115,6 +87,9 @@ func updateHardCodedVersions(version string) {
 	regex = `"version": "(.*)"`
 	updateFileWithRegex(browserPackageJSONPath, regex, newVersionString)
 	newBrowserLiteDependencyString := fmt.Sprintf(`"@0x/mesh-browser-lite": "^%s"`, version)
+	// NOTE(jalextowle): `@0x/mesh-browser` uses the local version of `@0x/mesh-browser-lite`
```
albrow: @jalextowle `@0x/mesh-browser-lite` has been published now. Do we need to change anything here?

jalextowle: Nope. This comment is a bit misleading. Basically, we’re just overwriting the “@0x/mesh-browser-lite” dependency so that it refers to the new version number (which means that @0x/mesh-browser-lite should be published to npm first).

zeroex/orderwatch/order_watcher.go
jalextowle: Yep, I agree. I saw the other changes and thought that might have happened. I can include these changes in my post-9.2.0 cleanup PR

Issue #425, opened 2019-09-26
Clean up scenario package


albrow:
There's some duplicated code in the `scenario` package involving how orders are created and balances/allowances are set up. It looks like it can potentially be cleaned up and/or de-duped.

albrow:
One more thing we should change is that the `scenario` package should offer methods for creating signed orders without setting approvals or balances. The tests in https://github.com/0xProject/0x-mesh/pull/476 could benefit from this because the code we are testing doesn't actually care about balances/allowances. However because of how the `scenario` package currently works we had no choice but to use blockchain life cycles and run the tests in serial instead of parallel.

albrow:
Another thing we should do is add support for creating many orders in one batch, without waiting for approvals/balances to be set for each order one at a time.

albrow:
Implemented in #779.

PR #760, opened 2020-03-17
Fix a bug in the ordersync pagination subprotocol

albrow:
This PR fixes a critical bug in the `FilteredPaginationSubProtocol` for ordersync. There was an off-by-one error in the for loop where we were incrementing `currentPage` _before_ checking for the end condition in each iteration of the loop. This caused every other page to be skipped, resulting in around 50% of the orders actually being shared.

I wanted to create a regression test as part of this PR but doing so has proven difficult. We need to create at least `paginationSubprotocolPerPage * 2 + 1` orders in order to see this bug in the tests. This has proven difficult for two reasons:

1. Since `paginationSubprotocolPerPage` is not configurable via an environment variable, we are stuck with the default value of 500. We don't want to make it configurable as it is too easy for users of Mesh to pick the wrong value. It's better for us to tune it over time based on results from Mesh Manager and in the real world.
2. The `scenario` package does not support creating more than one order (and the associated balances/allowances) at a time. This means that it would have to create at least 1001 orders one at a time, waiting for a new block to be mined for each order. It ends up taking way too long (more than 10 minutes). See #425.

We should add regression tests, but it's better to do that in a separate PR. This is fairly critical bug fix so we should try to get it out as quickly as possible.

I did perform an ad hoc regression test by manually changing `paginationSubprotocolPerPage` to 5, increasing the timeout for the test to 2 minutes, and creating 50 orders (10 pages). This ad hoc test fails prior to the bug fix and passes after it was implemented.

### DIFF
--- a/core/ordersync_subprotocols.go
+++ b/core/ordersync_subprotocols.go
@@ -72,13 +72,19 @@ func (p *FilteredPaginationSubProtocol) HandleOrderSyncRequest(ctx context.Conte
 			return nil, fmt.Errorf("FilteredPaginationSubProtocol received request with wrong metadata type (got %T)", req.Metadata)
 		}
 	}
+
 	// It's possible that none of the orders in the current page match the filter.
 	// We don't want to respond with zero orders, so keep iterating until we find
 	// at least some orders that match the filter.
 	filteredOrders := []*zeroex.SignedOrder{}
 	var snapshotID string
-	var currentPage int
-	for currentPage = metadata.Page; len(filteredOrders) == 0; currentPage += 1 {
+	currentPage := metadata.Page
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
 		// Get the orders for this page.
 		ordersResp, err := p.app.GetOrders(currentPage, p.perPage, metadata.SnapshotID)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


core/ordersync_subprotocols.go
```diff
@@ -72,13 +72,19 @@ func (p *FilteredPaginationSubProtocol) HandleOrderSyncRequest(ctx context.Conte
...
 		}
 	}
+
 	// It's possible that none of the orders in the current page match the filter.
 	// We don't want to respond with zero orders, so keep iterating until we find
 	// at least some orders that match the filter.
 	filteredOrders := []*zeroex.SignedOrder{}
 	var snapshotID string
-	var currentPage int
-	for currentPage = metadata.Page; len(filteredOrders) == 0; currentPage += 1 {
```
albrow: The bug here is really subtle. `currentPage += 1` gets executed _before_ we check the exit condition `len(filteredOrders) == 0` for the loop . As a result, the returned `ordersync.Response` always had a `Page` that was one higher than it was supposed to be. The requester always requests the _next_ page (`response.Page + 1`) so this meant we were skipping every other page.

jalextowle: Nice catch. This is pretty weird behavior for a for-loop, but it’s good to know.

PR #764, opened 2020-03-18
Periodically rerun ordersync instead of running it only once on startup

albrow:
This PR adds a new method to the ordersync service called `PeriodicallyGetOrders` which simply calls `GetOrders` on a regular interval (~~currently set to 5 minutes~~). Update: now set to 1 hour. See [comment below](https://github.com/0xProject/0x-mesh/pull/764#issuecomment-601440504).

This accounts for situations where we receive an order which is not valid now but may be valid in the future. There are a couple scenarios where this could happen. Here's just a few examples:

1. A block re-org causes filled or cancelled orders to be fillable again. While we do account for this kind of re-org by marking existing orders as removed before we ultimately delete them from the database, that doesn't account for orders that were unfillable _the first time_ we saw them. Those orders are currently never stored at all.
2. The maker may not have the appropriate balance/allowance when we first see the order, but that may change in a subsequent block. It is not uncommon for nodes in the network to be behind or ahead of each other by 1-2 blocks.
3. The maximum expiration time may be decreased, which means some orders that were previously considered invalid because they expire too far in the future would now be valid.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -67,7 +67,7 @@ const (
 	// ordersyncApproxDelay is the approximate amount of time to wait between each
 	// run of the ordersync protocol (as a requester). We always request orders
 	// immediately on startup. This delay only applies to subsequent runs.
-	ordersyncApproxDelay         = 5 * time.Minute
+	ordersyncApproxDelay         = 1 * time.Hour
 	paginationSubprotocolPerPage = 500
 )
 
--- a/core/core.go
+++ b/core/core.go
@@ -662,6 +662,12 @@ func (app *App) Start(ctx context.Context) error {
 		defer func() {
 			log.Debug("closing ordersync service")
 		}()
+		log.WithFields(map[string]interface{}{
+			"approxDelay":  ordersyncApproxDelay,
+			"perPage":      paginationSubprotocolPerPage,
+			"subprotocols": []string{"FilteredPaginationSubProtocol"},
+		}).Info("starting ordersync service")
+
 		if err := app.ordersyncService.PeriodicallyGetOrders(innerCtx, ordersyncMinPeers, ordersyncApproxDelay); err != nil {
 			orderSyncErrChan <- err
 		}
--- a/core/core.go
+++ b/core/core.go
@@ -63,7 +63,11 @@ const (
 	version          = "development"
 	// ordersyncMinPeers is the minimum amount of peers to receive orders from
 	// before considering the ordersync process finished.
-	ordersyncMinPeers            = 5
+	ordersyncMinPeers = 5
+	// ordersyncMinDelay is the amount of time to wait between each run of the
+	// ordersync protocol (as a requester). We always request orders immediately
+	// on startup. This delay only applies to subsequent runs.
+	ordersyncMinDelay            = 5 * time.Minute
 	paginationSubprotocolPerPage = 500
 )
 
@@ -658,7 +662,7 @@ func (app *App) Start(ctx context.Context) error {
 		defer func() {
 			log.Debug("closing ordersync service")
 		}()
-		if err := app.ordersyncService.GetOrders(innerCtx, ordersyncMinPeers); err != nil {
+		if err := app.ordersyncService.PeriodicallyGetOrders(innerCtx, ordersyncMinPeers, ordersyncMinDelay); err != nil {
 			orderSyncErrChan <- err
 		}
 	}()
--- a/core/core.go
+++ b/core/core.go
@@ -64,10 +64,10 @@ const (
 	// ordersyncMinPeers is the minimum amount of peers to receive orders from
 	// before considering the ordersync process finished.
 	ordersyncMinPeers = 5
-	// ordersyncMinDelay is the amount of time to wait between each run of the
-	// ordersync protocol (as a requester). We always request orders immediately
-	// on startup. This delay only applies to subsequent runs.
-	ordersyncMinDelay            = 5 * time.Minute
+	// ordersyncApproxDelay is the approximate amount of time to wait between each
+	// run of the ordersync protocol (as a requester). We always request orders
+	// immediately on startup. This delay only applies to subsequent runs.
+	ordersyncApproxDelay         = 5 * time.Minute
 	paginationSubprotocolPerPage = 500
 )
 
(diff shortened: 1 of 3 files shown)


core/core.go
```diff
@@ -658,7 +662,7 @@ func (app *App) Start(ctx context.Context) error {
 		defer func() {
 			log.Debug("closing ordersync service")
 		}()
-		if err := app.ordersyncService.GetOrders(innerCtx, ordersyncMinPeers); err != nil {
+		if err := app.ordersyncService.PeriodicallyGetOrders(innerCtx, ordersyncMinPeers, ordersyncMinDelay); err != nil {
```
jalextowle: How would you feel about including a log like "started the ordersync service" here? This would make this more consistent with the other goroutines that are created in this function.

core/ordersync/ordersync.go
```diff
@@ -340,6 +340,29 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
 	return nil
 }
 
+// PeriodicallyGetOrders periodically calls GetOrders. It waits a minimum of
+// minDelay between each call. It will block until there is a critical error or
+// the given context is canceled.
+func (s *Service) PeriodicallyGetOrders(ctx context.Context, minPeers int, minDelay time.Duration) error {
```
jalextowle: How hard would it be to add a test for this? FWIW this is pretty simple, so I don't think it's absolutely necessary, but if it's easy, I'd be in favor of adding it to prevent regressions.

core/core.go
albrow: Good idea. Just added.

core/ordersync/ordersync.go
albrow: Talked to Fabio about this too. I think the code in question is simple enough that our time would be better spent elsewhere. However, I did add a test for the new `calculateDelayWithJitter` since it's a little less straightforward.

albrow:
After running some tests with Mesh Manager, we decided to increase the delay between ordersync requests from 5 minutes to 1 hour. We were seeing a lot more bandwidth usage than we expected, so reducing the frequency will help counter that. This change means that this PR now realistically only addresses point (3) and doesn't help much with (1) or (2). We feel that (3) is the most likely to occur and (1) and (2) should be rare in practice (at least for now).

We can potentially reduce the delay in the future if/when we introduce new subprotocols for ordersync that use less bandwidth. This will become more important if we start seeing (1) or (2) more often in the wild.

core/ordersync/ordersync.go
jalextowle: I think that's a fair assessment. Thanks for adding the other test.

core/ordersync/ordersync.go
```diff
@@ -39,6 +39,12 @@ const (
 	maxRequestsPerSecond = 30
 	// requestsBurst is the maximum number of requests to allow at once.
 	requestsBurst = 10
+	// ordersyncJitterAmount is the amount of random jitter to add to the delay before
+	// each run of ordersync in PeriodicallyGetOrders. It is bound by:
+	//
+	//    (approxDelay - approxDelay * jitter) <= actualDelay <= (approxDelay + approxDelay * jitter)
```
jalextowle: I don't feel strongly about this suggestion, but I think this could be a bit clearer if it is written:

```
approxDelay * (1 - jitter) <= actualDelay <= approxDelay * (1 + jitter)
```

My brain pattern matches this faster because it's what I would expect to see, but it's simple, so I'll leave this to your discretion.

core/ordersync/ordersync.go
```diff
@@ -340,6 +346,39 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		if err := s.GetOrders(ctx, minPeers); err != nil {
+			return err
+		}
+
+		// Note(albrow): The random jitter here helps smooth out the frequency of ordersync
```
jalextowle: Did we write tests in Mesh Manager to verify this hypothesis, or is this just speculation? I could definitely see this being the case; however, I also think it's possible that different jitter values would improve this even more, and I also think it's possible that in some circumstances a jitter of 0 would be best.

core/ordersync/ordersync.go
```diff
@@ -340,6 +346,39 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(delay):
+		}
+	}
+}
+
+func calculateDelayWithJitter(approxDelay time.Duration, jitterAmount float64) time.Duration {
+	jitterBounds := int(float64(approxDelay) * jitterAmount * 2)
+	delta := rand.Intn(jitterBounds) - jitterBounds/2
```
jalextowle: From https://golang.org/pkg/math/rand/#NormFloat64, `0 <= rand.Intn(jitterBounds) < jitterBounds`. Given this, the bounds given in the comment above `ordersyncJitterAmount` should be

```
approxDelay * (1 - jitter) <= actualDelay < approxDelay * (1 + jitter)
```
Unfortunately, `jitter + math.SmallestNonzeroFloat{32,64} == jitter` in most cases because the floating point numbers are too many orders of magnitude different, so it could be tricky to make this inclusive. I'd say that we just update the comment.

core/ordersync/ordersync.go
albrow: I like this better too. Just updated it.

core/ordersync/ordersync.go
albrow: This idea is based on one simple assumption: when we re-deploy or update Mesh nodes we tend to do them all at once. This is not just the case for our infra; it's probably true for any other team that runs more than one node. If each of those nodes always used the same exact interval, it would lead to spiky behavior where the network as a whole sees a lot of ordersync requests coming in at the same time. Adding some random jitter makes this a lot less likely.

We can't really measure it in Mesh Manager right now because Mesh Manager tests only run for 15 minutes and the new interval is set to 1 hour. That said, the assumption seems really straightforward to me and at the very least adding a random jitter is unlikely to cause any harm.

core/ordersync/ordersync.go
albrow: I just made the comment more accurate by changing the the upper bound to exclusive. There's no inherent reason to prefer an inclusive upper bound.

core/ordersync/ordersync.go
jalextowle: I thought that was the assumption that you were using. I can see this being true of the network today, but I suspect that it will not always be the case.

I didn't think that it was likely to cause harm either, hence the "non-blocking" preface.

core/ordersync/ordersync.go
jalextowle: Thanks, yep I agree that this is just as good as an inclusive upper bound.

core/ordersync/ordersync.go
jalextowle: I thought that was probably the assumption that you were using. Thanks for clarifying that for me.

PR #765, opened 2020-03-20
Update ganache addresses

fabioberger:


opaolini:
LGTM if the tests pass, address taken from https://github.com/0xProject/0x-monorepo/blob/development/packages/contract-addresses/addresses.json#L145

PR #768, opened 2020-03-24
Add custom order filters documentation

fabioberger:

### DIFF
--- a/docs/custom_order_filters.md
+++ b/docs/custom_order_filters.md
@@ -1,0 +1,58 @@
+# Custom Order Filters
+
+Mesh supports the creation of separate sub-networks where 0x orders that adhere to a specific schema are shared. Each sub-network is built around a custom order filter. The custom filter defines which orders are allowed to be shared within a sub-network. For example:
+
+- All orders for a specific asset pair (e.g., ETH/DAI)
+- All orders for non-fungibles (i.e., ERC721, ERC1155)
+- All orders used by a specific DApp
+
+A custom filter may be passed into Mesh as a [JSON Schema](https://json-schema.org/) via the `CUSTOM_ORDER_FILTER` environment variable. Messages that contain orders that don't match this schema will be dropped. As a limitation, filtering is only possible by looking at the static fields of an order. So for example, it is not possible to filter orders by doing an on-chain check or sending an HTTP request to a third-party API. We don't expect that this limitation is going to be a problem in practice and it comes with the huge benefit of enabling cross-topic forwarding in the future (more on that later).
+
+### New order and message schemas.
+
+All orders must match the following JSON Schema:
+
+```json
+{
+	"id": "/rootOrder",
+	"allOf": [{
+		"$ref": "/customOrder"
+	}, {
+		"$ref": "/signedOrder"
+	}]
+}
+```
+
+- `/signedOrder` is the JSON Schema that will match any valid 0x orders.
+- `/customOrder` is the custom schema passed in through the `CUSTOM_ORDER_FILTER` environment variable.
+
+Organizing the JSON Schema for orders like this means that `CUSTOM_ORDER_FILTER` can be relatively small. It doesn't need to contain all the required fields for a signed 0x order. It just needs to contain any _additional_ requirements on top of the default ones.
+
+#### Example custom order schemas
+
+The following `CUSTOM_ORDER_FILTER` doesn't add any additional requirements. All valid signed 0x orders will be accepted. This is the default value if no custom filter is passed in.
+
+```json
+{}
+```
+
+The following `CUSTOM_ORDER_FILTER` matches any valid signed 0x orders with a specific sender address:
+
+```json
+{
+	"properties": {
+		"senderAddress": {
+			"pattern": "0x00000000000000000000000000000000ba5eba11",
+			"type": "string"
+		}
+	}
+}
+```
+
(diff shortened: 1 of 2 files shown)


docs/custom_order_filters.md
```diff
@@ -0,0 +1,58 @@
+# Custom Order Filters
+
+Mesh supports the creation of separate sub-networks where 0x orders that adhere to a specific schema are shared. Each sub-network is built around a custom order filter. The custom filter defines which orders are allowed to be shared within a sub-network. For example:
+
+- All orders for a specific asset pair (e.g., ETH/DAI)
```
jalextowle: Non-blocking suggestion: It would probably be helpful to add explicit examples for the filters described here. I imagine that these will be used a lot.

docs/custom_order_filters.md
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+This can easily be tweaked to filter orders by asset type, maker/taker address, or fee recipient. JSON Schema has support for [regular expressions](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html) allowing for partial matching of any 0x order field.
+
+### Limitations
+
+Nodes that are spun up with a custom filter will share all their orders with nodes that are either using the same filter or the default "all" filter. They will not share orders with nodes using other custom filters since each filter results in a separate sub-network. Therefore, custom filters are most useful for applications where users care about a distinct sub-set of 0x orders. 
+
+If you wanted to connect two sub-networks with overlapping valid orders, you could spin up a Mesh node for each and additionally run a [bridge script](https://github.com/0xProject/0x-mesh/blob/master/cmd/mesh-bridge/main.go) to sends orders from one sub-network to the other. Longer term, we hope to add support for cross-topic forwarding which will allow Mesh nodes to do this under-the-hood.
```
jalextowle: I think this would be clearer if we said `for each sub-network` so that there is not ambiguity about whether you are referring to the valid orders or the sub-networks. It's up to you though.

Nit: `to sends` => `to send`

Also, I think we need a comment before `which` in the last sentence. I double checked with grammarly: https://www.grammarly.com/blog/comma-before-which/

docs/custom_order_filters.md
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+}
+```
+
+This can easily be tweaked to filter orders by asset type, maker/taker address, or fee recipient. JSON Schema has support for [regular expressions](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html) allowing for partial matching of any 0x order field.
+
+### Limitations
+
+Nodes that are spun up with a custom filter will share all their orders with nodes that are either using the same filter or the default "all" filter. They will not share orders with nodes using other custom filters since each filter results in a separate sub-network. Therefore, custom filters are most useful for applications where users care about a distinct sub-set of 0x orders.
```
albrow:
```suggestion
Nodes that are spun up with a custom filter will share all their orders with nodes that are either using the exact same filter or the default "all" filter. They will _not_ share orders with nodes using different custom filters (even if a given order matches both filters) because each filter results in a separate sub-network. Therefore, custom filters are most useful for applications where users care about a distinct subset of 0x orders.
```

albrow: Some optional nits here but the main thing I want to draw attention to is that if two nodes are using different order filters, they will not share orders with one another, even if some orders match both filters.

Depending on how much detail we want to go into here, this is _technically_ only partly true. The current implementation of `ordersync` _does_ allow providers to share orders with requesters even if they are using different filters.

docs/custom_order_filters.md
albrow: Seconded.

docs/custom_order_filters.md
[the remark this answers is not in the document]
jalextowle: I am un-resolving this because none of these comments were addressed.

docs/custom_order_filters.md
jalextowle: Ditto.

docs/custom_order_filters.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Limitations
+
+Nodes that are spun up with a custom filter will share all their orders with nodes that are either using the exact same filter or the default "all" filter (i.e., "{}"). They will _not_ share orders with nodes using different custom filters (even if a given order matches both filters) because each filter results in a separate sub-network. Therefore, custom filters are most useful for applications where users care about a distinct subset of 0x orders.
+
+If you wanted to connect two sub-networks with overlapping valid orders, you could spin up a Mesh node for each sub-network and additionally run a [bridge script](https://github.com/0xProject/0x-mesh/blob/master/cmd/mesh-bridge/main.go) to send orders from one sub-network to the other. Longer term, we hope to add support for cross-topic forwarding, which will allow Mesh nodes to do this under-the-hood.
+
+### Examples
+
+##### WETH <-> DAI orders:
```
jalextowle: This is great! Two (completely optional) suggestions would be to (1) mention what network these addresses are on just in case someone uses these addresses on the wrong network and (2) mention these examples in the top section. Having said this, these examples really improve the documentation, and I think it's looking really good.

docs/custom_order_filters.md
```diff
@@ -0,0 +1,131 @@
...
+	}]
+}
+```
+
+- `/signedOrder` is the JSON Schema that will match any valid 0x orders.
+- `/customOrder` is the custom schema passed in through the `CUSTOM_ORDER_FILTER` environment variable.
+
+Organizing the JSON Schema for orders like this means that `CUSTOM_ORDER_FILTER` can be relatively small. It doesn't need to contain all the required fields for a signed 0x order. It just needs to contain any _additional_ requirements on top of the default ones.
+
+#### Example custom order schemas
```
jalextowle: Should this be in the `Examples` section?

PR #771, opened 2020-03-30
Fully Automate Document Generation

jalextowle:
## Description
Currently document generation is a somewhat manual process because of the fact that we generate docs for all exported fields (even if they aren't exported at the package level) and because TypeDoc `v0.15.0` does not document functions. This PR fixes this by adding a small script to `cut-release`.

## Testing Instructions
```
git checkout docs/automate-doc-generation
VERSION=100.0.0 go run cmd/cut-release/main.go

# Don't actually run this :p
behold "the glory of automated doc generation"
```

### DIFF
--- a/cmd/cut-release/main.go
+++ b/cmd/cut-release/main.go
@@ -13,39 +13,7 @@ import (
 	"github.com/plaid/go-envvar/envvar"
 )
 
-var functionDocsTemplate = `
-# Functions
-
-## loadMeshStreamingWithURLAsync
-▸ **loadMeshStreamingWithURLAsync**(` + "`" + `url` + "`" + `: ` + "`" + `string` + "`" + `): *Promise‹` + "`" + `void` + "`" + `›*
-
-*Defined in [index.ts:7](https://github.com/0xProject/0x-mesh/blob/%s/packages/browser-lite/src/index.ts#L7)*
-
-Loads the Wasm module that is provided by fetching a url.
-
-**Parameters:**
-
-Name | Type | Description |
------- | ------ | ------ |
-` + "`" + `url` + "`" + ` | ` + "`" + `string` + "`" + ` | The URL to query for the Wasm binary |
-
-<hr />
-
-## loadMeshStreamingAsync
-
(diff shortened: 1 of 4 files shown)


cmd/cut-release/main.go
```diff
@@ -5,13 +5,16 @@ import (
...
 	"log"
+	"os"
 	"os/exec"
 	"regexp"
 	"strings"
 
 	"github.com/plaid/go-envvar/envvar"
 )
 
+var functionDocsTemplate = "\n# Functions\n\n## loadMeshStreamingForURLAsync\n▸ **loadMeshStreamingWithURLAsync**(`url`: `string`): *Promise‹`void`›*\n\n*Defined in [index.ts:7](https://github.com/0xProject/0x-mesh/blob/%s/packages/browser-lite/src/index.ts#L7)*\n\nLoads the Wasm module that is provided by fetching a url.\n\n**Parameters:**\n\nName | Type | Description |\n------ | ------ | ------ |\n`url` | `string` | The URL to query for the Wasm binary |\n\n<hr />\n\n## loadMeshStreamingAsync\n\n▸ **loadMeshStreamingAsync**(`response`: `Response | Promise<Response>`): *Promise‹`void`›*\n\n*Defined in [index.ts:15](https://github.com/0xProject/0x-mesh/blob/%s/packages/browser-lite/src/index.ts#L15)*\n\nLoads the Wasm module that is provided by a response.\n\n**Parameters:**\n\nName | Type | Description |\n------ | ------ | ------ |\n`response` | `Response &#124; Promise<Response>` | The Wasm response that supplies the Wasm binary |\n\n<hr />"
```
albrow: I think it would be cleaner to use a raw string literal (i.e., backticks) here? Then you could include literal line breaks instead of `/n`. You would have to escape the backticks in this case, but imo it will still be more readable overall.

jalextowle: I thought about that, but there are a lot of backticks. I really wish that Golang would allow us to use backticks in raw strings (I'm really starting to dislike some of their language choices -- these issues provide some context on why you can't just escape backticks in raw strings https://github.com/golang/go/issues/18221 and https://github.com/golang/go/issues/24475). 

I'll change this and then we can decide which version is cleaner

albrow: Forgot that you couldn't escape backticks inside a raw string literal without breaking it up into separate strings like this. I think this is easier to read but will likely be harder to change. I say we go back to the way it was before.

PR #777, opened 2020-04-02
expirationwatch: prune orders expiring at exactly block timestamp

fabioberger:
Fix bug causing orders expiring at _exactly_ the next block timestamp from not being pruned until the next block is mined.


fabioberger:
Duplicate.

Issue #706, opened 2020-02-06
orderwatch: order rejected with OrderUnfunded, but it is funded


fabioberger:
There is currently a pretty bad footgun in how Mesh works internally that can lead to orders being rejected with status `OrderUnfunded` despite the allowances and balances being correctly set. What leads to this incongruity is the fact that Mesh is periodically syncing up to the latest block, and if it has not yet performed this sync, it continues to validate received orders at the previous latest block height it knows above. If the maker's balance / allowance has been updated in a subsequent block, Mesh doesn't know about those changes yet and therefore evaluates the order as unfunded.

A simple solution to this problem is to give Mesh some time to process the latest block after setting an allowance / balance. This can be done by introducing a `sleep()` in tests and by reducing the `BLOCK_POLLING_INTERVAL` so that the sleep doesn't need to be very long.

Ideally, Mesh would always validate orders at the `latest` block height. To do this would require updating the order watching logic from it's current "stop the world" design, an initiative that is being tracked here: https://github.com/0xProject/0x-mesh/issues/625

albrow:
If we just want to tackle the footgun in our tests, we may be able to draw from the solution in __order_watcher_test.go__. Instead of `time.Sleep`, these tests call `blockWatcher.SyncToLatestBlock()` followed by `orderWatcher.WaitForAtLeastOneBlockToBeProcessed(ctx)` (e.g. look inside [`setupOrderWatcher`](https://github.com/0xProject/0x-mesh/blob/8afe758d075e6290aef3a86f56fbde3e9463cb0f/zeroex/orderwatch/order_watcher_test.go#L1583)).

PR #784, opened 2020-04-08
scenario: Add ability to create multiple orders in a single batch

albrow:
This PR is a follow-up on #779 and adds support for creating more than one order in a single batch via the `NewSignedTestOrdersBatch` function. This is a huge performance optimization because we only set the required balances and allowances once instead of doing it (and waiting for the corresponding transaction to be mined) for each order one at a time. This is not just optimization for the sake of optimization; it will enable important new test cases by making it practical to create hundreds of orders at a time (see https://github.com/0xProject/0x-mesh/pull/760 for an example of how this limitation has blocked us in the past).

### DIFF
--- a/integration-tests/rpc_integration_test.go
+++ b/integration-tests/rpc_integration_test.go
@@ -127,12 +127,9 @@ func runGetOrdersTest(t *testing.T, rpcEndpointPrefix, rpcServerType string, rpc
 	require.NoError(t, err)
 
 	// Create 10 new valid orders.
-	// TODO(albrow): Update this when scenario supports creating orders in a batch.
 	numOrders := 10
-	signedTestOrders := make([]*zeroex.SignedOrder, numOrders)
-	for i := 0; i < numOrders; i++ {
-		signedTestOrders[i] = scenario.NewSignedTestOrder(t, orderopts.SetupMakerState(true))
-	}
+	orderOptions := scenario.OptionsForAll(orderopts.SetupMakerState(true))
+	signedTestOrders := scenario.NewSignedTestOrdersBatch(t, numOrders, orderOptions)
 	// Creating a valid order involves transferring sufficient funds to the maker, and setting their allowance for
 	// the maker asset. These transactions must be mined and Mesh's BlockWatcher poller must process these blocks
 	// in order for the order validation run at order submission to occur at a block number equal or higher then
--- a/scenario/scenario.go
+++ b/scenario/scenario.go
@@ -96,6 +96,72 @@ func NewSignedTestOrder(t *testing.T, opts ...orderopts.Option) *zeroex.SignedOr
 	return signedOrder
 }
 
+// NewSignedTestOrdersBatch effeciently creates numOrders orders with independent options.
+// If the options require setting up maker or taker state, that state will be set up effeciently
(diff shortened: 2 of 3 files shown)


integration-tests/rpc_integration_test.go
```diff
@@ -127,12 +127,9 @@ func runGetOrdersTest(t *testing.T, rpcEndpointPrefix, rpcServerType string, rpc
...
 
 	// Create 10 new valid orders.
-	// TODO(albrow): Update this when scenario supports creating orders in a batch.
 	numOrders := 10
-	signedTestOrders := make([]*zeroex.SignedOrder, numOrders)
-	for i := 0; i < numOrders; i++ {
-		signedTestOrders[i] = scenario.NewSignedTestOrder(t, orderopts.SetupMakerState(true))
-	}
+	orderOptions := scenario.OptionsForAll(orderopts.SetupMakerState(true))
+	signedTestOrders := scenario.NewSignedTestOrdersBatch(t, numOrders, orderOptions)
```
jalextowle, line 132: Would it make sense to encapsulate the `time.Sleep` call in `NewSignedTestOrder` and `NewSignedTestOrdersBatch`?

scenario/scenario.go
```diff
@@ -96,6 +96,72 @@ func NewSignedTestOrder(t *testing.T, opts ...orderopts.Option) *zeroex.SignedOr
...
+		// Apply the options (if any) for the order we will create at this index.
+		cfg := defaultConfig()
+		if optionsForIndex != nil {
+			opts := optionsForIndex(i)
+			if opts != nil {
+				require.NoError(t, cfg.Apply(opts...))
+			}
+		}
+
+		// Crate the order based on the cfg.
```
jalextowle, line 121: Nit: `Crate` => `Create`

scenario/scenario.go
```diff
@@ -96,6 +96,72 @@ func NewSignedTestOrder(t *testing.T, opts ...orderopts.Option) *zeroex.SignedOr
...
+			takerBalances, found := allRequiredBalances[cfg.SetupTakerAddress]
+			if !found {
+				allRequiredBalances[cfg.SetupTakerAddress] = takerBalancesForThisOrder
+			} else {
+				takerBalances.add(takerBalancesForThisOrder)
+			}
+		}
+	}
+
+	// Setup all the rquired balances.
```
jalextowle, line 148: Nit: `rquired` => `required`

integration-tests/rpc_integration_test.go
albrow: It's not quite that simple. There are some places where we don't need to call `time.Sleep` and have a more efficient solution outside of the `scenario` package (e.g. most of the tests in __order_watcher_test.go__). There is probably room for improvement here and we might be able to take some parts of the solution from __order_watcher_test.go__ and use them everywhere with a few tweaks, but that is probably best suited for a separate PR.

albrow: This is already being tracked in https://github.com/0xProject/0x-mesh/issues/706.

jalextowle: Okay, that makes sense to me. Thanks for the explanation.

PR #789, opened 2020-04-16
Release version 9.3.0

albrow:


constants/constants.go
```diff
@@ -58,7 +58,7 @@ var GanacheAccountToPrivateKey = map[common.Address][]byte{
 var GanacheDummyERC721TokenAddress = common.HexToAddress("0x07f96aa816c1f244cbc6ef114bb2b023ba54a2eb")
 
 // GanacheDummyERC1155MintableAddress is the dummy ERC1155 token address in the Ganache snapshot
-var GanacheDummyERC1155MintableAddress = common.HexToAddress("0x8d42e38980ce74736c21c059b2240df09958d3c8")
+var GanacheDummyERC1155MintableAddress = common.HexToAddress("0x038f9b392fb9a9676dbaddf78ea5fdbf6c7d9710")
```
jalextowle, line 61: Was this wrong before? I wasn't aware that we changed it's location in the migration script.

docs/browser-bindings/browser-lite/reference.md
```diff
@@ -2067,14 +2068,14 @@ Name | Type | Description |
...
+*Defined in [index.ts:15](https://github.com/0xProject/0x-mesh/blob/8df8910/packages/browser-lite/src/index.ts#L15)*
 
 Loads the Wasm module that is provided by a response.
 
 **Parameters:**
 
 Name | Type | Description |
 ------ | ------ | ------ |
-`response` | `Response or Promise<Response>` | The Wasm response that supplies the Wasm binary |
+`response` | `Response &#124; Promise<Response>` | The Wasm response that supplies the Wasm binary |
```
jalextowle, line 2079: 🤞 that this renders correctly.

constants/constants.go
fabioberger: It did change with the updated snapshot we're now using (`6.2.4`). Prior to this version, we were on version `5.1.0`. It must have changed sometime between those two versions.

PR #809, opened 2020-05-19
Optimize `MatchOrder` and `ValidateOrder` using `ajv`

jalextowle:


orderfilter/validate_js.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (f *Filter) MatchOrderMessageJSON(messageJSON []byte) (bool, error) {
+	//	if !messageValidatorLoaded {
+	//		loaded := checkForMessageValidator()
+	//		if !loaded {
+	//			return false, errors.New("message validator not loaded")
+	//		}
+	//	}
+
+	jsResult := js.Global().Get("schemaValidator").Call("messageValidator", js.ValueOf(string(messageJSON)))
```
albrow, line 101: If `ajv` accepts `Uint8Array` or `Buffer`, it might be more efficient to use [`CopyBytesToGo`](https://golang.org/pkg/syscall/js/#CopyBytesToGo) and/or [`CopyBytesToJS`](https://golang.org/pkg/syscall/js/#CopyBytesToJS) instead of casting to/from `string`. Worth checking.

orderfilter/validate_js.go
```diff
@@ -0,0 +1,130 @@
...
+// This libarary is used to increase the performance of Mesh nodes that run in the browser.
+func (f *Filter) ValidateOrderJSON(orderJSON []byte) (*SchemaValidationResult, error) {
+	//	if !orderValidatorLoaded {
+	//		loaded := checkForOrderValidator()
+	//		if !loaded {
+	//			return nil, errors.New("order validator not loaded")
+	//		}
+	//	}
+
+	jsResult := js.Global().Get("schemaValidator").Call("orderValidator", js.ValueOf(string(orderJSON)))
```
albrow, line 68: `js.Global().Get("schemaValidator")` could return `js.Null` or `js.Undefined` here right? I think we should either return an error if that happens or fallback to the less efficient implementation. Also instead of checking inside of every method, we could check once and encapsulate it via a field on `Filter`.

packages/browser-lite/src/wasm_exec.ts
```diff
@@ -12,6 +11,7 @@
  * NOTE(jalextowle): This comment must be here so that typedoc knows that the above
  * comment is a module comment
  */
+/* tslint:disable */
```
albrow, line 14: Maybe I'm missing something but I don't think this change is necessary.

Based on https://palantir.github.io/tslint/usage/rule-flags/:

> /* tslint:disable */ - Disable all rules for the rest of the file


We don't need to move the `tslint:disable` comment nor do we need to add `tslint:enable` at the end of the file.

test-wasm/node_shim.js
```diff
@@ -1,12 +1,146 @@
 // This file adds RTCPeerConnection to the global context, making Node.js more
 // closely match the browser API for WebRTC.
 
+const ajv = require("ajv");
```
albrow, line 4: Is there another place where we can add this? __node_shim.js__ is specifically about adding APIs that exist in the browser but not in Node.js. We have not used it so far to pull in third-party libraries outside of that context. For example, we don't load in `BrowserFS` here.

If there is no other place, let's update the comment so that the purpose of this file is clear.

packages/browser-lite/src/mesh.ts
```diff
@@ -129,6 +129,81 @@ window.addEventListener(loadEventName, () => {
...
+const addressSchema = {
+    $id: 'http://example.com/address',
+    type: 'string',
+    pattern: '^0x[0-9a-fA-F]{40}$',
+};
+const wholeNumberSchema = {
+    $id: 'http://example.com/wholeNumber',
+    anyOf: [{ type: 'string', pattern: '^\\d+$' }, { type: 'integer' }],
+};
+const hexSchema = { $id: 'http://example.com/hex', type: 'string', pattern: '^0x(([0-9a-fA-F][0-9a-fA-F])+)?$' };
```
albrow, line 141: Do we have to change the `$id` fields here compared to the standalone Go implementation? We specifically chose an order for compiling the JSON schema such that common types like `/address` and `/wholeNumber` are exposed and can be used in custom order filters. Changing this technically breaks backwards compatibility and breaks compatibility between the standalone and browser nodes.

orderfilter/validate_js.go
jalextowle: From what I can tell, it doesn't. I'll experiment with this a little bit more (in a simpler example), and see what happens.

packages/browser-lite/src/wasm_exec.ts
jalextowle: I was just moving it because the `NOTE` is now in a better location. I removed the `tslint:enable` though

orderfilter/validate_js.go
albrow: No worries, don't spend anymore time on it if we already have something fast enough.

packages/browser-lite/src/mesh.ts
jalextowle: It turned out that we didn't! I needed to use `addSchema` instead of passing them into the constructor, but it appears to work like a charm now.

orderfilter/validate_js.go
jalextowle: Sounds good. I'll probably try it out this weekend, but unless I read the `ajv` source incorrectly (it's pretty opaque, so that's completely possible), it validates the data "as-is" without any other processing.

orderfilter/validate_test.go
```diff
@@ -246,6 +248,7 @@ func TestFilterValidateOrderJSON(t *testing.T) {
 		loop:
 			for _, expectedErr := range tc.expectedErrors {
 				for _, actualErr := range actualResult.Errors() {
+					fmt.Printf("%+v\n", actualErr)
```
albrow, line 251: I think you meant to remove this?

packages/browser-lite/src/mesh.ts
```diff
@@ -129,6 +129,9 @@ window.addEventListener(loadEventName, () => {
     isWasmLoaded = true;
 });
 
+const schemaValidator: any = {};
```
albrow, line 132: This creates a race condition. Here, `window.schemaValidator` is defined before `schemaValidator.orderValidator` and `schemaValidator.messageValidator` are set. In __validate_js.go__, we only check that `window.schemaValidator` is not `null` or `undefined` or `null` and then we immediately call `orderValidator` and `messageValidator`, without checking if _those_ are `undefined` or `null`.

It also creates additional complexity since there are two ways of setting `window.schemaValidator`, one here and one in __browser_shim.ts__. That can potentially cause bugs down the road if we change one but forget to change the other.

Here's what I think we should do instead:

1. Move `setSchemaValidator` from __browser_shim.ts__ into __schema_validator.ts__. __browser_shim.ts__ should just import the existing function and assign it to `window`.
2. Use the same function (i.e. `setSchemaValidator`) whether we are loading the schema validator in `@0x/mesh-browser`, `@0x/mesh-browser-lite`, or for our tests.
3. Eliminate the race condition by not setting `window.schemaValidator` before setting `orderValidator` and `messageValidator`.

Step 3 might require the addition of a new global variable to indicate to the Go code that the schema validator will be loaded (e.g. similar to what we did for BrowserFS). Or better yet, we could restructure things so that the schema validator is passed through to Go code via `core.New`.

albrow: Also the type of `schemaValidator` should be `SchemaValidator`, not `any`.

```suggestion
const schemaValidator: SchemaValidator = {};
```

Or depending on what you do, it might need to be `SchemaValidator | null`.

packages/browser-lite/src/schema_validator.ts
```diff
@@ -0,0 +1,166 @@
...
+            }
+        } catch (error) {
+            result.fatal = JSON.stringify(error);
+        }
+        return result;
+    };
+}
+
+const addressSchema = {
+    $async: false,
```
albrow, line 90: I think we can remove `$async: false` everywhere now. AFACT from looking at the `ajv` docs, it doesn't actually do anything unless you call async methods like `compileAsync`, which we are not.

Eventually, I would also like to use one single canonical source for our JSON Schemas to eliminate potential bugs/headaches down the road. For example, we could use the Go definitions and pass them through to `ajv` (e.g. by calling `ajv.addSchema` from Go). But if you like we can do that in a separate PR.

packages/webpack-example/src/index.ts
```diff
@@ -52,6 +52,8 @@ import { BigNumber, Mesh, OrderEvent, SignedOrder, SupportedProvider } from '@0x
     // Add the order and log the result.
     const result = await mesh.addOrdersAsync([order]);
     console.log(result);
+
+    (window as any).getStatsAsync = mesh.getStatsAsync();
```
albrow, line 56: I think you meant to remove this?

repeated_test~
```diff
@@ -0,0 +1 @@
+#!/bin/bash
```
albrow, line 1: Please remove this file.

zeroex/order_test.go
```diff
@@ -111,5 +111,7 @@ func TestMarshalUnmarshalOrderEvent(t *testing.T) {
 	signedOrder.ResetHash()
 
 	require.NoError(t, json.NewDecoder(buf).Decode(&decoded))
-	assert.Equal(t, orderEvent, decoded)
+	// TODO(jalextowle): An `assert` here will allow wasm browser tests to fail regardless
+	// of their actual result. Change this to use `assert` once this has been fixed in `testify`
+	require.Equal(t, orderEvent, decoded)
```
albrow, line 116: I don't think this has anything to do with `testify` and rather is a problem with either our tests or `wasmbrowsertest`. We should be able to go back to using `assert` here after I added the workaround in https://github.com/0xProject/wasmbrowsertest/pull/1.

packages/browser-lite/src/schema_validator.ts
```diff
@@ -0,0 +1,166 @@
...
+    exchangeAddress: string,
+    customOrderFilter?: JsonSchema,
+): SchemaValidator {
+    const chainIdSchema = {
+        $async: false,
+        $id: '/chainId',
+        const: chainId,
+    };
+
+    // NOTE(jalextowle): The current implementation of `zeroex.SignedOrder.MarhsalJSON`
```
albrow, line 34: I don't think we should make this change here. It is potentially breaking depending on how people are interacting with Mesh.

I would not expect an `enum` with two values to result in a big performance hit anyways. Is that not the case?

orderfilter/validate.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func (f *Filter) ValidateOrderJSON(orderJSON []byte) (*jsonschema.Result, error) {
+	return f.orderSchema.Validate(jsonschema.NewBytesLoader(orderJSON))
+}
+
+// MatchOrder returns true if the order passes the filter. It only returns an
+// error if there was a problem with validation. For details about
+// orders that do not pass the filter, use ValidateOrder.
+func (f *Filter) MatchOrder(order *zeroex.SignedOrder) (bool, error) {
```
albrow, line 22: I think `MatchOrder` can be shared between Go and JS/Wasm? It just calls `ValidateOrder` and `reult.Valid`.

orderfilter/validate.go
```diff
@@ -0,0 +1,55 @@
...
+	return f.orderSchema.Validate(jsonschema.NewGoLoader(order))
+}
+
+// Dummy declaration to ensure that ValidatePubSubMessage matches the expected
+// signature for pubsub.Validator.
+var _ pubsub.Validator = (&Filter{}).ValidatePubSubMessage
+
+// ValidatePubSubMessage is an implementation of pubsub.Validator and will
+// return true if the contents of the message pass the message JSON Schema.
+func (f *Filter) ValidatePubSubMessage(ctx context.Context, sender peer.ID, msg *pubsub.Message) bool {
```
albrow, line 48: I think `ValidatePubSubMessage` can also be shared between Go and JS/Wasm. But let me know if I'm missing something.

albrow, line 48: I think `ValidatePubSubMessage` can also be shared between Go and JS/Wasm. But let me know if I'm missing something.

orderfilter/validate_js.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/0xProject/0x-mesh/zeroex"
+	peer "github.com/libp2p/go-libp2p-core/peer"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	orderValidatorLoaded   = false
+	messageValidatorLoaded = false
+	// validatorLoadCheckInterval is frequently to check whether the schema
```
albrow, line 22: These variables do not appear to be actually used.

repeated_test~
jalextowle: Sorry, this was an Emacs temporary file so I missed this. I'm going to go ahead and add another `.gitignore` line for these types of files

albrow: IMO this is the sort of thing that should go in your global __.gitignore__ not in the __.gitignore__ for Mesh. This is specific to your development setup, and it is not possible/realistic for us to account for every possible development environment in Mesh's __.gitignore__.

packages/browser-lite/src/schema_validator.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Sets up a schema validator on the window object.
+ * @param customOrderSchema The custom filter that will be used by the root
+ *        schemas.
+ * @param schemas These are all of the schemas that can be compiled before the
+ *        customOrderSchema.
+ * @param rootSchemas The root schemas. These must be compiled last.
+ */
+export function setSchemaValidator(customOrderSchemaString: string, schemas: string[], rootSchemas: string[]): void {
+    // NOTE(jalextowle): This causes `ajv` to look at `id` fields rather than
```
albrow, line 23: Is this comment still relevant? What is it referring to?

packages/browser-lite/src/schema_validator.ts
```diff
@@ -0,0 +1,67 @@
...
+        orderValidator: constructValidationFunctionWrapper(orderValidate),
+        messageValidator: constructValidationFunctionWrapper(messageValidate),
+    };
+}
+
+function constructValidationFunctionWrapper(
+    validationFunction: ajv.ValidateFunction,
+): (input: string) => SchemaValidationResult {
+    return (input: string) => {
+        const result: any = { success: false, errors: [] };
```
albrow, line 56: Can you do a find all search for `: any`? There should be no reason to use it except when assigning to window via `(window as any)`.

I tried making this change:

```suggestion
        const result = { success: false, errors: [] };
```

but it causes type errors due to the return type of `validationFunction`. I think this is a potential bug in the code. It's either that or the type definitions for `ajv.ValidateFunction` are wrong. In either case, setting the type to `any` hides the problem and should be avoided in general.

Screen Shot 2020-05-26 at 8 36 44 PM

orderfilter/validate_js_test.go
```diff
@@ -0,0 +1,38 @@
+// +build js, wasm
+
+package orderfilter
+
+var (
+	testFilterValidateOrderExpectedErrors = [][]string{
```
albrow, line 6: IMO This approach makes the tests too brittle. They will break if you add a new test case, change the order of test cases, or comment out some test cases. Instead of creating an implicit dependence on the ordering of test cases and the corresponding errors could we make each expected error a named variable? Something like:

```go
// +build !js

var (
    missingMakerAddressErrors = []string{
        "put the expected errors for Go here",
    }
)
```

```go
// +build js,wasm

var (
    missingMakerAddressErrors = []string{
        "put the expected errors for JavaScript/Wasm here",
    }
)
```

```diff
// In each test case, reference the expected errors by name...
{
	note:              "order with missing makerAddress",
	chainID:           constants.TestChainID,
	customOrderSchema: DefaultCustomOrderSchema,
	orderJSON:         []byte{ /* ... */ },
+	expectedErrors:    missingMakerAddressErrors,	
},
```

This way the tests will not break if you change the order of test cases, comment some out, or add new ones. And if you add the expected errors for one environment but not the other, the compiler will yell at you.

orderfilter/filter_js.go
```diff
@@ -0,0 +1,46 @@
...
+	chainID              int
+	rawCustomOrderSchema string
+}
+
+func New(chainID int, customOrderSchema string, contractAddresses ethereum.ContractAddresses) (*Filter, error) {
+	chainIDSchema := fmt.Sprintf(`{"$id": "/chainId", "const":%d}`, chainID)
+	exchangeAddressSchema := fmt.Sprintf(`{"$id": "/exchangeAddress", "enum":[%q,%q]}`, contractAddresses.Exchange.Hex(), strings.ToLower(contractAddresses.Exchange.Hex()))
+	// NOTE(jalextowle): The order of the schemas within the two arrays
+	// defines their order of compilation.
+	js.Global().Call(
```
albrow, line 25: We should fail fast and return an error if `setSchemaValidator` is null or undefined. This can happen if we didn't load the JavaScript correctly or are using the wrong version of a dependency somewhere.

orderfilter/filter_js.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+	"strings"
+	"syscall/js"
+
+	"github.com/0xProject/0x-mesh/ethereum"
+)
+
+type Filter struct {
+	validatorLoaded      bool
```
albrow, line 14: I think this field, along with the `checkForValidator` function are no longer necessary because we are setting the validator directly by calling `setSchemaValidator`?

This also brings up the question, why do we need to set the value on `window` at all? Could we simply return the validator from `setSchemaValidator` and store it as a `js.Value` field on `Filter`? If so we should probably rename `setSchemaValidator` to something like `createSchemaValidator` or use a JavaScript constructor directly by calling `(js.Value).New`.

jalextowle: Good point 👍

packages/browser-lite/src/schema_validator.ts
jalextowle: I was able to address several areas where we use `as any` in `packages/`. There are still several lingering `as any` casts in the `@0x/mesh-rpc-client` package. I'll address these in a follow-up PR if you're alright with this.

orderfilter/filter_js.go
```diff
@@ -38,8 +45,17 @@ func New(chainID int, customOrderSchema string, contractAddresses ethereum.Contr
 			rootOrderSchema,
 			rootOrderMessageSchema,
 		})
+	orderValidator := schemaValidator.Get("orderValidator")
```
albrow, line 48: Much cleaner 😄

PR #811, opened 2020-05-20
Added `initFile` to `test-wasm-browser`

jalextowle:


core/ordersync/ordersync_test.go
```diff
@@ -1,12 +1,19 @@
...
 
 import (
+	"flag"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 )
 
+// NOTE(jalextowle): We must ignore this flag to prevent the flag package from
```
jalextowle, line 11: @albrow I ended up needing to do this so that I wouldn't get a "flag provided but not defined" error. This init needs to be in every package that is tested in the browser, but it can only be in a single file in that package since you can't define flags twice.

Can you think of a better way to do this?

jalextowle: I could create a `testutils` package and have every test file inherit from that package. I'll hold off on this until I hear your thoughts.

PR #805, opened 2020-05-07
blockwatch: Debug Parity Block Cannot Be Found Issue

fabioberger:
We've noticed a Mesh node on the Kovan network getting stuck at a specific block height and repeatedly logging `One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found` before finally exiting with `too many blocks (128) behind the latest block msg:block watcher exited with error`.

After digging into the Mesh and Parity codebases, the only place this error is emitted is in response to a `eth_getLogs` request. In examining that part of the code, there was no immediately obvious reason why it would get stuck in a loop due to this error.

This PR therefore does two things:
1. Adds the offending blockHash for which the logs were unable to be fetched to the error message, so that it is logged.
2. Emits logs for this error message as warnings, as opposed to errors, since it _should_ be a warning equivalent to when Geth returns `not found`.

PR #813, opened 2020-05-27
Release 9.4.0

jalextowle:
This release includes:
- A performance enhancement for the `orderfilter` package in the browser (#809).
- A debugging feature (#805).


docs/browser-bindings/browser-lite/reference.md
```diff
@@ -429,55 +429,135 @@ ___
...
 
 • **Warn**: = 3
 
-*Defined in [types.ts:209](https://github.com/0xProject/0x-mesh/blob/8df8910/packages/browser-lite/src/types.ts#L209)*
+*Defined in [types.ts:209](https://github.com/0xProject/0x-mesh/blob/a929a43e/packages/browser-lite/src/types.ts#L209)*
+
+
+<hr />
+
+# Interface: SchemaValidationResult
```
jalextowle, line 485: I'll need to `@ignore` these, but I don't think it should block the release.

packages/browser/conversion-tests/conversion_test.ts
```diff
@@ -404,7 +404,7 @@ function testGetOrdersResponse(getOrdersResponse: WrapperGetOrdersResponse[]): v
     printer(
         'orderInfo.signedOrder.makerAssetData',
         getOrdersResponse[1].ordersInfos[0].signedOrder.makerAssetData ===
-        '0xf47261b0000000000000000000000000871dd7c2b4b25e1aa18728e9d5f2af4c4e431f5c',
+            '0xf47261b0000000000000000000000000871dd7c2b4b25e1aa18728e9d5f2af4c4e431f5c',
```
jalextowle, line 407: This was courtesy of my prettier tool. We should probably add a CI step that verifies that the Typescript has been prettified (I'll handle this after the release).

packages/browser-lite/src/schema_validator.ts
```diff
@@ -0,0 +1,79 @@
...
+    AJV.addSchema({
+        ...customOrderSchema,
+        $id: '/customOrder',
+    });
+    for (const schema of rootSchemas) {
+        AJV.addSchema(JSON.parse(schema));
+    }
+
+    const orderValidate = AJV.getSchema('/rootOrder');
+    // tslint:disable-next-line:no-non-null-assertion
```
jalextowle, line 47: This is unnecessary :(. I'll fix it in my post-release PR

packages/browser-lite/package.json
```diff
@@ -14,14 +15,17 @@
         "docsPath": "../../docs/browser-bindings/browser-lite"
     },
     "dependencies": {
+        "@0x/contract-addresses": "^4.9.0",
```
albrow, line 18: I wonder where this dependency came from? Mesh manages its own contract addresses.

jalextowle: Ah, I added this in my PR because I was using contract addresses before the last review. I need to remove this

PR #814, opened 2020-05-30
Adds cleanup hooks to several webpack configs

jalextowle:


albrow:
@jalextowle can you provide some more context about what these changes actually do?

Issue #726, opened 2020-02-26
Upgrade to Go 1.14


albrow:
[Go 1.14 has just been released](https://golang.org/doc/go1.14). We should upgrade Mesh as soon as possible. One of the most relevant changes for us is that "JavaScript values referenced from Go via js.Value objects can now be garbage collected." This could greatly reduce memory usage in the browser.

PR #815, opened 2020-06-02
Upgrade Mesh to Go 1.14

jalextowle:
This PR upgrades Mesh to Go version 1.14. All of the Go forks were updated, and their branches have been documented within the `go.mod` file.

Fixes https://github.com/0xProject/0x-mesh/issues/726.


tsconfig.json
```diff
@@ -14,7 +14,8 @@
         "composite": true,
         "declaration": true,
         "declarationMap": true,
-        "sourceMap": true
+        "sourceMap": true,
+        "allowJs": true
```
jalextowle, line 18: @albrow This setting allows us to include Javascript files along with our Typescript. `tsc` generates the appropriate declaration and map files.

This allowed me to avoid converting `wasm_exec.js` to a Typescript file. Any reason why you think this is a bad idea?

zeroex/asset_data_decoder.go
```diff
@@ -198,7 +198,7 @@ func (a *AssetDataDecoder) Decode(assetData []byte, decodedAssetData interface{}
 	}
 
 	// This is necessary to prevent a nil pointer exception for ABIs with no inputs
-	if info.abi.Methods[info.name].Inputs.LengthNonIndexed() == 0 {
+	if len(info.abi.Methods[info.name].Inputs.NonIndexed()) == 0 {
```
jalextowle, line 201: Geth removed the `LengthNonIndexed` function entirely.

zeroex/order.go
```diff
@@ -513,10 +513,10 @@ func SignTestOrder(order *Order) (*SignedOrder, error) {
 	return signedOrder, nil
 }
 
-// Trim converts the order to a TrimmedOrder, which is the format expected by
+// Trim converts the order to a LibOrderOrder, which is the format expected by
 // our smart contracts. It removes the ChainID and ExchangeAddress fields.
-func (s *SignedOrder) Trim() wrappers.TrimmedOrder {
-	return wrappers.TrimmedOrder{
+func (s *SignedOrder) Trim() wrappers.LibOrderOrder {
```
jalextowle, line 518: `abigen` now generates good names for all of the generated types, so it wasn't necessary to alias these names to `TrimmedOrder` anymore.

.circleci/config.yml
```diff
@@ -28,13 +28,15 @@ jobs:
       - run:
           name: Install dependencies
           command: make deps-no-lockfile
-      - run:
-          name: Vendor Go dependencies as a workaround for https://github.com/golangci/golangci-lint/issues/865
-          command: go mod vendor
       - run:
           name: Install Go linter
-          command: curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.22.2
+          command: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
```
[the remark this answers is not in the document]
albrow: The `golangci` team probably just changed their recommendation. This looks fine to me.

tsconfig.json
jalextowle: It has been moved to `@0x/mesh-browser-lite`'s `tsconfig.json` now, but this comment is still relevant.

db/open_js.go
```diff
@@ -26,7 +27,7 @@ func Open(path string) (*DB, error) {
 	// The global willLoadBrowserFS variable indicates whether browserFS will be
 	// loaded. browserFS has to be explicitly loaded in by JavaScript (and
 	// typically Webpack) and can't be loaded here.
-	if willLoadBrowserFS := js.Global().Get("willLoadBrowserFS"); willLoadBrowserFS != js.Undefined() && willLoadBrowserFS.Bool() == true {
+	if willLoadBrowserFS := js.Global().Get("willLoadBrowserFS"); !jsutil.IsNullOrUndefined(willLoadBrowserFS) && willLoadBrowserFS.Bool() == true {
```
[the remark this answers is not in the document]
albrow: We never set it to `null` at any point. But it doesn't really hurt to check anyways so I'm okay with this change.

zeroex/orderwatch/decoder/event_decoder.go
```diff
@@ -874,7 +874,7 @@ func (d *Decoder) decodeERC721(log types.Log, decodedLog interface{}) error {
 
 	erc721Err := unpackLog(decodedLog, eventName, log, d.erc721ABI)
 	if erc721Err != nil {
-		if _, ok := erc721Err.(UnsupportedEventError); ok {
+		if strings.Contains(erc721Err.Error(), "Unpack(no-values unmarshalled") {
```
albrow, line 877: What's the reason for this change? It functionally changes how we handle `UnsupportedEventError` and looks like it introduces a bug.

albrow:
@jalextowle Before we merge I still want to double check the changes that you made to `go-ethereum`, `go-ws-transport`, and `goleveldb`. We also have an open discussion on Slack about using `--composite` together with `--allowJs`.

zeroex/orderwatch/decoder/event_decoder.go
jalextowle: This change was made due to a change in the way that Go's `abi` package unmarshals events. There is no longer a way to distinguish invalid events by looking at the difference in length between `indexed` and `len(topics) - 1` from what I could tell by trying to fix the bug that was introduced by upgrading. 

The error that is now emitted when an invalid ERC721 asset data is passed to `unpackLog` is something that begins with the error seen above. 

FWIW, this passes all of the tests with the same behavior as the old `if` statement, but I'll grant you that the tests passing is not the same as a bug not existing.

albrow: Can you move the `strings.Contains(erc721Err.Error(), "Unpack(no-values unmarshalled")` to `unpackLog` instead of doing it here?

Basically change this:

```
if len(indexed) != len(log.Topics[1:]) {
	return UnsupportedEventError{Topics: log.Topics, ContractAddress: log.Address}
}
```

To:

```
if err := _abi.Unpack(decodedEvent, event, log.Data); err != nil {
	if strings.Contains(err.Error(), "Unpack(no-values unmarshalled") {
		return UnsupportedEventError{Topics: log.Topics, ContractAddress: log.Address}
	}
	return err
}
```

We should avoid error string comparisons altogether if possible, however, we can't avoid it here because of [how the error is created in `go-ethereum`](https://github.com/ethereum/go-ethereum/blob/f3f1e59eea922e5fb0d78d55aba471c35c20b022/accounts/abi/argument.go#L95). In this case, we should try and keep the string comparison as close as possible to the original call site (i.e. in this case where we call `ABI.Unpack`).

This will also make the code more consistent with the previous behavior. As it is currently written there are two differences that could cause problems:

1. There are situations where we used to return `UnsupportedEventError` from `unpackLog` but we currently do not. Namely, if `ABI.Unpack` fails.
2. There are situations where we used to return `UnsupportedEventError` from `decodeERC721` but  we currently do not. Namely, if the fallback encoding fails we return `erc721Err`.

zeroex/orderwatch/decoder/event_decoder.go
```diff
@@ -916,6 +914,9 @@ func (d *Decoder) decodeExchange(log types.Log, decodedLog interface{}) error {
 func unpackLog(decodedEvent interface{}, event string, log types.Log, _abi abi.ABI) error {
 	if len(log.Data) > 0 {
 		if err := _abi.Unpack(decodedEvent, event, log.Data); err != nil {
+			if strings.Contains(err.Error(), "Unpack(no-values unmarshalled") {
```
jalextowle, line 917: @albrow I needed to place this in a slightly different place than what you suggested. This is actually where the event seems to be emitted. 

What's weird is that it seems like the event only has two topics, which suggests that there could be a bug in the way that Geth handles non-indexed topics.

PR #817, opened 2020-06-04
Post v9.4.0 Cleanup

jalextowle:
## Changes
- Hides the `schema_validator` module so that our documentation does not include these un-exported (from `index.ts`) types and functions.
- Removes the `@0x/contract-addresses` dependency from `@0x/mesh-browser-lite`.

### DIFF
--- a/packages/browser-lite/package.json
+++ b/packages/browser-lite/package.json
@@ -15,7 +15,6 @@
         "docsPath": "../../docs/browser-bindings/browser-lite"
     },
     "dependencies": {
-        "@0x/contract-addresses": "^4.9.0",
         "@0x/order-utils": "^10.2.0",
         "@0x/utils": "^5.4.0",
         "ajv": "^6.12.2",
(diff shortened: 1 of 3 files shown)


jalextowle:
@albrow I wanted to speak with you about spending time implementing this, but I think that it would make sense to create a `prettier` command and add a `prettier --check` step in CI.

albrow:
@jalextowle I have already mentioned before that I am in favor of using stricter Prettier and TSLint rules, but we need to do it right. Let's agree on a __.prettierrc__ and __tslint.json__, include them in the repository, and add instructions to __CONTRIBUTING.md__ for automatically configuring your editor to use them. Then we can add a check in CI. I don't want to take half measures or skip any of those steps since it will potentially make development more difficult.

jalextowle:
@albrow Cool, this is exactly what I was looking for regarding what you are looking for from a prettier setup.

I would prefer to include this in the DevEx PR. I mostly dropped this in here because my remaining comment on the v9.4.0 release PR regarded prettier.

PR #818, opened 2020-06-05
Improves the DevEx of working on Typescript

jalextowle:
## Changes
- Fixes a problem that caused `yarn` to write local dependencies to a local (and un-symlinked) `node_modules/` folder. This made it necessary to use `yarn install --force` to propagate changes between referenced packages.
- Added `prettier` to Mesh.
- Fixed a problem with `yarn clean` that could cause `@0x/mesh-webpack-example-lite` to not be cleaned.


albrow:
@jalextowle I fixed it so that running `yarn build` inside a single package (e.g. __packages/test-wasm__) will also automatically rebuild that packages dependencies. Here's what I did:

1. Separate __tsconfig.json__ a the root of the repository into __tsconfig.json__ and __tsconfig-base.json__. The former contains all the references to __./packages__ and is what the TypeScript compiler uses when you run `tsc -b` at the root of the repository. The latter is what all the other tsconfig files inherit from.
2. Added an explicit `"references"` field to the __tsconfig.json__ file for each package in __packages/__. There is one reference for each dependency. So for example, __packages/test-wasm__ has `"references": [{"path": "../browser-lite" }]` because it depends on the `@0x/mesh-browser-lite` package.
3. Added `tsc -b` in __package.json__ for each package in __packages/__. Some packages which were built with Webpack didn't have this.

Importing from __lib/__ turns out not to be a problem, so we should keep doing that. It's actually what TypeScript does by default if you do `import "package-name"` for a referenced project.

packages/integration-tests/tsconfig.json
```diff
@@ -1,9 +1,14 @@
 {
-    "extends": "../../tsconfig",
-    "compilerOptions": { 
-        "outDir": "./dist", 
-        "rootDir": "./src" ,
+    "extends": "../../tsconfig-base",
+    "compilerOptions": {
+        "outDir": "./dist",
```
jalextowle, line 4: Should this also be changed to `./lib`?

tsconfig-base.json
```diff
@@ -0,0 +1,21 @@
...
+        "pretty": true,
+        "skipLibCheck": true,
+        "typeRoots": ["node_modules/@0x/typescript-typings/types", "node_modules/@types"],
+        "strict": true,
+        // These settings are required for TypeScript project references
+        "composite": true,
+        "declaration": true,
+        "declarationMap": true,
+        "sourceMap": true,
+        "allowJs": true
```
jalextowle, line 19: This isn't required for Typescript project references, so it should be in the top part of the file.

PR #820, opened 2020-06-08
Optimize InefficientlyConvertToJS

albrow:
This PR optimizes `jsutil.InefficientlyConvertToJS` by using `JSON.parse` in JavaScript land.

I also added some benchmarks, which converts slices of `testValue` with lengths between 1 and 1,000. The benchmarks show up to a 10x performance increase for large values.

For future optimizations, we should look at manually writing [`JSValue`](https://golang.org/pkg/syscall/js/#Wrapper) or [`MarshalJSON`](https://golang.org/pkg/encoding/json/#Marshaler) methods for types that we need to frequently convert in hot code paths.

Here are the results with the old version:

```
> WASM_INIT_FILE="$(pwd)/packages/test-wasm/dist/browser_shim.js" GOOS=js GOARCH=wasm go test -exec="wasmbrowsertest" ./packages/browser/go/jsutil -bench=. -run=NONE
goos: js
goarch: wasm
pkg: github.com/0xProject/0x-mesh/packages/browser/go/jsutil
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_1         	   10683	    100016 ns/op
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_10        	    1705	    737314 ns/op
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_100       	     172	   7050261 ns/op
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_1000      	      15	  89819682 ns/op
```

And here are the results with the new version:

```
> WASM_INIT_FILE="$(pwd)/packages/test-wasm/dist/browser_shim.js" GOOS=js GOARCH=wasm go test -exec="wasmbrowsertest" ./packages/browser/go/jsutil -bench=. -run=NONE
goos: js
goarch: wasm
pkg: github.com/0xProject/0x-mesh/packages/browser/go/jsutil
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_1         	   36669	     31170 ns/op
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_10        	    9052	    118822 ns/op
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_100       	    1170	    933889 ns/op
BenchmarkInefficientlyConvertToJS/BenchmarkInefficientlyConvertToJS_1000      	     130	   9021154 ns/op
```

|  Size  | Old         | New        | Change  |
|-------|-------------|------------|---------|
| 1     | 100,016 ns   | 31,170 ns   | -68.8% |
| 10    | 737,314 ns   | 118,822 ns  | -83.9% |
| 100   | 7,050,261 ns  | 933,889 ns  | -86.7% |
| 1,000 | 89,819,682 ns | 9,021,154 ns | -89.9% |


packages/browser/go/jsutil/jsutil.go
```diff
@@ -47,16 +46,22 @@ func WrapInPromise(f func() (interface{}, error)) js.Value {
...
 // encoding to JSON and then decoding it. This function is not very efficient
 // and its use should be phased out over time as much as possible.
-func InefficientlyConvertToJS(value interface{}) (js.Value, error) {
-	var jsValue interface{}
-	buf := bytes.Buffer{}
-	if err := json.NewEncoder(&buf).Encode(value); err != nil {
-		return js.Undefined(), err
-	}
-	if err := json.NewDecoder(&buf).Decode(&jsValue); err != nil {
+func InefficientlyConvertToJS(value interface{}) (result js.Value, err error) {
```
jalextowle, line 49: Do you have a sense of how inefficient this is compared to other `JSValue` functions? How deserving is this function of its name?

packages/browser/go/jsutil/jsutil_bench_test.go
```diff
@@ -0,0 +1,108 @@
...
+	Float32 float32   `json:"float32"`
+	Float64 float64   `json:"float64"`
+	Byte    byte      `json:"byte"`
+	Rune    rune      `json:"rune"`
+	String  string    `json:"string"`
+	Bool    bool      `json:"bool"`
+	Time    time.Time `json:"time"`
+}
+
+func newTestValue() *testValue {
```
jalextowle, line 35: Hypothetically, these test values could skew the benchmark if the savings are considerably different for small number types compared to things like large strings. I think that it would be good to benchmark against arrays of asset data, big ints, or a mix of types that better describe our use case (like `zeroex.SignedOrder`). 

This is not to say that this benchmark is not useful. I just think that it would be good to have one more that is more tailored to our use-case.

packages/browser/go/jsutil/jsutil.go
albrow: No, as I mentioned in the PR description, we should explore that separately.

packages/browser/go/jsutil/jsutil_bench_test.go
albrow: There's no single type that we could use here that would cover all the places we use `InefficientlyConvertToJS` throughout the codebase. The idea here is just to give us a decent idea of how `InefficientlyConvertToJS` performs generally. Based on what I have seen so far, I think it is extremely unlikely that this change will make `InefficientlyConvertToJS` _slower_, no mater what type we are converting.

 If we have a need to optimize the encoding/conversion of a specific type (e.g. in the `database` package), we should just use a separate benchmark that looks at that type specifically.

jalextowle: I also think that’s unlikely. I think the real point is that we should have some benchmarks on more realistic benchmarks to see how much of a performance boost we get for Mesh specific use cases.

pgebheim:
Lovely example of micro optimizations when they're needed.

Thanks guys

PR #832, opened 2020-06-11
Re-implement dynamic max expiration time

albrow:
This PR re-implements dynamic max expiration time on top of the new database. This was one of the major remaining tasks from https://github.com/0xProject/0x-mesh/pull/793.

It is also simpler than the old implementation due to the removal of the `slowcounter` package and the `MaxExpirationTime` column in the `metadata` table. We now just look at the `orders` table in the database to determine the effective max expiration time for each call to `Watcher.meshSpecificOrderValidation`.

Note that if you are testing this PR yourself, you will need to wipe your database because this PR changes the schema for the `metadata` table.


albrow:
@jalextowle still WIP. As I mentioned in Slack I want to add a test for _increasing_ max expiration time.

db/sql_implementation.go
```diff
@@ -452,40 +484,35 @@ func (db *DB) UpdateOrder(hash common.Hash, updateFunc func(existingOrder *types
...
+	})
 }
 
 func (db *DB) AddMiniHeaders(miniHeaders []*types.MiniHeader) (added []*types.MiniHeader, removed []*types.MiniHeader, err error) {
 	defer func() {
 		err = convertErr(err)
 	}()
-	var miniHeadersToRemove []*sqltypes.MiniHeader
+
+	addedMap := map[common.Hash]*types.MiniHeader{}
```
albrow, line 515: I found a more efficient way to remove duplicates from the `added` and `removed` sets. Applied the same approach to both `AddMiniHeaders` and `AddOrders` for both the SQL and Dexie implementations.

zeroex/orderwatch/order_watcher.go
```diff
@@ -1766,33 +1818,54 @@ func (w *Watcher) removeAssetDataAddressFromEventDecoder(assetData []byte) error
...
 				"oldMaxExpirationTime": w.maxExpirationTime.String(),
 				"newMaxExpirationTime": fmt.Sprint(newMaxExpiration),
 			}).Debug("increasing max expiration time")
-			w.maxExpirationTime.Set(newMaxExpiration)
-			w.saveMaxExpirationTime(newMaxExpiration)
+			w.setMaxExpirationTime(newMaxExpiration)
+		}
+	}
+
+	// The minimum max expiration time should be the latest block timestamp + 1.
```
albrow, line 1838: This change is a departure from the previous implementation of dynamic max expiration time. I don't think it ever makes sense to have the max expiration time be lower than the current block timestamp. There are some cases where this can happen while we are waiting for the slow counter to catch up.

zeroex/orderwatch/order_watcher.go
```diff
@@ -1429,7 +1420,7 @@ func (w *Watcher) meshSpecificOrderValidation(orders []*zeroex.SignedOrder, chai
 			})
 			continue
 		}
-		if order.ExpirationTimeSeconds.Cmp(w.MaxExpirationTime()) == 1 {
+		if order.ExpirationTimeSeconds.Cmp(maxExpirationTime) != -1 {
```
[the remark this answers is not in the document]
jalextowle: To elaborate, in `ValidateAndStoreValidOrders`, we first call `meshSpecificOrderValidation`, which will mark orders with a lower `expirationTimeSeconds` than `maxExpirationTime` as `Rejected`.

If we suppose that some of the orders being "rejected" are `pinned`, then these orders will never even be considered by the `db.Add` function call.

zeroex/orderwatch/order_watcher.go
```diff
@@ -940,8 +865,55 @@ func (w *Watcher) add(orderInfos []*ordervalidator.AcceptedOrderInfo, validation
...
+	// TODO(albrow): In the future, we should return an error and then react to
+	// that error differently depending on whether the order was received via
+	// RPC or from a peer. In the former case, we should return an RPC error
+	// response indicating that the order was not in fact added. In the latter
+	// case, we should effectively no-op, neither penalizing the peer or
+	// emitting any order events.
+	for _, orderToAdd := range orderInfos {
+		_, wasAdded := addedMap[orderToAdd.OrderHash]
+		if !wasAdded {
+			addedEvent := &zeroex.OrderEvent{
```
z2trillion, line 889: In case this is still relevant with the planned changes:
We add an `addedEvent` for each order in `orderInfos`, so it'd be more clear if this loop just added the `stoppedWatchingEvent`'s.

We might not need to handle this edge case at all if we include the newly expired orders in both `removedOrders` and `addedOrders`?

zeroex/orderwatch/order_watcher.go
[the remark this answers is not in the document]
z2trillion: 👍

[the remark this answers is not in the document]
albrow: I fixed this by passing in `pinned` to `meshSpecificOrderValidation`.

zeroex/orderwatch/order_watcher.go
albrow: I started working on refactoring `orderwatch.Watcher.add` as we discussed in our call. But it turned out to be a huge rabbit hole and felt out of scope for this PR. Instead, I opted to keep the hack where we emit an `ADDED` event immediately followed by a `STOPPED_WATCHING` event. And I applied your first suggestion here.

I created https://github.com/0xProject/0x-mesh/issues/837 to track the tech debt associated with our order validation pipeline. It is definitely something that deserves another look.

db/db_test.go
```diff
@@ -50,6 +50,102 @@ func TestAddOrders(t *testing.T) {
 	}
 }
 
+func TestAddOrdersMaxExpirationTime(t *testing.T) {
```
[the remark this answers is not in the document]
albrow: I added an additional case on the end of `TestOrderWatcherDecreaseExpirationTime`. Also confirmed that the test failed prior to your suggested change here: https://github.com/0xProject/0x-mesh/pull/832#discussion_r439537661 and passed after implementing that change.

zeroex/orderwatch/order_watcher.go
```diff
@@ -940,8 +870,48 @@ func (w *Watcher) add(orderInfos []*ordervalidator.AcceptedOrderInfo, validation
...
+	// were never added due to the max expiration time effectively changing
+	// within the database transaction above. In other words, new orders that
+	// _were_ added can change the effective max expiration time, meaning some
+	// orders in orderInfos were actually not added. This should not happen
+	// often. For now, we respond by emitting an ADDED event (above) immediately
+	// followed by a STOPPED_WATCHING event. If this order was submitted via
+	// RPC, the RPC client will see a response that indicates the order was
+	// successfully added, and then it will look like we immediately stopped
+	// watching it. This is not too far off from what really happened but is
+	// slightly inefficient.
```
jalextowle, line 883: Should these orders just be considered "rejected"?

jalextowle: The case where a `pinned` order from RPC gets marked as `ADDED` and then `STOPPED_WATCHING` is what made me think of it, but I actually think it's the right approach for all orders.

albrow: > Should these orders just be considered "rejected"?

Ideally yes. This is something I talked to @z2trillion about. It turns out that making this change is more complicated than it seems. I'd rather address it later and I created https://github.com/0xProject/0x-mesh/issues/837 to track this piece of tech debt.

PR #839, opened 2020-06-16
Update removedChecker to use countInterval

z2trillion:
Before this, `permanentlyDeleteStaleRemovedOrders` runs once each `minRemovedCheckInterval`. With this change, we will check if the database utilization rates exceeds `databaseUtilizationThreshold` each `countInterval` and run `permanentlyDeleteStaleRemovedOrders` if so. Additionally, if `permanentlyDeleteStaleRemovedOrders` hasn't been run within the last `maxDeleteInterval`, we will run it.

### DIFF
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -21,6 +21,7 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/event"
 	logger "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
@@ -32,8 +33,8 @@ const (
 	// maxDeleteInterval specifies the maximum amount of time between calls to permanentlyDeleteStaleRemovedOrders
 	maxDeleteInterval = 5 * time.Minute
 
-	// countInterval specifies the amount of time between calls to db.CountOrders. If the count is higher than a threshold, then permanentlyDeleteStaleRemovedOrders will be called, so this is the minimum interval between calls to permanentlyDeleteStaleRemovedOrders.
-	countInterval = 5 * time.Second
+	// checkDatabaseUtilizationThresholdInterval specifies the amount of time between calls to db.CountOrders. If the count is higher than a threshold, then permanentlyDeleteStaleRemovedOrders will be called, so this is the minimum interval between calls to permanentlyDeleteStaleRemovedOrders.
+	checkDatabaseUtilizationThresholdInterval = 5 * time.Second
 
 	// If, after a call to db.CountOrders, the database utilization exceeds databaseUtilizationThreshold, then permanentlyDeleteStaleRemovedOrders will be called.
 	databaseUtilizationThreshold = 0.5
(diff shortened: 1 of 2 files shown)


zeroex/orderwatch/order_watcher.go
```diff
@@ -29,9 +29,14 @@ const (
...
 	minCleanupInterval = 1 * time.Hour
 
-	// minRemovedCheckInterval specifies the minimum amount of time between checks
-	// on whether to remove orders flaggged for removal from the DB
-	minRemovedCheckInterval = 5 * time.Minute
+	// maxDeleteInterval specifies the maximum amount of time between calls to permanentlyDeleteStaleRemovedOrders
+	maxDeleteInterval = 5 * time.Minute
+
+	// countInterval specifies the amount of time between calls to db.CountOrders. If the count is higher than a threshold, then permanentlyDeleteStaleRemovedOrders will be called, so this is the minimum interval between calls to permanentlyDeleteStaleRemovedOrders.
+	countInterval = 5 * time.Second
```
z2trillion, line 36: @albrow what do you think about this value and the one for `databaseUtilizationThreshold`?

zeroex/orderwatch/order_watcher.go
```diff
@@ -264,20 +267,31 @@ func (w *Watcher) cleanupLoop(ctx context.Context) error {
 }
 
 func (w *Watcher) removedCheckerLoop(ctx context.Context) error {
-	for {
-		start := time.Now()
-		if err := w.permanentlyDeleteStaleRemovedOrders(ctx); err != nil {
-			return err
-		}
+	lastDeleted := time.Now()
```
albrow, line 270: We still should call `permanentlyDeleteStaleRemovedOrders` at the start of `removedCheckerLoop`. It is likely that there are stale orders in the database when we are starting up, so we should remove them right away instead of waiting.

zeroex/orderwatch/order_watcher.go
```diff
@@ -264,20 +267,31 @@ func (w *Watcher) cleanupLoop(ctx context.Context) error {
...
+	for {
 		select {
 		case <-ctx.Done():
 			return nil
-		// Wait minRemovedCheckInterval before calling permanentlyDeleteStaleRemovedOrders again. Since
-		// we only start waiting _after_ permanentlyDeleteStaleRemovedOrders completes, we will never
-		// have multiple calls to permanentlyDeleteStaleRemovedOrders running in parallel
-		case <-time.After(minRemovedCheckInterval - time.Since(start)):
-			continue
+		case <-time.After(maxDeleteInterval - time.Since(lastDeleted)):
```
albrow, line 277: One effect of writing the `for` loop this way is that when _either_ `case` is triggered it causes us to break out of the `select` statement and continue to the next iteration of the `for` loop. This causes us to call `time.After` twice more, which would result in unnecessarily creating two new channels. Can we coalesce this into a single call to `time.After` instead? This would cut calls to `time.After` (and therefore the number of new channels created) in half.

For example, something like this:

```go
for {
	select {
	case <-ctx.Done():
		return nil
	case <-time.After(countInterval - time.Since(lastCounted)):
		if time.Since(lastDeleted) > maxDeleteInterval {
			// If it has been longer than maxDeleteInterval since the last time
			// we deleted stale orders, go ahead and delete them now.
			if err := w.permanentlyDeleteStaleRemovedOrders(ctx); err != nil {
				return err
			}
			lastDeleted = time.Now()
		} else {
			// Otherwise, check the current number of orders in the dataase to see
			// if we are above databaseUtilizationThreshold.
			count, err := w.db.CountOrders(nil)
			if err != nil {
				return err
			}
			lastCounted = time.Now()

			if float64(count)/float64(w.maxOrders) > databaseUtilizationThreshold {
				if err := w.permanentlyDeleteStaleRemovedOrders(ctx); err != nil {
					return err
				}
				lastDeleted = time.Now()
			}
		}
	}
}
```

Technically this means we could be waiting up to `countInterval` longer than `maxDeleteInterval` between calls to `permanentlyDeleteStaleRemovedOrders`, but I don't think that's really an issue.

zeroex/orderwatch/order_watcher.go
albrow, line 36: I think naming this `countInterval` is not specific enough. Keep in mind that this constant is in scope for the entire `orderwatch` package. How about `checkDatabaseUtilizationThresholdInterval`?

z2trillion:
I ran this on the browser node for `MaxOrdersInStorage` = 1000 and 1e4, and in both cases performance seemed to be ok. At `MaxOrdersInStorage` = 1000, the database utilization was 100%, meaning that `permanentlyDeleteStaleRemovedOrders` was being called every 5 seconds.

PR #840, opened 2020-06-16
Optimize FindMiniHeaders by using gob encoding for event logs

albrow:
This PR optimizes `FindMiniHeaders` in the browser by using gob encoding for event logs instead of JSON. We could see from the Chrome profiles that this method was one of the slowest ones after the database upgrade. The reason it was slow was due to the fact that we were encoding a large number of event logs to JSON (sometimes up to 80kB of data). I made several different attempts to optimize the encoding, including custom encodings like comma separated strings and exploiting the fact that some fields are fixed length. Gob encoding turned out to have the best performance.

As a consequence of this change, we no longer have the ability to query MiniHeaders by event logs. Gob encoding is completely opaque in JavaScript land and just looks like a jumbled up string. It is also not guaranteed to be in any particular order. However, we were only querying event logs in the db tests and we have not had a need to do this in Mesh itself. I think this trade-off is worth it.

I added a new benchmark in __db_bench_test.go__ which calls `FindMiniHeaders` with 20 MiniHeaders in the database. Each one uses a set of event logs that I grabbed from the real world. This represents a realistic worst case scenario for `FindMiniHeaders`.

The average time for this worst case scenario changed from ~115ms to ~27ms.

You can run the benchmarks with:

```
WASM_INIT_FILE="$(pwd)/packages/test-wasm/dist/browser_shim.js" GOOS=js GOARCH=wasm go test -exec="wasmbrowsertest" ./db -bench=. -run=NONE
```

Before this optimization:

```
BenchmarkFindMiniHeaders 	       9	 115438336 ns/op
```

After this optimization:

```
BenchmarkFindMiniHeaders 	      45	  27245221 ns/op
```

We can potentially add further improvements by implementing an in-memory write-through cache for `blockwatch.Stack`. We previously had something like this but it was removed in https://github.com/0xProject/0x-mesh/pull/793 due to the complexity of the implementation. At the time, we didn't realize the average size of event logs and the performance impact of always reading from the database, so it may be worth reconsidering that decision. Even with caching, this PR is still useful since we will sometimes need to read from the database as well (especially on startup).


albrow:
@jalextowle should we include these new functions in our conversion tests? If so, what is the best way to do that? This is the first time in the db package where we actually do some hand-written conversion to `JSValue` as opposed to just using `jsutil.InefficientlyConvertToJS`.

albrow:
@jalextowle added a new test baed on what we talked about. Can you take a look at https://github.com/0xProject/0x-mesh/pull/840/commits/e8afa0d4e2f9e7b460f8ea37aba61f50bacda3df and see if that is sufficient?

albrow:
@jalextowle

> Should this be “gobMiniHeader”, or is “got” the prefix that you intended?

It was intended to be `gotMiniHeader` but I renamed it to make it more clear.

PR #841, opened 2020-06-16
Break up `MatchOrder` and `ValidateOrder` to prevent UI stutter

jalextowle:
# Description

Despite becoming significantly faster after #809, calls to `MatchOrder` still clog up the event loop for up to 50 ms (with an average of around 20 ms). This PR makes validation asynchronous, which significantly improves this situation.

Another related issue is that the event loop is held up by the `resume` call that ultimately calls `HandleOrderSyncResponse`. This can take up to 1 second to be processed, which is an untenable amount of time. This problem is tracked in an issue on the Golang Github repository https://github.com/golang/go/issues/39620.

### DIFF
--- a/orderfilter/validate_js.go
+++ b/orderfilter/validate_js.go
@@ -4,6 +4,8 @@ package orderfilter
 
 import (
 	"errors"
+	"syscall/js"
+	"time"
 
 	"github.com/0xProject/0x-mesh/packages/browser/go/jsutil"
 	"github.com/0xProject/0x-mesh/zeroex"
@@ -33,12 +35,25 @@ func (s *SchemaValidationResult) Errors() []*SchemaValidationError {
 // ValidateOrderJSON Validates a JSON encoded signed order using the AJV javascript library.
 // This libarary is used to increase the performance of Mesh nodes that run in the browser.
 func (f *Filter) ValidateOrderJSON(orderJSON []byte) (*SchemaValidationResult, error) {
-	jsResult, err := jsutil.AwaitPromise(f.orderValidator.Invoke(string(orderJSON)))
-	if err != nil {
+	resultChan := make(chan js.Value, 1)
+	errChan := make(chan error, 1)
+	go func() {
+		// FIXME(jalextowle): Reduce this if possible
+		time.Sleep(time.Millisecond)
+		result, err := jsutil.AwaitPromise(f.orderValidator.Invoke(string(orderJSON)))
+		if err != nil {
+			errChan <- err
+		}
+		resultChan <- result
+	}()
+	var result js.Value
+	select {
+	case err := <-errChan:
 		return nil, err
+	case result = <-resultChan:
 	}
-	valid := jsResult.Get("success").Bool()
-	jsErrors := jsResult.Get("errors")
+	valid := result.Get("success").Bool()
+	jsErrors := result.Get("errors")
 	var convertedErrors []*SchemaValidationError
 	for i := 0; i < jsErrors.Length(); i++ {
 		convertedErrors = append(convertedErrors, &SchemaValidationError{errors.New(jsErrors.Index(i).String())})
@@ -47,11 +62,24 @@ func (f *Filter) ValidateOrderJSON(orderJSON []byte) (*SchemaValidationResult, e
 }
 
 func (f *Filter) MatchOrderMessageJSON(messageJSON []byte) (bool, error) {
-	jsResult, err := jsutil.AwaitPromise(f.messageValidator.Invoke(string(messageJSON)))
-	if err != nil {
+	resultChan := make(chan js.Value, 1)
+	errChan := make(chan error, 1)
+	go func() {
+		// FIXME(jalextowle): Reduce this if possible
+		time.Sleep(time.Millisecond)
+		result, err := jsutil.AwaitPromise(f.messageValidator.Invoke(string(messageJSON)))
+		if err != nil {
+			errChan <- err
+		}
+		resultChan <- result
+	}()
+	var result js.Value
+	select {
+	case err := <-errChan:
 		return false, err
+	case result = <-resultChan:
 	}
-	return jsResult.Get("success").Bool(), nil
+	return result.Get("success").Bool(), nil
 }
 
 func (f *Filter) ValidateOrder(order *zeroex.SignedOrder) (*SchemaValidationResult, error) {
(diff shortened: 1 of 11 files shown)


zeroex/ordervalidator/order_validator_js.go
```diff
@@ -43,3 +54,47 @@ func (s RejectedOrderStatus) JSValue() js.Value {
...
+					// This case should never be hit since we enforce that EthereumRPCMaxContentLength >= maxOrderSizeInBytes
+					log.WithField("signedOrder", signedOrder).Panic("EthereumRPCMaxContentLength is set so low, a single 0x order cannot fit beneath the payload limit")
+				}
+				chunkSizes = append(chunkSizes, nextChunkSize)
+				nextChunkSize = 1
+				payloadLength = jsonRPCPayloadByteLength + encodedSignedOrderByteLength
+			}
+			batchIdx = batchIdx + 1
+		}
+		wg.Done()
```
albrow, line 92: Can you move this `wg.Done()` call to the top of the goroutine and use `defer`? This guarantees we always call it and prevents deadlock.

orderfilter/validate_js.go
```diff
@@ -4,11 +4,17 @@ package orderfilter
...
 	"errors"
+	"syscall/js"
+	"time"
 
 	"github.com/0xProject/0x-mesh/packages/browser/go/jsutil"
 	"github.com/0xProject/0x-mesh/zeroex"
 )
 
+// sleepTime is used to force processes that block the event loop to give the
+// event loop time to continue. This should be as low as possible.
```
albrow, line 15: nit: This might be slightly more clear.

```suggestion
// JavaScript event loop time to continue. This should be as low as possible.
```

orderfilter/validate_js.go
```diff
@@ -33,13 +39,24 @@ func (s *SchemaValidationResult) Errors() []*SchemaValidationError {
...
 func (f *Filter) ValidateOrderJSON(orderJSON []byte) (*SchemaValidationResult, error) {
-	jsResult := f.orderValidator.Invoke(string(orderJSON))
-	fatal := jsResult.Get("fatal")
-	if !jsutil.IsNullOrUndefined(fatal) {
-		return nil, errors.New(fatal.String())
+	resultChan := make(chan js.Value, 1)
+	errChan := make(chan error, 1)
+	go func() {
+		time.Sleep(sleepTime)
+		result, err := jsutil.AwaitPromise(f.orderValidator.Invoke(string(orderJSON)))
```
albrow, line 46: Can we use `AwaitPromiseContext` here? We should really be using that everywhere going forward. `AwaitPromise` will never timeout and can potentially hang forever.

orderfilter/validate_js.go
```diff
@@ -48,12 +65,23 @@ func (f *Filter) ValidateOrderJSON(orderJSON []byte) (*SchemaValidationResult, e
...
 func (f *Filter) MatchOrderMessageJSON(messageJSON []byte) (bool, error) {
-	jsResult := f.messageValidator.Invoke(string(messageJSON))
-	fatal := jsResult.Get("fatal")
-	if !jsutil.IsNullOrUndefined(fatal) {
-		return false, errors.New(fatal.String())
+	resultChan := make(chan js.Value, 1)
+	errChan := make(chan error, 1)
+	go func() {
+		time.Sleep(sleepTime)
+		result, err := jsutil.AwaitPromise(f.messageValidator.Invoke(string(messageJSON)))
```
albrow, line 72: Same here regarding `AwaitPromise` vs. `AwaitPromiseContext`.

orderfilter/validate_js.go
```diff
@@ -33,13 +39,24 @@ func (s *SchemaValidationResult) Errors() []*SchemaValidationError {
...
 // This libarary is used to increase the performance of Mesh nodes that run in the browser.
 func (f *Filter) ValidateOrderJSON(orderJSON []byte) (*SchemaValidationResult, error) {
-	jsResult := f.orderValidator.Invoke(string(orderJSON))
-	fatal := jsResult.Get("fatal")
-	if !jsutil.IsNullOrUndefined(fatal) {
-		return nil, errors.New(fatal.String())
+	resultChan := make(chan js.Value, 1)
+	errChan := make(chan error, 1)
+	go func() {
+		time.Sleep(sleepTime)
```
albrow, line 45: `AwaitPromise` and `AwaitPromiseContext` already create a goroutine. Could we potentially add this `time.Sleep` statement there instead? The intention of these functions is to not block the main thread, so if `time.Sleep` is required in order to do that, it feels appropriate to always include it.

zeroex/ordervalidator/order_validator_js.go
```diff
@@ -43,3 +54,47 @@ func (s RejectedOrderStatus) JSValue() js.Value {
...
+	chunkSizes := []int{}
+
+	payloadLength := jsonRPCPayloadByteLength
+	nextChunkSize := 0
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		batchIdx := 0
+		for _, signedOrder := range signedOrders {
+			if batchIdx%computeOptimalChunkBatchSize == 2 {
```
albrow, line 74: This definitely deserves a comment. I'm assuming the intention is to only call `time.Sleep` some of the time instead of once for every signed order?

zeroex/ordervalidator/order_validator_js.go
```diff
@@ -43,3 +54,47 @@ func (s RejectedOrderStatus) JSValue() js.Value {
...
+// maxRequestContentLength as possible.
+func (o *OrderValidator) computeOptimalChunkSizes(signedOrders []*zeroex.SignedOrder) []int {
+	chunkSizes := []int{}
+
+	payloadLength := jsonRPCPayloadByteLength
+	nextChunkSize := 0
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		batchIdx := 0
```
albrow, line 72: What is "Idx" an abbreviation for?

zeroex/ordervalidator/order_validator_js.go
```diff
@@ -43,3 +54,47 @@ func (s RejectedOrderStatus) JSValue() js.Value {
 		"message": s.Message,
 	})
 }
+
+const computeOptimalChunkBatchSize = 3
```
albrow, line 58: I think this also deserves a comment.

zeroex/ordervalidator/order_validator_js.go
jalextowle: Index

orderfilter/validate_js.go
jalextowle: I don't think so. `AwaitPromise` only creates `go-routines` for `.then` and `.catch`. I'll try it out, but I'm pretty sure that we need to sleep before the promise begins to execute.

PR #842, opened 2020-06-19
Update mesh.ts

pgebheim:
Fix issue where initial query for orderInfos can come back empty.


jalextowle:
This is also a problem here: https://github.com/0xProject/0x-mesh/blob/development/packages/rpc-client/src/ws_client.ts#L317.

PR #846, opened 2020-06-19
Directly compute packed length of RPC call

z2trillion:
Avoid a call to `devUtilsABI.Pack` in `computeOptimalChunkSizes` by directly computing the packed length of the RPC call instead of packing it to determine its length.

Before this, running `WASM_INIT_FILE="$(pwd)/packages/test-wasm/dist/browser_shim.js" GOOS=js GOARCH=wasm go test -exec="wasmbrowsertest" ./zeroex/ordervalidator -bench=BenchmarkComputeOptimalChunkSizes -run=NONE` gave
```
BenchmarkComputeOptimalChunkSizes 	      16	  64794064 ns/op
```

After this:
```
BenchmarkComputeOptimalChunkSizes 	  532500	      1930 ns/op
```


zeroex/ordervalidator/order_validator_test.go
```diff
@@ -435,6 +439,42 @@ func TestComputeOptimalChunkSizesMultiAssetOrder(t *testing.T) {
...
+// get the number of bytes taken up by the SignedOrder alone.
+// i.e.: len(`"0x7f46448d0000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"`)
+const emptyGetOrderRelevantStatesCallDataByteLength = 268
+
+func TestComputeABIEncodedSignedOrderByteLength(t *testing.T) {
+	signedOrder := scenario.NewSignedTestOrder(t)
+	signedOrder.Order.MakerAssetData = common.Hex2Bytes("123412304102340120350120340123041023401234102341234234523452345234")
+	signedOrder.Order.MakerAssetData = common.Hex2Bytes("132519348523094582039457283452")
+	signedOrder.Order.MakerAssetData = common.Hex2Bytes("")
+	signedOrder.Order.MakerAssetData = common.Hex2Bytes("324857203942034562893723452345246529837")
```
jalextowle, line 454: Let's add more test cases for this function with longer asset data (more than 2 words). This might be a good place to whip out the `multiAssetData`, but I really just want to have more assurances that this function is right across a wider range of inputs.

zeroex/ordervalidator/order_validator.go
```diff
@@ -837,23 +821,16 @@ const emptyGetOrderRelevantStatesCallDataByteLength = 268
...
-		return 0, err
-	}
-	encodedSignedOrderByteLength := len(encodedData) - emptyGetOrderRelevantStatesCallDataByteLength
-	return encodedSignedOrderByteLength, nil
+func numWords(bytes []byte) int {
+	return (len(bytes) + 31) / 32
+}
+
+func computeABIEncodedSignedOrderByteLength(signedOrder *zeroex.SignedOrder) int {
+	// The fixed size fields in a SignedOrder take up 1536 bytes. The variable length fields take up 64 bytes per 256-bit word.
```
jalextowle, line 829: Where did you get that the variable length fields take up 64 bytes per 256 bit word? 256 bits is 32 bytes, and ABI encoding should only include an extra offset and length slot. I could be misunderstanding this, but this is fairly confusing to me without more context.

zeroex/ordervalidator/order_validator_test.go
jalextowle, line 454: Thanks for adding this test. This clarifies that the ABI encoder was not using a "packed" encoding.

zeroex/ordervalidator/order_validator_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+		TakerFeeAssetData:     multiAssetAssetData,
+		Salt:                  big.NewInt(int64(time.Now().Nanosecond())),
+		MakerFee:              big.NewInt(0),
+		TakerFee:              big.NewInt(0),
+		MakerAssetAmount:      big.NewInt(100),
+		TakerAssetAmount:      big.NewInt(42),
+		ExpirationTimeSeconds: big.NewInt(time.Now().Add(24 * time.Hour).Unix()),
+		ExchangeAddress:       ethereum.GanacheAddresses.Exchange,
+	}
+	signedOrder, err := zeroex.SignTestOrder(&order)
```
jalextowle, line 61: Optional Nit: You can probably use the `scenario` package to simplify this.

zeroex/ordervalidator/order_validator_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
...00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000204a7cb5fb70000000000000000000000001dc4c1cefef38a777b15aa20260a54e584b16c480000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001800000000000000000000000000000000000000000000000000000000000000003000000000000000000000000000000000000000000000000000000000000006400000000000000000000000000000000000000000000000000000000000003e90000000000000000000000000000000000000000000000000000000000002711000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000000000000000000000000000000000000c800000000000000000000000000000000000000000000000000000000000007d10000000000000000000000000000000000000000000000000000000000004e210000000000000000000000000000000000000000000000000000000000000044025717920000000000000000000000001dc4c1cefef38a777b15aa20260a54e584b16c4800000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000")
```
jalextowle, line 42: Nit: `multiAssetAssetData` => `multiAssetData` (at least, this is what I would normally call this. I can understand a difference of opinions on this).

jalextowle: I'd like to see some benchmarks on more normal asset data than this. This is a pretty pathological case that will be significantly harder on the ABI decoder than normal asset data. It would be nice to see what the time difference is on an order that only uses `erc20AssetData` for `makerAssetData` and `takerAssetData` and has empty fee asset data (this much closer to the average case for an order that is added to Mesh).

z2trillion: There is no ABI decoding occurring, only encoding. Only the length of the data should affect the encoding time, not the contents.

zeroex/ordervalidator/order_validator_bench_test.go
z2trillion: It doesn't work for `*testing.B`? Also, this is in the init, so it doesn't have access to `b` yet?

zeroex/ordervalidator/order_validator.go
z2trillion: it takes up 64 bytes because in JSON each byte requires two characters. I'll add that to the comment.

zeroex/ordervalidator/order_validator_bench_test.go
jalextowle: Good point. Forgot about that.

zeroex/ordervalidator/order_validator_bench_test.go
jalextowle: What I'm saying is that we should have more realistic benchmarks. These asset data lengths are not very realistic or the average case.

My mistake on the decoding vs encoding. I don't think that changes my suggestion though as this will be more taxing to encode as well.

z2trillion:
New before benchmarks:
```BenchmarkErc20ComputeOptimalChunkSizes      	      40	  26752755 ns/op
BenchmarkMultiAssetComputeOptimalChunkSizes 	      18	  62220288 ns/op```

New after benchmarks
```BenchmarkErc20ComputeOptimalChunkSizes      	 1409587	       791 ns/op
BenchmarkMultiAssetComputeOptimalChunkSizes 	  784584	      1402 ns/op```

File: zeroex/ordervalidator/order_validator_test.go [hunk 5]
```diff
@@ -435,6 +446,60 @@ func TestComputeOptimalChunkSizesMultiAssetOrder(t *testing.T) {
...
+
+	encoded, err := abiEncode(signedOrder)
+	require.NoError(t, err)
+	expectedLength := len(encoded) - emptyGetOrderRelevantStatesCallDataStringLength
+
+	length := computeABIEncodedSignedOrderStringLength(signedOrder)
+
+	assert.Equal(t, expectedLength, length)
+}
```
jalextowle, line 501: Can you combine these into a single test case using table driven testing? An example of this pattern can be found here: https://github.com/0xProject/0x-mesh/blob/master/orderfilter/filter_test.go#L27. 

Let's add one more case that makes use of the `multiAssetData` from the other file.

zeroex/ordervalidator/order_validator.go
```diff
@@ -837,23 +821,16 @@ const emptyGetOrderRelevantStatesCallDataByteLength = 268
...
-		return 0, err
-	}
-	encodedSignedOrderByteLength := len(encodedData) - emptyGetOrderRelevantStatesCallDataByteLength
-	return encodedSignedOrderByteLength, nil
+func numWords(bytes []byte) int {
+	return (len(bytes) + 31) / 32
+}
+
+func computeABIEncodedSignedOrderStringLength(signedOrder *zeroex.SignedOrder) int {
+	// The fixed size fields in a SignedOrder take up 1536 bytes. The variable length fields take up 64 characters per 256-bit word.
```
jalextowle, line 829: Can you add the comment about a byte being encoded into two bytes in JSON?

PR #848, opened 2020-06-20
Parallelize requesting from peers in `ordersync`

jalextowle:
# Description

This PR addresses one of the lowest hanging pieces of fruit in `ordersync`. Previously, the `GetOrders` function would try to get orders from peers one at a time. This was easy to understand, but was also suboptimal from a performance perspective. This PR makes changes that allows this function to request orders from multiple peers at once.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ lint: lint-go lint-ts lint-prettier
 
 .PHONY: lint-go
 lint-go:
-	golangci-lint run
+	golangci-lint run --timeout 2m 
 
 
 .PHONY: lint-ts
--- a/core/ordersync/ordersync.go
+++ b/core/ordersync/ordersync.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/0xProject/0x-mesh/p2p"
@@ -44,6 +45,9 @@ const (
 	//    approxDelay * (1 - jitter) <= actualDelay < approxDelay * (1 + jitter)
 	//
 	ordersyncJitterAmount = 0.1
+	// maxRequestPeersInParallel is the largest number of peers that `GetOrders`
+	// will try to pull orders from at once.
+	maxRequestPeersInParallel = 10
 )
 
 var (
@@ -285,10 +289,11 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
 		default:
 		}
 
-		// TODO(albrow): As a performance optimization, do this for loop
-		// partly in parallel.
 		currentNeighbors := s.node.Neighbors()
 		shufflePeers(currentNeighbors)
+		i := 0
+		wg := &sync.WaitGroup{}
+		waitChan := make(chan struct{}, 1)
 		for _, peerID := range currentNeighbors {
 			if len(successfullySyncedPeers) >= minPeers {
 				return nil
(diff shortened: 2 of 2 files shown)


core/ordersync/ordersync.go
```diff
@@ -300,30 +306,47 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
+					}).Warn("could not get orders from peer via ordersync")
+				} else {
+					// TODO(albrow): Handle case where no orders were returned from this
+					// peer. This could be considered a valid response, depending on the implementation
+					// details of the subprotocol. We need to not try them again, but also not count
+					// them toward the number of peers we have successfully synced with.
+					log.WithFields(log.Fields{
+						"provider": id.Pretty(),
+					}).Trace("succesfully got orders from peer via ordersync")
+					successfullySyncedPeers.Add(id.Pretty())
```
albrow, line 337: `successfullySyncedPeers` is based on a `map` type under the hood and isn't threadsafe. We need to protect any reads or writes with a mutex.

core/ordersync/ordersync.go
```diff
@@ -300,30 +306,47 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
-				// details of the subprotocol. We need to not try them again, but also not count
-				// them toward the number of peers we have successfully synced with.
-				log.WithFields(log.Fields{
-					"provider": peerID.Pretty(),
-				}).Trace("succesfully got orders from peer via ordersync")
-				successfullySyncedPeers.Add(peerID.Pretty())
-			}
+			wg.Add(1)
+			go func(id peer.ID) {
+				select {
```
albrow, line 312: Usually I like to include `defer wg.Done()` as the very first line of the goroutine to make sure we don't forget it and it always runs regardless of code path. Is there a reason to not do that here? Failing to call `wg.Done` even once will cause deadlock.

Makefile
```diff
@@ -73,7 +73,7 @@ lint: lint-go lint-ts lint-prettier
 
 .PHONY: lint-go
 lint-go:
-	golangci-lint run
+	golangci-lint run --timeout 2m
```
jalextowle, line 76: This has been a problem on CI recently for some reason. I'm not 100% sure why (it started to occur after we switched to `go 1.14`). Let me know if you disagree with this change @albrow.

core/ordersync/ordersync.go
```diff
@@ -285,50 +289,79 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
+		semaphore := make(chan struct{}, maxRequestPeersInParallel)
 		currentNeighbors := s.node.Neighbors()
 		shufflePeers(currentNeighbors)
+		innerCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
 		for _, peerID := range currentNeighbors {
-			if len(successfullySyncedPeers) >= minPeers {
+			m.Lock()
+			successfullySyncedPeerLength := len(successfullySyncedPeers)
+			successfullySynced := successfullySyncedPeers.Contains(peerID.Pretty())
```
jalextowle, line 302: @albrow This shouldn't need to be in the mutex unless we have duplicate peer IDs in the `currentNeighbors` array. My guess is that we don't, but I also wanted to protect us from bugs in libp2p that could cause this assumption to not be true. Let me know if you think it's okay to access this outside of a lock.

Makefile
albrow: 👍 this is okay with me.

core/ordersync/ordersync.go
albrow: It doesn't matter whether there are any duplicate peer IDs. [Maps in Go are not goroutine safe](https://golang.org/doc/faq#atomic_maps). We need to protect reads and writes to the map with a `sync.RWMutex`.

jalextowle: Oh I see. I didn't realize that it could crash the program -- I assumed that it was merely a synchronization issue. I'll leave it in the mutex and change that to a `RWMutex`.

core/ordersync/ordersync.go
```diff
@@ -285,50 +289,79 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
+				} else {
+					// TODO(albrow): Handle case where no orders were returned from this
+					// peer. This could be considered a valid response, depending on the implementation
+					// details of the subprotocol. We need to not try them again, but also not count
+					// them toward the number of peers we have successfully synced with.
+					log.WithFields(log.Fields{
+						"provider": id.Pretty(),
+					}).Trace("succesfully got orders from peer via ordersync")
+					m.Lock()
+					successfullySyncedPeers.Add(id.Pretty())
```
albrow, line 340: Can we use `sync.RWMutex` here? Multiple map readers are okay as long as there are no writers, which makes this a good fit for `sync.RWMutex`. `successfullySyncedPeers.Add` is the only operation that should count as a write.

core/ordersync/ordersync.go
```diff
@@ -285,50 +289,81 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
+					m.Unlock()
+					m.RLock()
+					successfullySyncedPeerLength := len(successfullySyncedPeers)
```
jalextowle, line 343: @albrow I'm not sure that wrapping this in a `RLock` is worth the extra overhead, but it seems more technically correct. Do you have any insight into this?

albrow: My gut says just move `successfullySyncedPeerLength := len(successfullySyncedPeers)` before `m.Unlock`. Intuitively I would expect that check to be faster than acquiring a new mutex but I'm not 100% sure.

jalextowle: That's what I figured. I'll change it back.

PR #852, opened 2020-06-24
Use errgroup

z2trillion:

### DIFF
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -21,6 +21,7 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/event"
 	logger "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


zeroex/orderwatch/order_watcher.go
```diff
@@ -144,62 +145,21 @@ func (w *Watcher) Watch(ctx context.Context) error {
...
-		if err != nil {
-			logger.WithError(err).Error("error in orderwatcher cleanupLoop")
-			cancel()
-			return err
-		}
-	case err := <-removedCheckerLoopErrChan:
-		if err != nil {
-			logger.WithError(err).Error("error in orderwatcher removedCheckerLoop")
-			cancel()
+	loops := []func(context.Context) error{w.mainLoop, w.cleanupLoop, w.removedCheckerLoop}
```
jalextowle, line 150: Nit: It would be nice to have more descriptive error messages here. You could make this an array of structs that contain a loop function and function names, which would allow us to log a more descriptive error on line 156.

zeroex/orderwatch/order_watcher.go
```diff
@@ -144,62 +145,24 @@ func (w *Watcher) Watch(ctx context.Context) error {
...
-			return err
-		}
-	case err := <-removedCheckerLoopErrChan:
-		if err != nil {
-			logger.WithError(err).Error("error in orderwatcher removedCheckerLoop")
-			cancel()
+	namedLoops := []struct {
+		loop func(context.Context) error
+		name string
+	}{{w.mainLoop, "mainLoop"}, {w.cleanupLoop, "mainLoop"}, {w.removedCheckerLoop, "mainLoop"}}
```
jalextowle, line 153: Nit: These all have the same name right now

zeroex/orderwatch/order_watcher.go
```diff
@@ -144,62 +145,24 @@ func (w *Watcher) Watch(ctx context.Context) error {
...
-		if err != nil {
-			logger.WithError(err).Error("error in orderwatcher cleanupLoop")
-			cancel()
-			return err
-		}
-	case err := <-removedCheckerLoopErrChan:
-		if err != nil {
-			logger.WithError(err).Error("error in orderwatcher removedCheckerLoop")
-			cancel()
+	namedLoops := []struct {
```
albrow, line 150: I'm not sure if this process of setting up the `namedLoops` slice and iterating through it with a `for` loop is worth the complexity/overhead. Do you think it makes more sense to just call each function one at a time with `g.Go`?

zeroex/orderwatch/order_watcher.go
```diff
@@ -144,62 +145,24 @@ func (w *Watcher) Watch(ctx context.Context) error {
...
+		loop func(context.Context) error
+		name string
+	}{{w.mainLoop, "mainLoop"}, {w.cleanupLoop, "cleanupLoop"}, {w.removedCheckerLoop, "removedCheckerLoop"}}
+	for _, namedLoop := range namedLoops {
+		namedLoop := namedLoop
```
albrow, line 155: I realize this is here to avoid a common gotcha in Go, but it's not very obvious. Can you add a brief comment explaining why this is necessary?

zeroex/orderwatch/order_watcher.go
z2trillion: I like using the for loop because it makes it clear that we do exactly the same thing for all three loops.

zeroex/orderwatch/order_watcher.go
z2trillion: I added a link to the golang faq, which is what I have seen other go examples.

PR #853, opened 2020-06-29
Store order validation block

albrow:
This PR addresses one of the remaining TODOs from https://github.com/0xProject/0x-mesh/pull/793:

> Address problems that can arise from orders being validated at different block heights. Previously we used transactions in a few critical places to solve this problem, but the new database implementation cannot easily support transactions. We have also seen transactions that are open for too long contributing to poor CPU performance.

I've split this problem into two parts:

1. This PR adds two new fields to our database schema for orders. It also updates orderwatcher to set these values as needed.
2. A follow-up PR will use these new fields to implement stronger guarantees for order watching. Going forward, we will always be able to detect when an order was validated at a different block height and address the problem directly. There will no longer be the possibility of missing relevant block events.

This is a breaking change at the database layer and will require wiping your DB if you were previously running a build from the `development` branch.


zeroex/orderwatch/order_watcher.go
```diff
@@ -1157,42 +1159,46 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
...
+			// The order is expired if ExpirationTimeSeconds is equal to or less than the timestamp
+			// of the validation block.
+			validationBlockTimeStampSeconds := big.NewInt(validationBlock.Timestamp.Unix())
+			isOrderExpired := order.ExpirationTimeSeconds.Cmp(validationBlockTimeStampSeconds) != 1
 
 			if oldFillableAmount.Cmp(newFillableAmount) == 0 {
 				// If order was previously expired, check if it has become unexpired
-				if order.IsRemoved && oldFillableAmount.Cmp(big.NewInt(0)) != 0 && validationBlockTimestamp.Before(expiration) {
-					w.rewatchOrder(order, nil)
+				if order.IsRemoved && oldFillableAmount.Cmp(big.NewInt(0)) != 0 && !isOrderExpired {
```
z2trillion, line 1169: It would be nice to have `isOrderUnexpired := order.IsRemoved && oldFillableAmount.Cmp(big.NewInt(0)) != 0 && !isOrderExpired` and use it in the threelocations where we check.

zeroex/orderwatch/order_watcher_test.go
```diff
@@ -35,10 +35,16 @@ import (
 const (
 	ethereumRPCRequestTimeout   = 30 * time.Second
 	miniHeaderRetentionLimit    = 2
-	blockPollingInterval        = 1000 * time.Millisecond
+	blockPollingInterval        = 1 * time.Second
 	ethereumRPCMaxContentLength = 524288
 	maxEthRPCRequestsPer24HrUTC = 1000000
 	maxEthRPCRequestsPerSeconds = 1000.0
+
+	// processBlockSleepTime is the amount of time ot wait for order watcher to
```
z2trillion, line 43: ot  -> to

zeroex/orderwatch/order_watcher.go
```diff
@@ -1157,42 +1159,46 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
...
-				// No important state-change happened
+				// No important state-change happened. Still want to update lastValidatedBlock
+				w.updateOrderLastValidatedBlock(order, validationBlock)
 				continue
 			}
 			if oldFillableAmount.Cmp(big.NewInt(0)) == 1 && oldAmountIsMoreThenNewAmount {
 				// If order was previously expired, check if it has become unexpired
-				if order.IsRemoved && oldFillableAmount.Cmp(big.NewInt(0)) != 0 && validationBlockTimestamp.Before(expiration) {
-					w.rewatchOrder(order, newFillableAmount)
+				if order.IsRemoved && oldFillableAmount.Cmp(big.NewInt(0)) != 0 && !isOrderExpired {
```
z2trillion, line 1186: It's outside the scope of this PR, but we can simplify this code by checking for unexpiration, and then checking for amount changes.

common/types/types.go
```diff
@@ -132,6 +132,12 @@ type OrderWithMetadata struct {
 	ParsedMakerAssetData []*SingleAssetData `json:"parsedMakerAssetData"`
 	// Same as ParsedMakerAssetData but for MakerFeeAssetData instead of MakerAssetData.
 	ParsedMakerFeeAssetData []*SingleAssetData `json:"parsedMakerFeeAssetData"`
+	// LastValidatedBlockNumber is the block number at which the order was
+	// last validated.
+	LastValidatedBlockNumber *big.Int `json:"lastValidatedBlockNumber"`
+	// LastValidatedBlockHash is the hash of the block at which the order was
+	// last validated.
+	LastValidatedBlockHash common.Hash `json:"lastValidatedBlockHash"`
```
jalextowle, line 140: I feel like we don't really need these comments. That's just a personal preference though.

zeroex/orderwatch/order_watcher.go
albrow: Good suggestion. I know this logic is a bit of a mess right now. It's something we should look at again when we have time to address tech debt. This PR is just trying to do the bare minimum.

zeroex/orderwatch/order_watcher.go
albrow: 👍 I know this is pretty messy at the moment. Let's try and address it later when we have more time to spend on tech debt.

common/types/types.go
albrow: 🤷 I could go either way. I think since there already here and it's just one line we should leave them. Alright with you?

PR #856, opened 2020-07-08
Simplify unexpiration logic in convertValidationResultsIntoOrderEvents

z2trillion:

### DIFF
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -1178,7 +1178,7 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
 			expirationTimeIsValid := order.ExpirationTimeSeconds.Cmp(validationBlockTimestampSeconds) == 1
 			isOrderUnexpired := order.IsRemoved && expirationTimeIsValid
 
-			// If order was previously expired, check if it has become unexpired.
+			// We can tell that an order was previously expired if it was marked as removed with a non-zero fillable amount. There is no other explanation for this database state. The order is considered "unexpired" if it was previously expired but now has a valid expiration time based on the latest block timestamp.
 			if isOrderUnexpired {
 				w.rewatchOrder(order, newFillableAmount, validationBlock)
 				orderEvent := &zeroex.OrderEvent{
--- a/zeroex/orderwatch/order_watcher.go
+++ b/zeroex/orderwatch/order_watcher.go
@@ -1178,36 +1178,25 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
 			isOrderExpired := order.ExpirationTimeSeconds.Cmp(validationBlockTimeStampSeconds) != 1
 			isOrderUnexpired := order.IsRemoved && oldFillableAmount.Cmp(big.NewInt(0)) != 0 && !isOrderExpired
 
-			if oldFillableAmount.Cmp(newFillableAmount) == 0 {
-				// If order was previously expired, check if it has become unexpired
-				if isOrderUnexpired {
-					w.rewatchOrder(order, newFillableAmount, validationBlock)
-					orderEvent := &zeroex.OrderEvent{
-						Timestamp:                validationBlock.Timestamp,
-						OrderHash:                order.Hash,
-						SignedOrder:              order.SignedOrder(),
-						FillableTakerAssetAmount: order.FillableTakerAssetAmount,
-						EndState:                 zeroex.ESOrderUnexpired,
-					}
-					orderEvents = append(orderEvents, orderEvent)
+			if isOrderUnexpired {
+				w.rewatchOrder(order, newFillableAmount, validationBlock)
+				orderEvent := &zeroex.OrderEvent{
+					Timestamp:                validationBlock.Timestamp,
+					OrderHash:                order.Hash,
+					SignedOrder:              order.SignedOrder(),
+					FillableTakerAssetAmount: order.FillableTakerAssetAmount,
+					EndState:                 zeroex.ESOrderUnexpired,
 				}
+				orderEvents = append(orderEvents, orderEvent)
+			}
+
+			if oldFillableAmount.Cmp(newFillableAmount) == 0 {
 				// No important state-change happened. Still want to update lastValidatedBlock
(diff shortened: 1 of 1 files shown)


zeroex/orderwatch/order_watcher.go
```diff
@@ -1158,89 +1158,53 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
...
 				SignedOrder:              order.SignedOrder(),
-				FillableTakerAssetAmount: acceptedOrderInfo.FillableTakerAssetAmount,
+				FillableTakerAssetAmount: newFillableAmount,
 				EndState:                 zeroex.ESOrderAdded,
 				ContractEvents:           orderHashToEvents[order.Hash],
 			}
 			orderEvents = append(orderEvents, orderEvent)
 		} else {
-			// The order is expired if ExpirationTimeSeconds is equal to or less than the timestamp
+			// The time stamp is valid if ExpirationTimeSeconds is equal to or greater than the time stamp
```
jalextowle, line 1175: Nit: As is, this just be `greater than` rather than `equal to or greater than`. This will be correct when we make the change on the subsequent line.

Also, I normally see `time stamp` as one word (`timestamp`). I don't care which we use, but we should pick one and use it everywhere

zeroex/orderwatch/order_watcher.go
```diff
@@ -1158,89 +1158,53 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
...
 			// of the validation block.
 			validationBlockTimeStampSeconds := big.NewInt(validationBlock.Timestamp.Unix())
-			isOrderExpired := order.ExpirationTimeSeconds.Cmp(validationBlockTimeStampSeconds) != 1
+			timeStampIsValid := order.ExpirationTimeSeconds.Cmp(validationBlockTimeStampSeconds) == 1
```
jalextowle, line 1178: This is actually incorrect. The timestamp is only valid if it less than the expiration time: https://github.com/0xProject/0x-monorepo/blob/development/contracts/exchange/contracts/src/MixinExchangeCore.sol#L171. We've had to fix this elsewhere, so I guess this slipped through the cracks. I realize that this was in here before, but let's fix it in this PR since we're changing the line.

z2trillion: I think this is correct according to the code you linked? `timeStampIsValid` is true if `order.ExpirationTimeSeconds > validationBlockTimestampSeconds`?

jalextowle: Whoops, I looked at this the wrong way. You’re totally right

zeroex/orderwatch/order_watcher.go
```diff
@@ -1158,89 +1158,53 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
...
+				w.rewatchOrder(order, newFillableAmount, validationBlock)
 				orderEvent := &zeroex.OrderEvent{
 					Timestamp:                validationBlock.Timestamp,
-					OrderHash:                acceptedOrderInfo.OrderHash,
+					OrderHash:                order.Hash,
 					SignedOrder:              order.SignedOrder(),
-					EndState:                 zeroex.ESOrderFilled,
-					FillableTakerAssetAmount: acceptedOrderInfo.FillableTakerAssetAmount,
-					ContractEvents:           orderHashToEvents[order.Hash],
+					FillableTakerAssetAmount: order.FillableTakerAssetAmount,
```
albrow, line 1187: `order.FillableTakerAssetAmount` is the old fillable amount, which makes this a behavioral change. This change does have some advantages, namely that it makes reconstructing the order book state from order events a little bit more intuitive. But the downside is that it is slightly less efficient because MMs and other users would have to parse the next order event in order to see the latest fillable amount (even though  we already know what that is).

I think the two approaches are fairly equal in terms of pros/cons, but that means IMO we should keep the behavior the same instead of changing it. Was this just an oversight or was it an intentional change on your part? Curious to hear what you think?

```suggestion
					FillableTakerAssetAmount: newFillableAmount,
```

zeroex/orderwatch/order_watcher.go
```diff
@@ -1158,89 +1158,53 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
...
-						OrderHash:                order.Hash,
-						SignedOrder:              order.SignedOrder(),
-						FillableTakerAssetAmount: order.FillableTakerAssetAmount,
-						EndState:                 zeroex.ESOrderUnexpired,
-					}
-					orderEvents = append(orderEvents, orderEvent)
-				} else {
-					w.updateOrderFillableTakerAssetAmount(order, newFillableAmount, validationBlock)
+			} else {
+				w.updateOrderFillableTakerAssetAmount(order, newFillableAmount, validationBlock)
```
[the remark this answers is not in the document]
z2trillion: I renamed it to updateOrderFillableTakerAssetAmountAndBlockInfo. I'm going to make another PR that unifies `updateOrderFillableTakerAssetAmountAndBlockInfo`, `rewatchOrder`, and `unwatchOrder`, since they're all the same, except for how they handle `IsRemoved`.

zeroex/orderwatch/order_watcher.go
```diff
@@ -1178,6 +1178,7 @@ func (w *Watcher) convertValidationResultsIntoOrderEvents(
 			timestampIsValid := order.ExpirationTimeSeconds.Cmp(validationBlockTimestampSeconds) == 1
 			isOrderUnexpired := order.IsRemoved && timestampIsValid
 
+			// If order was previously expired, check if it has become unexpired.
```
albrow, line 1181: I think this comment belongs a line or two above. At this point we are not checking if the order has become unexpired, we _have_ already checked if the order has become unexpired (it has) and now we are emitting the appropriate events and updating the database based on that.

I also think we should add a little bit more context to the comment. In particular, it's important to note the relationship between fillable amount and whether an order was expired. The check for the fillable amount occurs above, so without a comment, the connection may not be clear. Something like:

```suggestion
			// We can tell that an order was previously expired if it was marked as removed with a non-zero fillable amount. There is no other explanation for this database state. The order is considered "unexpired" if it was previously expired but now has a valid expiration time based on the latest block timestamp.
```

PR #857, opened 2020-07-09
In-Memory Blockwatcher Syncing

jalextowle:
Prior to PR #793, the blockwatcher used an in-memory stack to build a new canonical chain prior to syncing the new chain with the database. This worsened the edge-case that occurs when an order is received by the orderwatcher while the blockwatcher is syncing and also had a performance impact. 

This PR brings back the old `simplestack` library and also makes a minor technical debt cleanup within the blockwatcher tests. The `simplestack` library can probably be optimized, but it is also battle-tested, so it made sense to bring it back and focus on higher priority issues prior to trying to optimize it.


ethereum/blockwatch/block_watcher.go
```diff
@@ -66,6 +67,7 @@ func (e TooMayBlocksBehindError) Error() string {
 
 // Config holds some configuration options for an instance of BlockWatcher.
 type Config struct {
+	MaxMiniHeaders  int
```
albrow, line 70: It looks like we never actually use this?

ethereum/blockwatch/block_watcher.go
```diff
@@ -302,14 +306,24 @@ func (w *Watcher) SyncToLatestBlock() error {
...
+		if err := w.stack.Reset(checkpoint); err != nil {
+			return err
+		}
+	} else {
+		_, err := w.stack.Checkpoint()
+		if err != nil {
 			return err
 		}
-		if _, _, err := w.stack.db.AddMiniHeaders(existingMiniHeaders); err != nil {
+		if _, err := w.db.DeleteMiniHeaders(nil); err != nil {
```
albrow, line 317: We need to do `DeleteMiniHeaders` and `AddMiniHeaders` in a transaction so that the database is always in a good state. Otherwise orderwatcher could read from the database in between when we call `DeleteMiniHeaders` and when we call `AddMiniHeaders`.

The current version of the `db` package does not expose a way to execute arbitrary transactions, partly for separation of concerns and partly due to the difficulty of designing a transaction interface that works with both SQL and Dexie.js. So I think our best option is to add a new method to the `db` package called something like `ResetMiniHeaders` which uses a transaction under the hood.

ethereum/blockwatch/block_watcher.go
```diff
@@ -302,14 +306,24 @@ func (w *Watcher) SyncToLatestBlock() error {
...
+			return err
+		}
+		newMiniheaders, err := w.stack.PeekAll()
+		if err != nil {
+			return err
+		}
+		if _, _, err := w.db.AddMiniHeaders(newMiniheaders); err != nil {
```
albrow, line 324: I believe we also need to update the database at the end of `FastSyncToLatestBlock` too. Maybe I'm just missing something but it looks like we update the in-memory stack without updating the database.

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,180 @@
+package simplestack
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/0xProject/0x-mesh/common/types"
+	// FIXME Remove this
```
albrow, line 8: Is this `FIXME` comment still relevant?

ethereum/blockwatch/block_watcher.go
jalextowle: Yikes, I probably needed to give this PR a once over before putting it up for review. I'll make sure I do that in the future

ethereum/blockwatch/block_watcher_test.go
```diff
@@ -347,7 +301,7 @@ func TestFilterLogsRecursively(t *testing.T) {
 		{
 			Label: "HAPPY_PATH",
 			rangeToFilterLogsResponse: map[string]filterLogsResponse{
-				"10-20": filterLogsResponse{
```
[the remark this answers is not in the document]
jalextowle: This is fixed in #861

ethereum/blockwatch/block_watcher.go
```diff
@@ -302,14 +316,18 @@ func (w *Watcher) SyncToLatestBlock() error {
 		return syncErr
 	}
 	if w.shouldRevertChanges(lastStoredHeader, allEvents) {
```
z2trillion, line 318: This function returns false if `len(events) == 0 || lastStoredHeader == nil`, but `len(events)` can't be since we would have returned in line 301 if so. I think that if `lastStoredHeader` is `nil` in line 242 (old) /  256 (new), then it will still be nil here, so maybe we should just do the revert there?

db/sql_implementation.go
```diff
@@ -567,6 +567,54 @@ func (db *DB) AddMiniHeaders(miniHeaders []*types.MiniHeader) (added []*types.Mi
 	return added, removed, nil
 }
 
+func (db *DB) ResetMiniHeaders(miniHeaders []*types.MiniHeader) (added []*types.MiniHeader, removed []*types.MiniHeader, err error) {
+	defer func() {
+		err = convertErr(err)
+	}()
+
+	removedMap := map[common.Hash]*types.MiniHeader{}
```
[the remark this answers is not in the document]
albrow: If you don't need to use `removed` or `added` for the purposes of this PR, then we don't need to return them. Let's just remove those return values and use the more performant option.

`sqlz` is a very thin abstraction layer and supports anything that SQL supports. In SQL the query to remove all rows in a table would be `DELETE FROM miniHeaders`. Or in `sqlz` syntax just `DeleteFrom("miniHeaders")`.

db/dexie_implementation.go
```diff
@@ -241,6 +241,22 @@ func (db *DB) AddMiniHeaders(miniHeaders []*types.MiniHeader) (added []*types.Mi
 	return dexietypes.MiniHeadersToCommonType(jsAdded), dexietypes.MiniHeadersToCommonType(jsRemoved), nil
 }
 
+func (db *DB) ResetMiniHeaders(miniHeaders []*types.MiniHeader) (added []*types.MiniHeader, removed []*types.MiniHeader, err error) {
```
albrow, line 244: I think we should add a comment explaining what `ResetMiniHeaders` does since it is less obvious than something like `DeleteOrder`. (Honestly, I should have done the same for other non-obvious methods like `UpdateOrder`).

This should also be added to the `Database` interface, which helps ensure that both the Dexie implementation and the SQL implementation have the exact same methods.

ethereum/blockwatch/block_watcher_test.go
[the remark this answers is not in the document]
albrow: I think this type of linter check should be considered extremely low priority. It doesn't functionally change anything and only offers a somewhat subjective legibility improvement.

ethereum/blockwatch/client.go
```diff
@@ -17,30 +16,34 @@ import (
...
+	bigIntParsingErrorString = "Failed to parse big.Int value from hex-encoded %s returned from %s"
 )
 
 // Client defines the methods needed to satisfy the client expected when
 // instantiating a Watcher instance.
 type Client interface {
-	HeaderByNumber(number *big.Int) (*types.MiniHeader, error)
-	HeaderByHash(hash common.Hash) (*types.MiniHeader, error)
-	FilterLogs(q ethereum.FilterQuery) ([]ethtypes.Log, error)
+	HeaderByNumber(ctx context.Context, number *big.Int) (*types.MiniHeader, error)
```
albrow, line 28: The only reason to pass in a separate context for each of these methods would be if we needed more fine-grained control over it. For example, if we wanted to cancel the context _independent_ of the main context that is created in the `core` package (or in __cmd/mesh__). I don't see any reason why we would want to do that.

Instead, can we add a `ctx` field to `RpcClient` and use that for all requests? That would significantly reduce the overhead and still behave the same way.

z2trillion: The documentation for context recommends not to do this: "Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx..." https://golang.org/pkg/context/

packages/mesh-browser-lite/src/database.ts
```diff
@@ -317,6 +322,45 @@ export class Database {
         };
     }
 
+    // ResetMiniHeaders(miniHeaders []*types.MiniHeader) (added []*types.MiniHeader, removed []*types.MiniHeader, err error)
```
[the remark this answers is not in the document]
jalextowle: We have this comment above every other database function. It's used to show what the corresponding function looks like in Go land.

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Return copy of miniHeaders array
+	m := make([]*types.MiniHeader, len(s.miniHeaders))
+	copy(m, s.miniHeaders)
+
+	return m, nil
+}
+
+// Clear removes all items from the stack and clears any set checkpoint
+func (s *SimpleStack) Clear() error {
```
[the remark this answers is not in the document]
jalextowle: I'm not sure honestly. I didn't actually write `simplestack`, I merely brought it back into the repo. I'm up to change small stuff like this though

ethereum/blockwatch/block_watcher.go
```diff
@@ -586,33 +604,28 @@ type blockRange struct {
 // blocks' logs twice.
 func (w *Watcher) getSubBlockRanges(from, to, rangeSize int) []*blockRange {
 	chunks := []*blockRange{}
-	numBlocksLeft := to - from
+	numBlocksLeft := to - from + 1
```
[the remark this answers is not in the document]
jalextowle: Good point

ethereum/blockwatch/block_watcher.go
```diff
@@ -139,10 +142,21 @@ func (w *Watcher) FastSyncToLatestBlock(ctx context.Context) (blocksElapsed int,
 			return blocksElapsed, err
 		}
 		if len(events) > 0 {
+			newMiniHeaders, err := w.stack.PeekAll()
+			if err != nil {
```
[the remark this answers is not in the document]
jalextowle: I think that we should always handle errors rather than leaving them be if they are always `nil` (this protects us if an error is returned in the future). In this case, `PeekAll` should just not return an error.

ethereum/simplestack/simple_stack.go
```diff
@@ -0,0 +1,177 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.updates = []*Update{}
+	s.latestCheckpointID++
+	return s.latestCheckpointID, nil
+}
+
+// Reset resets the in-memory stack with the contents from the latest checkpoint
+func (s *SimpleStack) Reset(checkpointID int) error {
```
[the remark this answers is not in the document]
jalextowle: Hmm, I'm not sure. This data structure can be used in concurrent data structures. Having a more permissive `Reset` would be dangerous when using this code in multiple different go-routines. I'm inclined to keep it the same in this case so that we're protected if someone ever tries to use `Reset` across go-routines. I'm interested to hear your thoughts on this though

ethereum/blockwatch/block_watcher.go
jalextowle: `lastStoredHeader` is updated on line 281 (of the version of the code that I'm looking at).

jalextowle: This logic doesn't make sense to me. If the function returns false, then we don't want to revert the changes. 

As a follow up, we don't know `nextLatestHeader` until after the for-loop (because `events` is added to by `buildCanonicalChain`). Having said this, I don't think that we can do the revert earlier unless me make a different refactor first.

ethereum/blockwatch/client.go
jalextowle: Adding context in a separate PR

albrow: @z2trillion this is one case where I do not agree with the documentation/recommendation.

db/dexie_implementation.go
```diff
@@ -241,6 +241,22 @@ func (db *DB) AddMiniHeaders(miniHeaders []*types.MiniHeader) (added []*types.Mi
 	return dexietypes.MiniHeadersToCommonType(jsAdded), dexietypes.MiniHeadersToCommonType(jsRemoved), nil
 }
 
+// ResetMiniHeaders deletes all of the existing miniheaders and then stores new
+// miniheaders in the database.
+func (db *DB) ResetMiniHeaders(miniHeaders []*types.MiniHeader) (err error) {
```
albrow, line 246: nit: I think we should either rename the parameter or change the comment to clarify that the "new miniheaders" means the given `miniHeaders` argument. I'm leaning toward the former.

```suggestion
func (db *DB) ResetMiniHeaders(newMiniHeaders []*types.MiniHeader) (err error) {
```

ethereum/blockwatch/block_watcher.go
```diff
@@ -586,38 +578,17 @@ type blockRange struct {
 // blocks' logs twice.
 func (w *Watcher) getSubBlockRanges(from, to, rangeSize int) []*blockRange {
 	chunks := []*blockRange{}
-	numBlocksLeft := to - from
-	if numBlocksLeft < rangeSize {
+	for from+rangeSize-1 < to {
```
albrow, line 581: How do you feel about moving `from += rangeSize` to the post condition of the for loop? That way it guarantees we won't forget to do it.

```suggestion
	for ; from+rangeSize-1 < to; from += rangeSize {
```

jalextowle: Works for me!

ethereum/simplestack/simple_stack.go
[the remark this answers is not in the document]
z2trillion: Ok, I hadn't considered that use case for this.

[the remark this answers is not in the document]
jalextowle: I only thought about it after you raised the question. We could hypothetically avoid this by merging the stack library with the blockwatcher, but I think it's better if we can swap out stack implementations easily in the future.

ethereum/blockwatch/block_watcher.go
jalextowle: No worries, this happens to me all the time. I'll check out the deadcode. If the event length is 0, then we really don't want to revert changes, so I want to be careful that this change wouldn't introduce problems if the function was re-used elsewhere.

jalextowle: I think I'm going to leave this in for now. Let's address this in a refactor, which may entail completely eliminating this function.

My rationale is that removing the deadcode makes the function more context specific (you have to check `len(events) == 0` elsewhere for it to work properly). This is okay in this case because it's only called once, but I don't really like that since we are encapsulating logic in a function that should be reusable.

Issue #854, opened 2020-07-01
GraphQL API Draft Specification


albrow:
# 0x Mesh GraphQL API Specification

This is a proposal for a new [GraphQL](https://graphql.org/) API for Mesh which will replace
the existing JSON-RPC API in Mesh version 10.0.0.

## Background and Motivation

There are several motivating factors for why we are introducing a new API:

### Improve Queryability of Mesh

Currently, Mesh does not offer much in the way of querying or filtering orders in its JSON-RPC API. Part
of the reason for this is that JSON-RPC does not lend itself well to complicated queries.

Because of this, if you need to query orders, you will need to follow a somewhat complicated process
for [syncing Mesh with your own database](https://0x-org.gitbook.io/mesh/v/release%2F9.4.0/advanced-topics/db_syncing)
and then running queries from there. Practically speaking, there is no other way to get only the orders
you care about and make sure they are up-to-date. This can create an unnecessary burden for developers, especially
in the browser where there are limited database options.

The new API will make it possible for many use cases to avoid syncing with their own database entirely.
Instead, Mesh can be thought of as the database for orders. For example, you can query Mesh directly to
get all ETH-DAI orders that expire no sooner than 60 seconds from now while filtering out dust orders.

Some users may still wish to sync their database (for example, if they want to do SQL joins with
application-specific data), but it is an _option_ rather than a _requirement_. Even for users that do want to
sync their own database, improved querying capabilities at the API layer can still make it easier to do so by
filtering out orders you don't care about earlier on in the pipeline.

### Unify and Simplify our Tooling

In the current 0x tech stack, there are many different APIs and libraries that effectively do the same
thing: get a set of orders that can be filled by anyone.

1. [SRA endpoints in 0x API](https://0x.org/docs/api#get-srav3orders)
2. [Mesh JSON-RPC API](https://0x-org.gitbook.io/mesh/getting-started/rpc_api)
3. [`@0x/mesh-browser`](https://www.npmjs.com/package/@0x/mesh-browser)
4. [`@0x/mesh-rpc-client`](https://www.npmjs.com/package/@0x/mesh-rpc-client)
5. [`@0x/connect`](https://www.npmjs.com/package/@0x/connect)
6. [`@0x/orderbook`](https://www.npmjs.com/package/@0x/orderbook)

All of these APIs and tools work differently (sometimes in subtle ways) and generally one cannot be used
as a drop in replacement for the other. Some of them are compatible with one another and others are not.
[...]

PR #858, opened 2020-07-09
Implement GraphQL API

albrow:
This is a WIP PR that implements the GraphQL API as described in https://github.com/0xProject/0x-mesh/issues/854.

### Major tasks:

- [x] Remove JSON-RPC API
- [x] Server Implementation
  - [x] `order` query
  - [x] `orders` query
  - [x] `stats` query
  - [x] `addOrders` mutation
  - [x] `orderEvents` subscription
- [x] Go client
  - [x] `AddOrders`
  - [x] `GetOrders`
  - [x] `GetOrder`
  - [x] `GetStats`
  - [x] `RawQuery` for arbitrary GraphQL queries.
- [x] TypeScript client
  - [x] `addOrdersAsync`
  - [x] `getOrdersAsync`
  - [x] `getOrderAsync`
  - [x] `getStatsAsync`
  - [x] `onOrderEvents`
    - [ ] ~~Unsubscribe?~~ Moving this to https://github.com/0xProject/0x-mesh/issues/902.
    - [x] Better WebSocket error handling
  - [x] `rawQueryAsync` for arbitrary GraphQL queries.
- [x] Testing
  - [x] Marshaling/Unmarshaling JSON and GraphQL
  - [x] Integration tests with Go client
  - [x] Integration tests with TypeScript client
- [x] Configuration (environment variables)
- [x] Update or remove code that depended on JSON-RPC API
  - [ ] ~~__examples/add_order__~~ Decided to remove this as @jalextowle and I feel it doesn't really add a lot of value and can create confusion because it doesn't set up the required balances/allowances.
  - [ ] ~~__examples/subscribe_to_orders__~~ The Go client doesn't currently support subscriptions. We can add support for this later on based on demand.
  - [ ] ~~__cmd/mesh-bridge__~~ Decided not to update the bridge in this PR. We can update it in the future if needed.
  - [x] Browser integration tests
- [x] Update Dockerfiles
- [x] Documentation
- [ ] ~~Type de-duplication? (Maybe address in follow-up PR)~~ Moving this to https://github.com/0xProject/0x-mesh/issues/900.
- [ ] ~~Improve subscription support in GraphiQL page? (Maybe address in follow-up PR)~~ Moving this to https://github.com/0xProject/0x-mesh/issues/903


albrow:
## Overview

It might be helpful to read up on some [background information on GraphQL](https://graphql.org/).

This PR implements a GraphQL server and two clients: Go and TypeScript. Both clients are designed to be usable by someone with no prior GraphQL experience, and therefore makes some assumptions about the kinds of queries you want to run (especially the shape of the data that will be returned). Users who are more familiar with GraphQL and want maximum flexibility may choose to create or generate their own clients.

The server uses https://github.com/99designs/gqlgen, which automatically generates some code for us based on the GraphQL schema. If we make any changes to the schema, we will need to run `go generate ./...` and then update the server code as needed. The relevant make tasks (e.g. `build`, `test-all`) will automatically run `go generate ./...`. See https://blog.golang.org/generate for more information about the `go generate` command.

The Go client uses https://github.com/machinebox/graphql, which is a low-level client library that doesn't generate anything for us. However, we are able to re-use some of the generated types from https://github.com/99designs/gqlgen. Notably, the Go client does not support subscriptions because the library we're using doesn't support them. We can add support for subscriptions based on demand.

The TypeScript client uses https://www.apollographql.com/docs/react/api/core/ApolloClient/. Apollo Client has a lot of interesting and powerful features when used in conjunction with React, but for the TypeScript client we are just handling the basics. The TypeScript client does support subscriptions.

## Files and directories

- __graphql/__ Contains the GraphQL schema, server, and Go code.
    - __client/___ The Go client.
    - __generated/__ Code generated by https://github.com/99designs/gqlgen. Should not be edited directly.
    - __gqltypes/__ Go types which map 1-to-1 with types defined in the GraphQL schema. Mix of generated and manually-written code.
    - __schema.graphql__ The GraphQL schema.
    - __schema.resolvers.go__ The resolver implementations used by the server (i.e. functions that define how each query is handled).
- __packages/__
    - __mesh-graphql-client/__ The TypeScript client, including integration tests.

albrow:
> two

packages/mesh-graphql-client/src/index.ts
```diff
@@ -0,0 +1,372 @@
...
+                const definition = getMainDefinition(query);
+                return definition.kind === 'OperationDefinition' && definition.operation === 'subscription';
+            },
+            wsLink,
+            httpLink,
+        );
+        const errorLink = onError(({ graphQLErrors, networkError }) => {
+            if (graphQLErrors != null && graphQLErrors.length > 0) {
+                // Throw the first error.
+                // TODO(albrow): Is there a clean way to include all the errors?
```
jalextowle, line 272: Is there any reason why we couldn't just create a table with the format that you use below for the first error?

packages/mesh-graphql-client/src/index.ts
```diff
@@ -0,0 +1,372 @@
...
+        return resp.data.orders.map(fromStringifiedOrderWithMetadata);
+    }
+
+    public onOrderEvents(): Observable<OrderEvent[]> {
+        const observable = this._client.subscribe({
+            query: orderEventsSubscription,
+        }) as Observable<FetchResult<OrderEventResponse>>;
+        return observable.map(result => {
+            if (result.errors != null && result.errors.length > 0) {
+                throw new Error(result.errors[0].message);
```
jalextowle, line 358: Let's add a `TODO` here to handle all of the errors rather than just the first.

packages/mesh-graphql-client/src/types.ts
```diff
@@ -0,0 +1,402 @@
...
+    // A human-readable message indicating why the order was rejected. This message may change
+    // in future releases and is not covered by backwards-compatibility guarantees.
+    message: string;
+}
+
+export enum RejectedOrderCode {
+    EthRpcRequestFailed = 'ETH_RPC_REQUEST_FAILED',
+    CoordinatorRequestFailed = 'COORDINATOR_REQUEST_FAILED',
+    CoordinatorSoftCancelled = 'COORDINATOR_SOFT_CANCELLED',
+    CoordinatorEndpointNotFound = 'COORDINATOR_ENDPOINT_NOT_FOUND',
```
jalextowle, line 86: If we choose to rip out coordinator support, let's add a `TODO` here.

packages/mesh-graphql-client/src/types.ts
```diff
@@ -0,0 +1,402 @@
+import { SignedOrder } from '@0x/types';
```
jalextowle, line 1: Some of the types in this file are duplicated from `packages/mesh-browser-lite/src/types.ts`. This is not a merge blocker for me, but I'd for us to try to consolidate these types eventually, unless there is a reason why you want to keep them separated.

packages/mesh-graphql-client/src/types.ts
```diff
@@ -0,0 +1,402 @@
...
+    };
+}
+
+/**
+ * Converts StringifiedSignedOrder to SignedOrder
+ */
+export function fromStringifiedSignedOrder(order: StringifiedSignedOrder): SignedOrder {
+    return {
+        ...order,
+        // tslint:disable-next-line: custom-no-magic-numbers
```
jalextowle, line 338: I'd be okay with getting rid of this linter rule entirely.

packages/mesh-graphql-client/test/graphql_client_test.ts
```diff
@@ -0,0 +1,424 @@
...
+            // HACK(jalextowle): We can't currently specify an out of process provider for a blockchainTests
+            // suite, so we need to update env.blockchainLifecycle so that the resets suite works as expected.
+            env.blockchainLifecycle = new BlockchainLifecycle(new Web3Wrapper(provider));
+
+            exchangeAddress = getContractAddressesForChainOrThrow(chainId).exchange;
+            exchange = new ExchangeContract(exchangeAddress, provider);
+            const erc20ProxyAddress = getContractAddressesForChainOrThrow(chainId).erc20Proxy;
+
+            // Configure two tokens and an order factory with a maker address so
+            // that valid orders can be created easily in the tests.
```
jalextowle, line 66: I realize that this is a comment that I wrote, but I think that it would be good to specify that the addresses used to create the `makerToken` and `feeToken` are the addresses of the dummy tokens deployed in our default ganache snapshot.

packages/mesh-graphql-client/test/graphql_client_test.ts
```diff
@@ -0,0 +1,424 @@
...
+        describe('#getStatsAsync', () => {
+            it('Ensure that the stats are correct when no orders have been added', async () => {
+                const stats = await deployment.client.getStatsAsync();
+
+                // NOTE(jalextowle): Ensure that the latest block of the returned
+                // stats is valid and then clear the field since we don't know
+                // the block number of the stats in this test a priori.
+                expect(stats.latestBlock).to.not.be.undefined();
+                // tslint:disable-next-line:no-non-null-assertion
+                // expect(stats.latestBlock!.number).to.be.greaterThan(0);
```
jalextowle, line 143: Is there a reason why this is commented out?

packages/mesh-graphql-client/test/graphql_client_test.ts
```diff
@@ -0,0 +1,424 @@
...
+                    ...order,
+                    hash: orderHashUtils.getOrderHashHex(order),
+                    fillableTakerAssetAmount: order.takerAssetAmount,
+                }));
+                expectContainsOrders(gotOrders, expectedOrders);
+            });
+            it('returns orders that match a given query', async () => {
+                const ordersLength = 10;
+                const orders = [];
+                // Create some orders with makerAssetamount = 1, 2, 3, etc.
```
jalextowle, line 220: Nit: `makerAssetamount` => `makerAssetAmount`

cmd/mesh/main.go
```diff
@@ -60,41 +61,14 @@ func main() {
...
-			selectedRPCAddr, err := waitForSelectedAddress(ctx, rpcServer)
-			if err != nil {
-				log.WithError(err).Warn("HTTP RPC server did not start")
-			}
-			log.WithField("address", selectedRPCAddr).Info("started HTTP RPC server")
-		}()
-		if err := rpcServer.Listen(ctx, rpc.HTTPHandler); err != nil {
-			httpRPCErrChan <- err
+		log.WithField("graphql_server_addr", config.GraphQLServerAddr).Info("starting GraphQL server")
+		if err := serveGraphQL(ctx, app, config.GraphQLServerAddr, config.EnableGraphiQL); err != nil {
```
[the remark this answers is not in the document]
albrow: As far as I can tell there is no way to actually confirm when the GraphQL server has started and is listening for incoming connections. `server.ListenAndServe()` blocks and only returns if there is an error. The best we can do is log right before calling `server.ListenAndServe()` but that's not too far off what we're doing right now anyways. Let me know if you have any specific ideas here.

integration-tests/browser_integration_test.go
```diff
@@ -96,14 +95,14 @@ func TestBrowserIntegration(t *testing.T) {
 	go func() {
 		defer wg.Done()
 		// Wait for the WS RPC server to start before sending the order.
-		_, err := waitForLogSubstring(ctx, standaloneLogMessages, "started WS RPC server")
-		require.NoError(t, err, "WS RPC server didn't start")
-		rpcClient, err := rpc.NewClient(standaloneWSRPCEndpointPrefix + strconv.Itoa(wsRPCPort+count))
+		_, err := waitForLogSubstring(ctx, standaloneLogMessages, "starting GraphQL server")
```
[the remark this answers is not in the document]
albrow: See https://github.com/0xProject/0x-mesh/pull/858#discussion_r467307005

graphql/client/types.go
```diff
@@ -0,0 +1,288 @@
...
+	TakerAssetData        []byte         `json:"takerAssetData"`
+	TakerAssetAmount      *big.Int       `json:"takerAssetAmount"`
+	TakerFeeAssetData     []byte         `json:"takerFeeAssetData"`
+	TakerFee              *big.Int       `json:"takerFee"`
+	SenderAddress         common.Address `json:"senderAddress"`
+	FeeRecipientAddress   common.Address `json:"feeRecipientAddress"`
+	ExpirationTimeSeconds *big.Int       `json:"expirationTimeSeconds"`
+	Salt                  *big.Int       `json:"salt"`
+	Signature             []byte         `json:"signature"`
+}
```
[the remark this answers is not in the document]
albrow: Good catch. We need both `Order` and `NewOrder` in the GraphQL schema itself because GraphQL requires that any type used in mutations must be an `input` type. That also carries over to the generated server code. However, we don't actually need two distinct types in the client code.

packages/mesh-graphql-client/src/index.ts
```diff
@@ -0,0 +1,372 @@
...
+                isRemoved
+                address
+                kind
+                parameters
+            }
+        }
+    }
+`;
+
+export class MeshGraphQLClient {
```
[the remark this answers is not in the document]
albrow: This is related to the "Better WebSocket error handling" task in the description of the PR. At the time you were reviewing, it was still a TODO item, but has now been addressed and the related commented out code removed (or uncommented and used).

packages/mesh-graphql-client/package.json
```diff
@@ -0,0 +1,58 @@
+{
```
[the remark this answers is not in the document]
albrow: @jalextowle do you still need to do this?

packages/mesh-graphql-client/src/index.ts
albrow: A table? What do you mean by that in the context of a JavaScript error?

The key word here is "clean". We can just append all the messages together in a single string but I was trying to think if there was a better way to do it. Any ideas?

packages/mesh-graphql-client/src/types.ts
albrow: I want to address this separately. See https://github.com/0xProject/0x-mesh/issues/859. It's true that we might be able to use some of the types from __packages/mesh-browser-lite/src/types.ts__ here. But we can't use all of them. There are some subtle differences in how some of the types are used in the context of GraphQL.

I know that this PR is leaving us in a (temporarily) worse place in regards to type duplication and bulky conversion code, but I have a plan for how to address this across the entire codebase. Will comment on https://github.com/0xProject/0x-mesh/issues/859 to elaborate soon.

packages/mesh-graphql-client/src/types.ts
albrow: We can't do that easily because of how we pull in the linter config from `@0x/tslint-config`. I say we just keep it. This also helps keep our code here _sort of_ in the same style as the monorepo.

albrow: Alternatively if you want to remove this rule from `@0x/tslint-config` you should ask what other people think.

cmd/mesh/main.go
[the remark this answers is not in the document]
jalextowle: Hmm, I don't have any ideas that are better than simply sleeping. We could "ping" the server every few milliseconds on startup, but I don't want to do that just for a better log. Let's just leave this as is.

packages/mesh-graphql-client/src/types.ts
jalextowle: Sounds good to me.

packages/mesh-graphql-client/src/types.ts
jalextowle: It's actually relatively simple. We have disabled this within the `contracts/` directory of the monorepo using this `tslint.json`: https://github.com/0xProject/0x-monorepo/blob/development/contracts/exchange/tslint.json. 

I'm okay with not making this change in this PR, but I think we should disable `custom-no-magic-numbers` eventually since it doesn't correlate with acceptable use of magic numbers.

packages/mesh-graphql-client/src/index.ts
jalextowle: I was thinking of creating a "table" by appending the messages together in a single string with new-lines. This may not be clean in practice, but I think that would be better if it formats well in the browser.

packages/mesh-graphql-client/package.json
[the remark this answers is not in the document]
jalextowle: Yes, I do. I'll audit them today (or find a command that can audit them for us).

docs/deployment.md
```diff
@@ -61,7 +82,6 @@ struct. They are copied here for convenience, although the source code is
 authoritative.
 
 ```go
-type Config struct {
```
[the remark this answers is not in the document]
albrow: This was not intentional. I added the struct back in.

integration-tests/browser_integration_test.go
```diff
@@ -90,20 +89,20 @@ func TestBrowserIntegration(t *testing.T) {
...
 	go func() {
 		defer wg.Done()
-		// Wait for the WS RPC server to start before sending the order.
-		_, err := waitForLogSubstring(ctx, standaloneLogMessages, "started WS RPC server")
-		require.NoError(t, err, "WS RPC server didn't start")
-		rpcClient, err := rpc.NewClient(standaloneWSRPCEndpointPrefix + strconv.Itoa(wsRPCPort+count))
+		// Wait for the GraphQL server to start before sending the order.
+		_, err := waitForLogSubstring(ctx, standaloneLogMessages, "starting GraphQL server")
+		require.NoError(t, err, "GraphQL server didn't start")
+		graphQLClient := gqlclient.New(graphQLServerURL)
```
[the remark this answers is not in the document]
albrow: We should. Seems like we were just getting lucky before and the server was ready by the time we called `graphQLClient.AddOrders`. I added the sleep.

packages/mesh-graphql-client/src/types.ts
```diff
@@ -0,0 +1,399 @@
...
+    message: string;
+}
+
+export enum RejectedOrderCode {
+    EthRpcRequestFailed = 'ETH_RPC_REQUEST_FAILED',
+    OrderHasInvalidMakerAssetAmount = 'ORDER_HAS_INVALID_MAKER_ASSET_AMOUNT',
+    OrderHasInvalidTakerAssetAmount = 'ORDER_HAS_INVALID_TAKER_ASSET_AMOUNT',
+    OrderExpired = 'ORDER_EXPIRED',
+    OrderFullyFilled = 'ORDER_FULLY_FILLED',
+    OrderCancelled = 'ORDER_CANCELLED',
```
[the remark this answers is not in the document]
albrow: 🇬🇧 Yeah we have to pick one. In the past we were using two Ls and I don't think it make sense to change it in the public API.

graphql/schema.graphql
```diff
@@ -0,0 +1,423 @@
...
+    The remaining amount of the maker asset which has not yet been filled.
+    """
+    fillableTakerAssetAmount: BigNumber!
+}
+
+"""
+An enum containing all the order fields for which filters and/or sorting is supported.
+"""
+enum OrderField {
+    hash
```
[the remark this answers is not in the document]
albrow: It's intended to map 1-to-1 to the fields of `OrderWithMetadata`. So in TypeScript it's just `type OrderField = keyof OrderWithMetadata`. This is technically a departure from GraphQL conventions (enums should usually be ALL_CAPS_WITH_UNDERSCORES), but I think it is easier to read and easier to implement this way. What do you think?

integration-tests/graphql_integration_test.go
```diff
@@ -0,0 +1,377 @@
+// +build !js
```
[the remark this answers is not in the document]
albrow: Let's address this separately if that's okay. It would touch a few different files.

packages/mesh-graphql-client/src/index.ts
albrow: Hmm.. I'm not sure how I feel about this. I generally don't put newlines in error messages and in the Go world this isn't really a thing. IMO error messages should be short and concise. Newlines render nicely in some contexts (e.g. logging the error message and nothing else with `console.log` or `console.error`) but render worse in others (e.g., in our logging infrastructure, the error is part of a map/JSON object so it would render as `\n` not an actual line break).

graphql/schema.graphql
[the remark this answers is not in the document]
jalextowle: That works for me. Thanks for clarifying

PR #873, opened 2020-07-24
Fix database is locked error

albrow:
This PR addresses the occasional `database is locked` errors that can occur when using SQLite on some platforms. See https://github.com/mattn/go-sqlite3/issues/607 and https://github.com/mattn/go-sqlite3/pull/827 for some more information. The fix involves adding a `sync.RWMutex` to __/db/sql_implementation.go__. I also moved some code around to try and hold on to the mutex for as little time as possible for each method.

### DIFF
--- a/db/sql_implementation.go
+++ b/db/sql_implementation.go
@@ -11,6 +11,7 @@ import (
 	"math/big"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/0xProject/0x-mesh/common/types"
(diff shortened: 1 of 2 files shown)


db/sql_implementation.go
```diff
@@ -262,15 +268,25 @@ func (db *DB) migrate() error {
 	return convertErr(err)
 }
 
+// ReadWriteTransactionalContext acqires a write lock, executes the transaction, then immediately releases the lock.
```
jalextowle, line 271: Nit: `acqires` => `acquires`

Issue #871, opened 2020-07-21
Missing orders during OrderSync


pgebheim:
# Context

Please provide any relevant information about your setup
- Mesh in browser (10.0.0-unstable3)
- Mac an Linux and Windows
- Firefox 78.0.2
- Brave version 1.10.60 Chromium: 81.0.4044.138 (Official Build) unknown (64-bit)

We are testing this by using a standalone node which is the recipient of orders which then relays them through the p2p network. During order sync we see behavior where orders are partially synced, and then refreshing the page will progressively sync more orders until we have all of the orders that we knowingly posted.

One issue is I don't have good logs for this because it is running in a webworker and at the moment I'm not actually getting those logs spit out to the FF or Chrome consoles even through they should be.

Here's some screenshots detailing what has happened after there are N-thousand orders posted on an augur market.

----
### Step 1
Window on the left: I wiped the databases entirely so we're starting from a fresh state. After OrderSync is done it looks like I received a total of 16 orders (this is the only market on mainnet, all orders are of size 10 -- 160 shares = 16 orders).

The window on the right was a browser open while orders were created. Those orders are still valid.

![1](https://user-images.githubusercontent.com/86010/88105321-94122880-cb58-11ea-8a5f-5bd610b63bb7.png)
So it looks like I got 16 orders during OrderSync...

### Step 2
Now, refresh the left window and orders are streaming in from OrderSync -- So the explanation here could be that we just happened to randomly select 5 peers that didn't have orders on the previous load.
![2](https://user-images.githubusercontent.com/86010/88105308-907ea180-cb58-11ea-987a-325f148051ab.png)

Now waiting I see a bunch of orders come in, and then it stops and I receive no more with the following state which is missing > 300 orders.
![3](https://user-images.githubusercontent.com/86010/88105296-8a88c080-cb58-11ea-8f61-fa02797d9cb1.png)

### Step 3
Refresh again, I get more orders and now I have all the ones that the other browser knows about.

Oddly, for 0.56 I actually find the 270 orders that I never received in the initial browser window (on the right). Also notable, I never get those orders on that side, I've refreshed a bunch etc.

So something must be wrong during ordersync because whatever nodes its talking to has these orders -- but the local syncing just stops getting them...
![4](https://user-images.githubusercontent.com/86010/88105279-852b7600-cb58-11ea-9728-f0a513152c45.png)

### Step 4
[...]

PR #874, opened 2020-07-24
fix: Wait for high quality peers to fully respond during ordersync

jalextowle:
Partially Addresses: #871 

This PR fixes two issues:
* Ordersync sub-protocols previously considered a sync that yielded 0 orders in total a success. This was a bug, and was resolved by this PR.
* A subtle issue that was introduced by #848. Because the parallelized version of ordersync would connect to several peers and would cancel the context when we had synced with a sufficient number of extra peers, it was possible for the response of several peers with a small amount of orders to short-circuit ordersync runs with other peers that had more orders. This would lead to the node not always getting all of the orders during ordersync.

After this PR, we will only make as many parallel requests as can actually be resolved without going over `ordersyncMinPeers`. Future PRs should refine the process of determining when ordersync has completed.

### DIFF
--- a/core/core_test.go
+++ b/core/core_test.go
@@ -236,7 +236,7 @@ func runOrdersyncTestCase(t *testing.T, testCase ordersyncTestCase) func(t *test
 			defer wg.Done()
 			if err := originalNode.Start(); err != nil && err != context.Canceled {
 				// context.Canceled is expected. For any other error, fail the test.
-				require.NoError(t, err)
+				panic(err)
 			}
 		}()
 
@@ -258,7 +258,7 @@ func runOrdersyncTestCase(t *testing.T, testCase ordersyncTestCase) func(t *test
 			defer wg.Done()
 			if err := newNode.Start(); err != nil && err != context.Canceled {
 				// context.Canceled is expected. For any other error, fail the test.
-				require.NoError(t, err)
+				panic(err)
 			}
 		}()
 		<-newNode.started
--- a/core/ordersync/ordersync.go
+++ b/core/ordersync/ordersync.go
@@ -45,9 +45,6 @@ const (
 	//    approxDelay * (1 - jitter) <= actualDelay < approxDelay * (1 + jitter)
 	//
 	ordersyncJitterAmount = 0.1
-	// maxRequestPeersInParallel is the largest number of peers that `GetOrders`
-	// will try to pull orders from at once.
-	maxRequestPeersInParallel = 10
 )
 
 var (
@@ -291,7 +288,7 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
 
 		m := &sync.RWMutex{}
 		wg := &sync.WaitGroup{}
-		semaphore := make(chan struct{}, maxRequestPeersInParallel)
+		semaphore := make(chan struct{}, minPeers)
 		currentNeighbors := s.node.Neighbors()
 		shufflePeers(currentNeighbors)
 		innerCtx, cancel := context.WithCancel(ctx)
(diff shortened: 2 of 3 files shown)


core/ordersync/ordersync.go
```diff
@@ -312,22 +309,30 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
-				case semaphore <- struct{}{}:
-					defer func() { <-semaphore }()
-				}
 				if err := s.getOrdersFromPeer(innerCtx, id); err != nil {
 					log.WithFields(log.Fields{
 						"error":    err.Error(),
 						"provider": id.Pretty(),
 					}).Warn("could not get orders from peer via ordersync")
+					// NOTE(jalextowle): If ordersync has completed unsuccessfully,
+					// we release the slot of the semaphore that was beind used by
```
z2trillion, line 327: beind ->  being

core/core_test.go
```diff
@@ -236,7 +236,7 @@ func runOrdersyncTestCase(t *testing.T, testCase ordersyncTestCase) func(t *test
 			defer wg.Done()
 			if err := originalNode.Start(); err != nil && err != context.Canceled {
 				// context.Canceled is expected. For any other error, fail the test.
-				require.NoError(t, err)
+				panic(err)
```
jalextowle, line 239: I ran into this while I was in the middle of refactoring the patch to `getOrdersFromPeer`. There's a good chance we have these in several other places in the codebase.

core/ordersync/ordersync.go
```diff
@@ -449,68 +445,88 @@ func (s *Service) getOrdersFromPeer(ctx context.Context, providerID peer.ID) err
...
+
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
 		}
-
-		var rawReq *rawRequest
+		nextReq, err = s.makeOrderSyncRequest(ctx, nextReq, stream, providerID)
```
jalextowle, line 468: The diff makes this change hard to review, but this function is much cleaner now. Any naming suggestions would be welcome, as `makeOrderSyncRequest` doesn't feel quite right (although it could be worse)

core/ordersync/ordersync.go
```diff
@@ -312,65 +311,60 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
+					wg.Done()
+					<-semaphore
+				}()
 				if err := s.getOrdersFromPeer(innerCtx, id); err != nil {
 					log.WithFields(log.Fields{
 						"error":    err.Error(),
 						"provider": id.Pretty(),
 					}).Warn("could not get orders from peer via ordersync")
 				} else {
-					// TODO(albrow): Handle case where no orders were returned from this
```
jalextowle, line 332: I just decided to treat a response of 0 as an error in my implementation of `makeOrderSyncRequest`. @albrow I'd be interested to hear your thoughts on whether or not this appropriate. I didn't think it was necessary personally, but I'd be open to logging a different message in the if-statement above.

core/ordersync/ordersync.go
z2trillion: I think this is fine. The way the interface is defined, I don't think there's a great name for what the function has to do.

core/ordersync/ordersync.go
```diff
@@ -291,7 +290,7 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
 
 		m := &sync.RWMutex{}
 		wg := &sync.WaitGroup{}
-		semaphore := make(chan struct{}, maxRequestPeersInParallel)
+		semaphore := make(chan struct{}, minPeers)
 		currentNeighbors := s.node.Neighbors()
 		shufflePeers(currentNeighbors)
 		innerCtx, cancel := context.WithCancel(ctx)
```
albrow, line 296: `defer cancel()` won't run until `GetOrders` returns, but I think we want to cancel `innerCtx` after each iteration of `for len(successfullySyncedPeers) < minPeers`. Otherwise we will keep creating new contexts without canceling them.

core/ordersync/ordersync.go
```diff
@@ -291,7 +290,7 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
```
albrow, line 290: I think we could definitely benefit from some comments here explaining the purpose of `m`, `wg`, and `semaphore`. Especially the `semaphore` since it is used in a somewhat subtle way. Comments could also help clarify what are we doing in each iteration of the for loop? Under what conditions do we move on to the next iteration or exit the loop?

core/ordersync/ordersync.go
```diff
@@ -312,65 +311,60 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
 				"provider": peerID.Pretty(),
 			}).Trace("requesting orders from neighbor via ordersync")
 
+			select {
```
albrow, line 314: I think this `select` statement belongs at the beginning of the `for _, peerID := range currentNeighbors` loop rather than here. We don't want to check `len(successfullySyncedPeers)` or log `"requesting orders from neighbor via ordersync"` until _after_ a slot opens up in the `semaphore` channel.

core/ordersync/ordersync.go
```diff
@@ -449,68 +443,88 @@ func (s *Service) getOrdersFromPeer(ctx context.Context, providerID peer.ID) err
...
+	nextReq := &rawRequest{
+		Type:         TypeRequest,
+		Subprotocols: s.SupportedSubprotocols(),
+		Metadata:     nil,
+	}
+	nextReq, err = s.makeOrderSyncRequest(ctx, nextReq, stream, providerID)
+	if err != nil {
+		return err
+	}
+	if nextReq == nil {
```
albrow, line 456: This doesn't look right to me. Just because`nextReq` is nil does not mean we didn't receive any orders from the provider. A successful ordersync will always result in the last `nextReq` being nil.

core/ordersync/ordersync.go
```diff
@@ -449,68 +443,88 @@ func (s *Service) getOrdersFromPeer(ctx context.Context, providerID peer.ID) err
...
+	}
+}
 
-		if err := json.NewEncoder(stream).Encode(rawReq); err != nil {
-			s.handlePeerScoreEvent(providerID, psUnexpectedDisconnect)
-			return err
-		}
+// makeOrderSyncRequest handles the process of making an ordersync request with
+// a given subprotocol to a provider and the process of decoding the response and
+// creating the next raw request (if applicable).
```
albrow, line 478: nit: Why not just:
```suggestion
// makeOrderSyncRequest sends an ordersync request with the given subprotocol
// to the provider, decodes the response, and returns the next raw request (if applicable).
```

core/ordersync/ordersync.go
```diff
@@ -314,63 +331,52 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
...
-				}
+				defer func() {
+					wg.Done()
+					<-semaphore
+				}()
 				if err := s.getOrdersFromPeer(innerCtx, id); err != nil {
 					log.WithFields(log.Fields{
 						"error":    err.Error(),
 						"provider": id.Pretty(),
 					}).Warn("could not get orders from peer via ordersync")
```
jalextowle, line 342: Nit: `Warn` => `Debug`

PR #875, opened 2020-07-26
fix: Mark an ordersync that yields zero orders unsuccessful

jalextowle:
Prior to this PR, a sync would be marked as a success provided that there was a nonzero number of orders returned in an ordersync response. Upon closer inspection, this was not strict enough since these orders may not pass the `orderfilter`. In this PR, we addressed the issue by keeping track of how many orders that match the filter have been contributed by a given ordersync peer.

### DIFF
--- a/core/ordersync/ordersync.go
+++ b/core/ordersync/ordersync.go
@@ -144,7 +144,7 @@ type Subprotocol interface {
 	// should be sent. If nextRequest is nil, the ordersync protocol is
 	// considered finished. HandleOrderSyncResponse is the implementation for the
 	// "requester" side of the subprotocol.
-	HandleOrderSyncResponse(context.Context, *Response) (nextRequest *Request, err error)
+	HandleOrderSyncResponse(context.Context, *Response) (nextRequest *Request, filteredOrders []*zeroex.SignedOrder, err error)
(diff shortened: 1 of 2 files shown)


core/ordersync/ordersync.go
```diff
@@ -144,7 +144,7 @@ type Subprotocol interface {
 	// should be sent. If nextRequest is nil, the ordersync protocol is
 	// considered finished. HandleOrderSyncResponse is the implementation for the
 	// "requester" side of the subprotocol.
-	HandleOrderSyncResponse(context.Context, *Response) (nextRequest *Request, err error)
+	HandleOrderSyncResponse(context.Context, *Response) (nextRequest *Request, filteredOrders []*zeroex.SignedOrder, err error)
```
albrow, line 147: Looks like we don't actually need the orders themselves here, just the number. Also how do you feel about calling it `numValidOrders` instead? Not all subprotocols in the future will need to "filter" the response from the provider (in fact, ideally none of them would).

```suggestion
	HandleOrderSyncResponse(context.Context, *Response) (nextRequest *Request, numFilteredOrders int, err error)
```

albrow: Can you also add this new requirement to the documentation for the `Subprotocol` interface?

PR #876, opened 2020-07-26
fix: update drone configuration

opaolini:
This PR adds the new `.drone.yml` configuration for new runner.


albrow:
I don't really know any way to test this other than merging it and seeing what happens in the next PR. Okay with you @opaolini and @jalextowle?

PR #877, opened 2020-07-27
Added the ability to track the progress of ordersyncing with peers

jalextowle:

### DIFF
--- a/core/ordersync/ordersync.go
+++ b/core/ordersync/ordersync.go
@@ -282,6 +282,11 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
 		Factor: 2,                      // Factor to multiple each successive back-off
 	}
 
+	// nextRequestForPeer tracks the last meaningful "next request" that was
+	// provided by a peer during ordersync. This allows us to pick up where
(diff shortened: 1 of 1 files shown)


core/ordersync/ordersync.go
```diff
@@ -445,46 +457,52 @@ func parseResponseWithSubprotocol(subprotocol Subprotocol, providerID peer.ID, r
 	}, nil
 }
 
-func (s *Service) getOrdersFromPeer(ctx context.Context, providerID peer.ID) error {
+func (s *Service) getOrdersFromPeer(ctx context.Context, providerID peer.ID, nextRequest *rawRequest) (*rawRequest, error) {
```
albrow, line 460: Maybe in the scope of the `getOrdersFromPeer` function we should name this `firstRequest`? It represents the first request that `getOrdersFromPeer` will send to the provider. Subsequent requests are determined by the subprotocol (i.e. the return value of `HandleOrderSyncResponse`).

```suggestion
func (s *Service) getOrdersFromPeer(ctx context.Context, providerID peer.ID, firstRequest *rawRequest) (*rawRequest, error) {
```

albrow:
Left one small comment. Looks pretty good overall.

PR #879, opened 2020-07-27
Made `MaxBytesPerSecond` configurable for standalone nodes

jalextowle:

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -192,8 +192,8 @@ type Config struct {
 	// provided, EthereumRPCURL will be ignored.
 	EthereumRPCClient ethclient.RPCClient `envvar:"-"`
 	// MaxBytesPerSecond is the maximum number of bytes per second that a peer is
-	// allowed to send before failing the bandwidth check. Defaults to 1 MiB.
-	MaxBytesPerSecond float64 `envvar:"MAX_BYTES_PER_SECOND" default:"1048576"`
+	// allowed to send before failing the bandwidth check. Defaults to 5 MiB.
+	MaxBytesPerSecond float64 `envvar:"MAX_BYTES_PER_SECOND" default:"5242880"`
 }
 
 type App struct {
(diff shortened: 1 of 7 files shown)


core/core.go
```diff
@@ -191,6 +191,10 @@ type Config struct {
 	// settable in browsers and cannot be set via environment variable. If
 	// provided, EthereumRPCURL will be ignored.
 	EthereumRPCClient ethclient.RPCClient `envvar:"-"`
+	// MaxBytesPerSecond is the maximum number of bytes per second that a peer is
+	// allowed to send before failing the bandwidth check. Defaults to 1 MiB, which
+	// is roughly 100x expected usage based on real world measurements.
```
albrow, line 196: We should remove the "is roughly 100x expected usage based on real world measurements" since that no longer appears to be true.

p2p/node.go
```diff
@@ -193,6 +196,9 @@ func New(ctx context.Context, config Config) (*Node, error) {
 	if config.PerPeerPubSubMessageBurst == 0 {
 		config.PerPeerPubSubMessageBurst = defaultPerPeerPubSubMessageBurst
 	}
+	if config.MaxBytesPerSecond == 0 {
+		config.MaxBytesPerSecond = defaultMaxBytesPerSecond
```
jalextowle, line 200: @albrow Should we use a different default, or do you think that 1 MB still makes sense?

albrow: Let's go ahead and bump it to 5MB.

packages/mesh-browser-lite/src/types.ts
```diff
@@ -186,6 +186,9 @@ export interface Config {
     // Offers the ability to use your own web3 provider for all Ethereum RPC
     // requests instead of the default.
     web3Provider?: SupportedProvider;
+    // The maximum number of bytes per second that a peer is allowed to send before
+    // failing the bandwidth check. Defaults to 1 MiB.
```
albrow, line 190: Can you double check to make sure there aren't any other outdated comments?

```suggestion
    // failing the bandwidth check. Defaults to 5 MiB.
```

PR #882, opened 2020-07-27
Fixed bug in ordersync that caused nodes to use the wrong orderfilter

jalextowle:

### DIFF
--- a/core/ordersync/ordersync.go
+++ b/core/ordersync/ordersync.go
@@ -14,9 +14,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/0xProject/0x-mesh/orderfilter"
 	"github.com/0xProject/0x-mesh/p2p"
 	"github.com/0xProject/0x-mesh/zeroex"
 	"github.com/albrow/stringset"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/jpillora/backoff"
 	network "github.com/libp2p/go-libp2p-core/network"
 	protocol "github.com/libp2p/go-libp2p-core/protocol"
@@ -112,6 +114,7 @@ type rawResponse struct {
 // responding to and sending ordersync requests.
 type Service struct {
 	ctx          context.Context
+	orderFilter  *orderfilter.Filter
 	node         *p2p.Node
 	subprotocols map[string]Subprotocol
 	// requestRateLimiter is a rate limiter for incoming ordersync requests. It's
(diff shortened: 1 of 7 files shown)


core/ordersync/ordersync.go
```diff
@@ -458,6 +462,31 @@ func parseResponseWithSubprotocol(subprotocol Subprotocol, providerID peer.ID, r
...
 }
 
+type requestMetadataForAllSubprotocols struct {
+	OrderFilter  *orderfilter.Filter `json:"orderfilter"`
+	Page         int                 `json:"page"`
+	SnapshotID   string              `json:"snapshotID"`
+	MinOrderHash common.Hash         `json:"minOrderHash"`
+}
+
+func (s *Service) createFirstRequestForAllSubprotocols() (*rawRequest, error) {
```
albrow, line 472: I don't think it is a good idea to combine data for each subprotocol into a single struct/object. It makes it harder to introduce new subprotocols in the future in a forwards-compatible way (e.g. because two different subprotocols can no longer use the same field name in their metadata without having the same semantics). Can we separate this by subprotocol and ideally have each subprotocol implementation be responsible for generating its own first request?

core/ordersync/ordersync.go
```diff
@@ -159,13 +162,14 @@ type Subprotocol interface {
 // them. New expects an array of subprotocols which the service will support, in the
 // order of preference. The service will automatically pick the most preferred protocol
 // that is supported by both peers for each request/response.
-func New(ctx context.Context, node *p2p.Node, subprotocols []Subprotocol) *Service {
+func New(ctx context.Context, orderfilter *orderfilter.Filter, node *p2p.Node, subprotocols []Subprotocol) *Service {
```
albrow, line 165: If possible, I would prefer not to pass in `orderfilter` here. Future subprotocols might not use filters in the same way or at all.

For example, in the very near future we could implement a subprotocol for getting orders for a specific market. A filter is an inefficient way to do that because it requires getting _all_ orders from the database in memory and checking if each one matches the filter. It would be more optimal to pass in the market asset data because that can be checked directly against the database.

Can we pass in the filter to each subprotocol that uses it instead? Currently that would be both `FilteredPaginationSubProtocolV0` and `FilteredPaginationSubProtocolV1`.

core/ordersync/ordersync.go
jalextowle: Good point. I was able to side-step this issue by leveraging some of the properties of the json.Unmarshal library relating to arrays.

core/ordersync/ordersync.go
```diff
@@ -458,6 +468,41 @@ func parseResponseWithSubprotocol(subprotocol Subprotocol, providerID peer.ID, r
...
 
+type FirstRequestsForSubprotocols struct {
+	MetadataForSubprotocol []json.RawMessage `json:"metadata"`
+}
+
+// createFirstRequestForAllSubprotocols creates an initial ordersync request that
+// contains metadata for all of the ordersync subprotocols. This allows more guarantees
+// to be made about the types of orders that will be received from a given ordersync
+// subprotocol, and these guarantees allow ordersync connections to be pruned much
+// more aggressively.
```
albrow, line 479:
```suggestion
// createFirstRequestForAllSubprotocols creates an initial ordersync request that
// contains metadata for all of the ordersync subprotocols.
```

nit: I know what you're talking about because I have additional context from our conversations in Slack, but I don't think you've provided enough context in the comment itself. It's not clear how creating an initial request allows us to prune connections more aggressively, and that pruning is not implemented anywhere yet. Are you okay with removing this part of the comment for now? When we do implement more aggressive connection pruning, I think this comment (or something like it) belongs at the point in the code where pruning occurs.

PR #888, opened 2020-07-28
Fixed issue that crashed Mesh node

jalextowle:


core/ordersync/ordersync.go
```diff
@@ -241,8 +241,19 @@ func (s *Service) HandleStream(stream network.Stream) {
...
-			if err == nil {
+
+			// NOTE(jalextowle): Older versions of Mesh did not include
+			// metadata in the first ordersync request. Ensure that
+			// the array is long enough before accessing it to avoid
+			// crashing the node.
+			if err == nil && len(firstRequests.MetadataForSubprotocol) > i {
 				rawReq.Metadata = firstRequests.MetadataForSubprotocol[i]
+			} else {
+				log.WithFields(log.Fields{
```
albrow, line 252: We should create the default metadata here instead of returning an error. Otherwise we are breaking backwards compatibility.

core/ordersync/ordersync.go
```diff
@@ -320,6 +329,10 @@ func (s *Service) GetOrders(ctx context.Context, minPeers int) error {
 		innerCtx, cancel := context.WithCancel(ctx)
 		defer cancel()
 		for _, peerID := range currentNeighbors {
+			if peerID.Pretty() != "16Uiu2HAmE9YZXp5zAwvexZGAKqfQyUSnRrAQ4sZ8jrxiMxJQ2uU1" {
```
albrow, line 332: I don't think you intended to leave this here?

core/ordersync/ordersync.go
```diff
@@ -111,9 +111,10 @@ type rawResponse struct {
 // Service is the main entrypoint for running the ordersync protocol. It handles
 // responding to and sending ordersync requests.
 type Service struct {
-	ctx          context.Context
-	node         *p2p.Node
-	subprotocols map[string]Subprotocol
+	ctx            context.Context
+	node           *p2p.Node
+	subprotocols   []string
```
albrow, line 116:
```suggestion
        // preferredSubprotocols is the list of supported subprotocol IDs in order of preference.
        // Providers will pick the first subprotocol from this list that they support.
	preferredSubprotocols   []string
```

albrow, line 116: We also need to use this instead of `SupportedSubprotocols()` when constructing the first request.

```
	return &rawRequest{
		Type:         TypeRequest,
		Subprotocols: s.SupportedSubprotocols(),
		Metadata:     encodedMetadata,
	}, nil
```

Should be:

```
	return &rawRequest{
		Type:         TypeRequest,
		Subprotocols: s.preferredSubprotocols,
		Metadata:     encodedMetadata,
	}, nil
```

core/ordersync/ordersync.go
```diff
@@ -477,7 +490,7 @@ type FirstRequestsForSubprotocols struct {
 func (s *Service) createFirstRequestForAllSubprotocols() (*rawRequest, error) {
 	metadata := []json.RawMessage{}
 	for _, subprotocolString := range s.SupportedSubprotocols() {
```
albrow, line 492: We should just iterate through `subprotocolSet` directly instead of iterating through `SupportedSubprotocols` and looking up the corresponding subprotocol in the set.

```suggestion
	for subprotocolId, subprotocol := range s.subprotocolSet {
```

core/ordersync/ordersync.go
jalextowle: I updated `SupportedSubprotocols` so that it would be identical to this. Having said this, there's no reason to do the for loop

core/ordersync/ordersync.go
jalextowle: I don't think we can do this. The metadata needs to correspond to the prioritization of the subprotocols. Wouldn't this be a random order?

albrow: 👍 You're right. I was commenting on the old version which iterated through `SupportedSubprotocols` which is also in random order. The new version is correct.

PR #890, opened 2020-07-29
test: Added more tests for ordersync

jalextowle:


core/ordersync/ordersync.go
```diff
@@ -162,6 +162,8 @@ func New(ctx context.Context, node *p2p.Node, subprotocols []Subprotocol) *Servi
 		sids = append(sids, subp.Name())
 		supportedSubprotocols[subp.Name()] = subp
 	}
+	// TODO(jalextowle): We should ensure that there were no duplicates -- there
```
jalextowle, line 165: @albrow I left this in intentionally to prompt a discussion about this. Do you agree that we should either (a) fail if there are duplicates or (b) filter them out?

orderfilter/shared_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This uses the same test cases as TestFilterValidateOrder, but it marshals and
+// then unmarshals the orderfilter before performing validation. This provides a
+// sanity check that order filters work properly after being encoded and then
+// decoded. More rigorous testing of these properties is tested in integration
+// tests of ordersync.
+func TestMarshalAndUnmarshalFilter(t *testing.T) {
```
jalextowle, line 22: @albrow I am planning to de-dupe this logic with that in the other test. I didn't want this to block the review, so I will say that the only change that was made is that we now encode and then decode the orderfilter into a new object to ensure that the decoded filter works properly (passed the same tests as a filter that has not been encoded).

orderfilter/filter_test.go
```diff
@@ -21,9 +21,31 @@ var (
 	contractAddresses                  = ethereum.GanacheAddresses
 )
 
+func generateDecodedFilter(chainID int, customOrderSchema string, contractAddresses ethereum.ContractAddresses) (*Filter, error) {
```
albrow, line 24: Can you add a comment explaining the approach used here and how `generateDecodedFilter` fits in?

.gitignore
```diff
@@ -18,3 +18,6 @@ packages/**/generated/**
 !integration-tests/data/standalone-0/keys/privKey
 integration-tests/data/standalone-*/
 integration-tests/data/bootstrap-0/p2p/
+
+core/ordersync/dht
```
albrow, line 22: Can you change the tests so that these directories aren't created in the first place? Should be as simple as changing `DATA_DIR`. Let's move things to the __/tmp__ directory if possible.

jalextowle: Yes, I can do that. I originally tried to use `/dev/null`, but `/tmp` would actually work!

core/ordersync/ordersync_test.go
```diff
@@ -18,3 +28,204 @@ func TestCalculateDelayWithJitters(t *testing.T) {
...
+	assert.Equal(t, res.Subprotocol, subp0.Name())
+	// Ensure that the correct metadata was returned
+	encodedMetadata, err = subp0.GenerateFirstRequestMetadata()
+	require.NoError(t, err)
+	assert.Equal(t, res.Metadata, json.RawMessage(encodedMetadata))
+}
+
+var _ p2p.MessageHandler = &simpleMessageHandler{}
+
+type simpleMessageHandler struct{}
```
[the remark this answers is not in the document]
jalextowle: They are not trying to mimic the filtered pagination subprotocols. My intention was just to create incredibly simple subprotocols. I went ahead and added the comments, but I haven't changed the names. Do you still think they should be changed?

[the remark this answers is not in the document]
albrow: My concern about the names `simpleOrderSyncSubprotocolV0` and `simpleOrderSyncSubprotocolV1`are: (1) the latter is not a newer version of the former and (2) their names give no information about what they do or the difference between them.

This is a nit and naming things is always a bit subjective, so I'm not going to block the PR over this.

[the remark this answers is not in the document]
jalextowle: I updated `V0` and `V1` to `0` and `1`. I'll try to give them better names.

PR #894, opened 2020-08-06
fix: Prevent blockwatcher from missing contract events on startup

jalextowle:
Credit to @albrow for finding the problem and identifying the fix.

### DIFF
--- a/ethereum/blockwatch/block_watcher_test.go
+++ b/ethereum/blockwatch/block_watcher_test.go
@@ -519,7 +519,9 @@ func setupOrderWatcher(t *testing.T, ctx context.Context, client Client) *Watche
 	require.NoError(t, err)
 	config.Client = client
 	config.DB = database
-	return New(blockRetentionLimit, config)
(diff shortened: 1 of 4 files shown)


ethereum/blockwatch/block_watcher_test.go
```diff
@@ -35,6 +36,43 @@ func dbOptions() *db.Options {
...
+	actualMiniHeaders := watcher.stack.PeekAll()
+	for _, expectedMiniHeader := range expectedMiniHeaders {
+		found := false
+		for _, actualMiniHeader := range actualMiniHeaders {
+			if expectedMiniHeader.Number.Cmp(actualMiniHeader.Number) == 0 {
+				assert.Equal(t, expectedMiniHeader, actualMiniHeader)
+				found = true
+			}
+		}
+		assert.True(t, found)
```
albrow, line 72: Can we make this failure a little more clear?

```suggestion
		assert.True(t, found, "miniHeader with hash %q was not stored in watcher.stack", expectedMiniHeader.Hash.Hex())
```

PR #896, opened 2020-08-07
feature: Reject taker addresses that are nonzero and not on the whitelist

jalextowle:


zeroex/orderwatch/order_watcher.go
```diff
@@ -132,6 +132,7 @@ func New(config Config) (*Watcher, error) {
 		blockEventsChan:            make(chan []*blockwatch.Event, 100),
 		atLeastOneBlockProcessed:   make(chan struct{}),
 		didProcessABlock:           false,
+		recentlyValidatedOrders:    []*types.OrderWithMetadata{},
```
albrow, line 135: Hmm.. why did this field show up in this PR?

PR #907, opened 2020-08-14
Persistent DHT and Peerstore

jalextowle:
## Overview
A regression was introduced by #793 that caused browser nodes to not be able to persistently store the peer and dht stores. This PR addresses this regression by leveraging the new database solution in order to implement the `ds.Batching` interface from the [`github.com/ipfs/go-datastore`](https://github.com/ipfs/go-datastore) package (reviewing this package will be useful for code review). Several developments are pertinent to code review:
- Native
     - We used [`github.com/0xProject/sql-datastore`](https://github.com/0xProject/sql-datastore) prior to this PR to provide a persistent key value store for Mesh bootstrap nodes. Part of this PR entailed rigorous testing of all of the key-value store implementations using the tests provided by the [`github.com/ipfs/go-datastore/test`](https://github.com/ipfs/go-datastore/tree/master/test) package. This testing uncovered enough issues in our sql-datastore implementation that I dropped this as a dependency in favor of using the project that it was originally forked from ([`github.com/ipfs/go-ds-sql`](https://github.com/ipfs/go-ds-ipfs)), which fixed the issues that the old implementation had. The migration to the new package made it necessary to bring the queries that were provided by the fork into the Mesh repository.
- WebAssembly
     - Unlike the native binary, we did not have an existing implementation of `ds.Batching` to use in the WebAssembly environment. In order to provide persistent storage, I ended up needing to implement this interface on top of Dexie. It is fairly expensive to wrap Golang functions so that they can be used in Javascript ([this documentation is relevant](https://golang.org/pkg/syscall/js/#FuncOf)).

## Notes
I did not add the tests for the key-value stores to CI because they take upwards of 20 minutes to execute. Since this implementation should be very stable, I am comfortable with only running the tests when changes need to be made to the implementation.


db/dexie_datastore.go
```diff
@@ -0,0 +1,202 @@
...
+	return nil
+}
+
+/// js conversions
+
+// FIXME - length checks and code dedupe
+func convertQueryToJS(q dsq.Query) js.Value {
+	jsFilters := make([]interface{}, len(q.Filters))
+	for i, filter := range q.Filters {
+		jsFilters[i] = js.FuncOf(func(this js.Value, args []js.Value) interface{} {
```
albrow, line 170: Every call to `js.FuncOf` _must_ eventually have a corresponding call to `Release`. Otherwise it creates a memory leak. Also, generally `js.FuncOf` should immediately start a new goroutine to prevent deadlock (although in some cases for very fast functions it is not necessary). See https://golang.org/pkg/syscall/js/#FuncOf

albrow: Maybe we can just do the filtering in Go land instead?

cmd/mesh-bootstrap/main.go
```diff
@@ -172,7 +172,12 @@ func main() {
 		newDHT = func(h host.Host) (routing.PeerRouting, error) {
 			var err error
 			dhtDir := getDHTDir(config)
-			kadDHT, err = p2p.NewDHT(ctx, dhtDir, h)
+			// Set up the DHT to use LevelDB.
+			store, err := leveldbStore.NewDatastore(dhtDir, nil)
+			if err != nil {
+				return nil, err
+			}
+			kadDHT, err = dht.New(ctx, h, dhtopts.Datastore(store), dhtopts.Protocols(p2p.DHTProtocolID))
```
jalextowle, line 180: We needed to do this in order to continue using leveldb in the Mesh bootstrap nodes. It would be better to use our custom database solution eventually, but this would require large refactors in the `db` package. Since production bootstrap nodes use the PostgreSQL storage solution primarily and the bootstrap node is a stable implementation, I'd prefer to do this in a separate PR.

Recmo: Oskar: The bootstrap nodes use LevelDB and it's not shared.

jalextowle: Thanks for following up on this. I'll remove PostgreSQL as a storage option since no one uses it.

packages/mesh-browser-lite/src/datastore.ts
```diff
@@ -0,0 +1,107 @@
...
+                    await this._table.put({ key: operation.key, value: operation.value });
+                } else {
+                    await this._table.delete(operation.key);
+                }
+            }
+        });
+    }
+
+    public async putAsync(key: string, value: Uint8Array): Promise<void> {
+        await this._table.put({ key, value });
```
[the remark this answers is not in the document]
jalextowle: I need to add template args to the `Dexie.Table`

db/sql_implementation.go
```diff
@@ -143,6 +153,16 @@ CREATE TABLE IF NOT EXISTS metadata (
...
+
+CREATE TABLE IF NOT EXISTS peerstore (
+	key  TEXT NOT NULL UNIQUE,
+	data BYTEA NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS dhtstore (
+	key  TEXT NOT NULL UNIQUE,
+	data BYTEA NOT NULL
+);
```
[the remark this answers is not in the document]
jalextowle: As mentioned offline, I haven't been able to find anything about hash vs b-tree indexes in sqlite. Do you know of any documentation that outlines when each type of index is created? 

It would make sense that they'd use a hash index in this case if it is supported.

packages/mesh-browser/go/jsutil/jsutil.go
```diff
@@ -7,10 +7,43 @@ package jsutil
...
+}
+
+func CopyBytesToGo(jsBytes js.Value) (bytes []byte, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = RecoverError(r)
+		}
+	}()
+
+	jsBytesLength := jsBytes.Get("length").Int()
```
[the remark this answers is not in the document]
Recmo: Not worth it. Just want to make sure it fails immediately and doesn't cause confusing problems later on. Are you sure it only accepts `Uint8Array` and not any object that has a `length` field?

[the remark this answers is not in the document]
jalextowle: According to the docs it will panic if `dst` is not of type `UintArray8`: https://golang.org/pkg/syscall/js/#CopyBytesToJS. The same is true of `CopyBytesToGo` (except that the type checking happens on `src`): https://golang.org/pkg/syscall/js/#CopyBytesToGo.

PR #918, opened 2020-09-04
Improved consistency of external APIs

jalextowle:
- Implements an `ApolloLink` subclass that allows browser nodes to support most GraphQL endpoints. Subscriptions and raw queries are excluded from this PR.
- Change `GetOrders` to `FindOrders` in several files.

## Small Changes
- Fixed a bug in the integration tests that would cause us not to fail tests due to browser errors.


packages/mesh-browser-lite/src/mesh.ts
```diff
@@ -151,8 +151,8 @@ export class Mesh {
      */
     public onError(handler: (err: Error) => void): void {
         this._errHandler = handler;
-        if (this._wrapper !== undefined) {
-            this._wrapper.onError(this._errHandler);
+        if (this.wrapper !== undefined) {
```
z2trillion, line 154: can this be shortened to `this.wrapper?.onError(this._errHandler)`?

z2trillion (approved):
I know this usage predates this PR, but on file lines that you touch, can you use the `?` operator instead of `if x !== undefined ...`?

graphql/generated/generated.go
```diff
@@ -3945,6 +3954,40 @@ func (ec *executionContext) _Stats_ethRPCRateLimitExpiredRequests(ctx context.Co
...
+	}
+	if resTmp == nil {
+		if !graphql.HasFieldError(ctx, fc) {
+			ec.Errorf(ctx, "must not be null")
+		}
+		return graphql.Null
+	}
+	res := resTmp.([]string)
+	fc.Result = res
+	return ec.marshalNString2ᚕstringᚄ(ctx, field.Selections, res)
```
Recmo, line 3988: Strange unicode characters here, but the same happens in some other functions, so not caused by this PR. Still, what's going on here?

Recmo (approved):
LGTM!

graphql/generated/generated.go
jalextowle: It's part of graphql generation. I'm not completely sure what these replacements mean and why unicode is being used, but I did find an instance where they are using these unicode characters during codegen: https://github.com/99designs/gqlgen/blob/f4669ba9b54cdda6340c6e0c16d05eff2ee4fa21/codegen/templates/templates.go#L232.

PR #919, opened 2020-09-08
[WIP] Add WASM runner and docker image

opaolini:


core/core.go
```diff
@@ -101,47 +101,47 @@ type Config struct {
 	P2PWebSocketsPort int `envvar:"P2P_WEBSOCKETS_PORT" default:"60559"`
 	// EthereumRPCURL is the URL of an Etheruem node which supports the JSON RPC
 	// API.
-	EthereumRPCURL string `envvar:"ETHEREUM_RPC_URL" json:"-"`
+	EthereumRPCURL string `envvar:"ETHEREUM_RPC_URL" json:"ethereumRPCURL"`
```
opaolini, line 104: this exports this env var by default, should probably comb through if we log it in any place explicitly

Makefile
```diff
@@ -131,3 +131,7 @@ docker-mesh-fluent-bit:
 .PHONY: docker-mesh-bridge
 docker-mesh-bridge:
 	docker build . -t 0xorg/mesh-bridge -f ./dockerfiles/mesh-bridge/Dockerfile
+
```
jalextowle, line 134: Style Nit: We usually put two newlines between commands (Alex B. introduced this convention a while ago).

dockerfiles/mesh-wasm-runner/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY --from=mesh-builder /0x-mesh/packages/runner-wasm/dist /usr/mesh/dist
+
+# Default port for TCP multiaddr
+EXPOSE 60558
+# Default port for WebSockets multiaddr
+EXPOSE 60559
+
+RUN chmod +x ./mesh-wasm-runner
+
+ENTRYPOINT ["dumb-init", "--"]
```
jalextowle, line 32: Can you add a comment that explains why we're using `dumb-init`?

jalextowle (approved):

Recmo (approved):
LGTM, but could use some high level documentation.

PR #924, opened 2020-09-15
Updates yarn.lock files and adds dependabot configuration

jalextowle:
## Description

Before this PR, we had package level `yarn.lock` files that were not getting updated, and our dependabot settings were suboptimal for our project. 

From now on, dependabot will make PRs against development and we will only have a single `yarn.lock` file at the top-level of the workspaces monorepo.


jalextowle:
Closing in favor of #925

PR #925, opened 2020-09-15
Updates yarn.lock files and adds dependabot configuration

jalextowle:
## Description

Before this PR, we had package level `yarn.lock` files that were not getting updated, and our dependabot settings were suboptimal for our project. 

From now on, dependabot will make PRs against development and we will only have a single `yarn.lock` file at the top-level of the workspaces monorepo.


.dependabot/config.yml
```diff
@@ -0,0 +1,8 @@
+version: 1
+update_configs:
+  # Keep package.json (& lockfiles) secure and up-to-date,
+  # batching pull requests as security updates arrive.
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "live"
+    target_branch: "development"
```
Recmo, line 8: Why do we have two versions of the dependabot config?

Recmo (approved):

PR #969, opened 2020-10-23
[WIP] Multiple Order Formats (Attempt 2)

jalextowle:


db/sqltypes/sqltypes.go
```diff
@@ -221,6 +221,7 @@ func (s *ParsedAssetData) Scan(value interface{}) error {
 
 // Order is the SQL database representation a 0x order along with some relevant metadata.
 type Order struct {
+	Version                  int              `db:"version"`
```
Recmo, line 224: Please remember that while we can assign numbers to the different versions, there is not going to be an obvious correspondence. Order versions could be `v3`, `v4`, `v4-rfqt`, `v4-simple`, `serum-v1`, `near-v2`,  etc...

db/sqltypes/sqltypes.go
```diff
@@ -352,6 +355,7 @@ func OrderFromCommonType(order *types.OrderWithMetadata) *Order {
 		return nil
 	}
 	return &Order{
+		Version:                  3,
```
Recmo, line 358: Where does this `3` come from?

db/sqltypes/sqltypes.go
jalextowle: 👍

PR #983, opened 2020-11-25
fix: database is locked and unhandled ABI parser errors

opaolini:


zeroex/orderwatch/decoder/event_decoder.go
```diff
@@ -609,6 +609,21 @@ func (e UntrackedTokenError) Error() string {
 	return fmt.Sprintf("event for an untracked token: contract address: %s, topic: %s", e.TokenAddress.Hex(), e.Topic.Hex())
 }
 
+// AbiParserError is thrown when the decoder fails to parse topics from a
+// retrieved log.
+// NOTE(oskar): this error occurs during abi.ParseTopics and can be a result of
+// certain tokens not conforming to the ERC20 event standard, for example not
+// using indexed log parameters.
+type AbiTopicParserError struct {
```
z2trillion, line 617: fwiw, i think `AbiParseTopicError` would be a better name, since `TopicParser` isn't the name of a struct.

PR #984, opened 2020-11-30
fix: database is locked and unhandled ABI parser errors

opaolini:

### DIFF
--- a/db/sql_implementation.go
+++ b/db/sql_implementation.go
@@ -45,7 +45,7 @@ type DB struct {
 func defaultOptions() *Options {
 	return &Options{
 		DriverName:     "sqlite3",
-		DataSourceName: "0x_mesh/db/db.sqlite",
+		DataSourceName: "0x_mesh/db/db.sqlite?_journal_mode=wal",
 		MaxOrders:      100000,
 		MaxMiniHeaders: 20,
 	}
@@ -55,7 +55,7 @@ func defaultOptions() *Options {
 func TestOptions() *Options {
 	return &Options{
 		DriverName:     "sqlite3",
-		DataSourceName: filepath.Join("/tmp", "mesh_testing", uuid.New().String(), "db.sqlite"),
+		DataSourceName: filepath.Join("/tmp", "mesh_testing", uuid.New().String(), "db.sqlite?_journal_mode=wal"),
 		MaxOrders:      100,
 		MaxMiniHeaders: 20,
 	}
--- a/db/sql_implementation.go
+++ b/db/sql_implementation.go
@@ -20,7 +20,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/ido50/sqlz"
 	ds "github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-ds-sql"
+	sqlds "github.com/ipfs/go-ds-sql"
 	"github.com/jmoiron/sqlx"
 	_ "github.com/mattn/go-sqlite3"
 )
@@ -45,7 +45,7 @@ type DB struct {
 func defaultOptions() *Options {
 	return &Options{
 		DriverName:     "sqlite3",
-		DataSourceName: "0x_mesh/db/db.sqlite?_journal_mode=wal",
+		DataSourceName: "0x_mesh/db/db.sqlite",
 		MaxOrders:      100000,
 		MaxMiniHeaders: 20,
 	}
@@ -55,7 +55,7 @@ func defaultOptions() *Options {
 func TestOptions() *Options {
 	return &Options{
 		DriverName:     "sqlite3",
-		DataSourceName: filepath.Join("/tmp", "mesh_testing", uuid.New().String(), "db.sqlite?_journal_mode=wal"),
+		DataSourceName: filepath.Join("/tmp", "mesh_testing", uuid.New().String(), "db.sqlite"),
 		MaxOrders:      100,
 		MaxMiniHeaders: 20,
 	}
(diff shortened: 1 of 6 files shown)


db/sql_implementation.go
```diff
@@ -69,25 +81,42 @@ func New(ctx context.Context, opts *Options) (*DB, error) {
...
-	if err := os.MkdirAll(filepath.Dir(opts.DataSourceName), os.ModePerm); err != nil && err != os.ErrExist {
-		return nil, err
+	pathsToCreate := []string{opts.DataSourceName, opts.DataSourceDHTName, opts.DataSourcePeerStoreName}
+	for _, path := range pathsToCreate {
+		if err := os.MkdirAll(filepath.Dir(path), os.ModePerm); err != nil && err != os.ErrExist {
+			return nil, err
+		}
 	}
 
 	sqldb, err := sqlx.ConnectContext(connectCtx, opts.DriverName, opts.DataSourceName)
```
z2trillion, line 91: Is it possible to open these three using a for loop instead?

db/sql_implementation.go
```diff
@@ -307,6 +339,8 @@ const updateMetadataQuery = `UPDATE metadata SET
 
 func (db *DB) migrate() error {
 	_, err := db.sqldb.ExecContext(db.ctx, schema)
```
z2trillion, line 341: Right now the errs for the first two `ExecContext` calls can get swallowed. can you just use a for loop for the three schemas?

zeroex/orderwatch/decoder/event_decoder_test.go
```diff
@@ -96,6 +100,28 @@ func TestDecodeERC20Approval(t *testing.T) {
 
 }
 
+func TestDecodeERC20ApprovalForNonConformingImplementation(t *testing.T) {
```
z2trillion, line 103: NonConforming should be Nonconforming i think.

zeroex/orderwatch/decoder/event_decoder_test.go
```diff
@@ -96,6 +100,28 @@ func TestDecodeERC20Approval(t *testing.T) {
...
+	var approvalLog types.Log
+	err := unmarshalLogStr(erc20ApprovalLogTest, &approvalLog)
+	if err != nil {
```
z2trillion, line 106: What's the difference between this block and `require.NoError(t, err)`?

z2trillion (approved):

db/sql_implementation.go
opaolini: Yes, to be fixed (linters also warn for this) before moving from draft :smile:

zeroex/orderwatch/decoder/event_decoder_test.go
opaolini: the latter takes just 1 line, but I just copied and tweaked over the `TestDecodeERC20Approval` one, which uses the standard `err != nil` .

I guess if this were to change we could change all the previous occurences in all tests.

db/sql_implementation.go
opaolini: We'd need either a `map[dbType]context` on the struct which would require changes to all uses of `db.sqldb` with a map getter, or add logic inside the loop to map it onto appropriate struct fields.

Since the number of contexts will most likely not change (`meshDB` , `peerstore`, `dht`) I think its fine if we keep this unrolled.

zeroex/orderwatch/order_watcher.go
```diff
@@ -73,7 +73,7 @@ type Watcher struct {
 	contractAddresses          ethereum.ContractAddresses
 	orderFeed                  event.Feed
 	orderScope                 event.SubscriptionScope // Subscription scope tracking current live listeners
-	contractAddressToSeenCount map[common.Address]uint
+	contractAddressToSeenCount *contractAddressesSeenCounter
```
Recmo, line 76: we can rename this field to `contractAddressSeenCounter` now that it is no longer a map.

zeroex/orderwatch/order_watcher.go
```diff
@@ -1959,8 +1959,22 @@ func (w *Watcher) checkDecodeErr(err error, eventType string) bool {
 		}).Warn("unsupported event found")
 		return true
 	}
+	// NOTE(oskar): What happens here is that some tokens which do not
+	// respect the ERC20 specification can throw ABI decoder errors, we
+	// should handle them here and return it as a non-critical error.
+	// TODO(oskar): Should this be handled the same way for all non-ABI
+	// conforming ERC20 implementations?
```
Recmo, line 1966: > Should this be handled the same way for all non-ABI  conforming ERC20 implementations?

Where the failures just ignored in the old behaviour? If so, the new behaviour just adds a log, which is backwards compatible in that if no-one had a problem with the old behaviour, they should be fine with the new behaviour too.

zeroex/orderwatch/order_watcher_utils.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type contractAddressesSeenCounter struct {
+	contractAddressToSeenCount map[common.Address]uint
+
+	mu sync.RWMutex
```
Recmo, line 12: This is a good place for atomic counters [`sync/atomic`](https://golang.org/pkg/sync/atomic/) instead of implementing our own using locks. They have some performance advantages under the hood, but I don't think that matters here.

Recmo: Then again, we still need a lock for map itself if entries get added, physically moved or removed.

Recmo (approved):
LGTM

PR #993, opened 2020-12-22
Release 10.2.0

opaolini:


Recmo (commented):
Currently 0x-Mesh nodes all advertise themselves with user-agent-version string`"github.com/0xProject/0x-mesh"`. I can not find where it is set, it could be a default from libp2p. But it would be useful if we added a version number here. The useragent can be set with [`libp2p.UserAgent`](https://godoc.org/github.com/libp2p/go-libp2p#UserAgent).

Edit: Yeah, seems to be default: [source](https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/id.go#L55) Polkadot uses `{impl_name}/v{impl_version}`. Ipfs uses `go-ipfs/" + CurrentVersionNumber + "/" + CurrentCommit`  [source](https://github.com/ipfs/go-ipfs/blob/2ed9254426e900cf00a9b35304dc5b5de8173208/version.go#L14)

Recmo (approved):
LGTM

PR #1007, opened 2021-01-20
OrderWatcher V4 support

Recmo:


db/sql_implementation.go
```diff
@@ -163,6 +163,21 @@ func (db *DB) ReadWriteTransactionalContext(ctx context.Context, opts *sql.TxOpt
 }
 
 func (db *DB) AddOrders(orders []*types.OrderWithMetadata) (alreadyStored []common.Hash, added []*types.OrderWithMetadata, removed []*types.OrderWithMetadata, err error) {
```
Recmo, line 165: This is backwards compatible since calling it with only V3 orders will behave as expected.

db/sql_implementation.go
```diff
@@ -434,6 +449,18 @@ func (db *DB) DeleteOrders(query *OrderQuery) (deleted []*types.OrderWithMetadat
 }
 
 func (db *DB) UpdateOrder(hash common.Hash, updateFunc func(existingOrder *types.OrderWithMetadata) (updatedOrder *types.OrderWithMetadata, err error)) (err error) {
```
Recmo, line 451: Same here

PR #1017, opened 2021-02-03
fix: graphql client V4 order query response structure

kimpers:
Fixes some minor query result format issues for v4.

### DIFF
--- a/packages/mesh-graphql-client/src/types.ts
+++ b/packages/mesh-graphql-client/src/types.ts
@@ -1,6 +1,6 @@
-import { SignatureType, SignedOrder } from '@0x/types';
-import { BigNumber } from '@0x/utils';
 import { LimitOrderFields, Signature } from '@0x/protocol-utils';
+import { SignedOrder } from '@0x/types';
+import { BigNumber } from '@0x/utils';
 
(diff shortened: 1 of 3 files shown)


packages/mesh-graphql-client/src/types.ts
```diff
@@ -339,8 +339,8 @@ export interface StringifiedRejectedOrderResult<K extends GenericStringifiedSign
 
 export interface StringifiedOrderEvent {
     timestamp: string;
-    order?: StringifiedOrderWithMetadata;
-    orderv4?: StringifiedOrderWithMetadataV4;
+    order?: StringifiedOrderWithMetadata | null;
+    orderv4?: StringifiedOrderWithMetadataV4 | null;
```
Recmo, line 343: Do we need `null` here? The `?` already implies a `| undefined`, afaik.

Recmo (approved):

packages/mesh-graphql-client/src/types.ts
kimpers: Updated to only use `null` as that is what graphql will return 👍

opaolini (approved):

PR #1019, opened 2021-02-04
Fix: crash when taker address not allowed

xianny:
Not sure what the base branch should be? help


Recmo (approved):
LGTM. Base branch is `development-v4`

Recmo:
Hmm.. maybe want to remove the dependabot commits and merge them through a separate PR, if at all.

PR #1053, opened 2021-03-29
add multiple external sources to grab public IP by mesh

opaolini:
- add an optional configuration for custom public ip sources
- update docs


xianny (approved):
looks good to me!

PR #1065, opened 2021-04-08
Update contract-addresses to allow bsc chain

xianny:


kimpers:
Nice! One thing to keep in mind is that the server side of Mesh has all contract addresses hardcoded in [ethereum/contract_addresses.go](https://github.com/0xProject/0x-mesh/blob/master/ethereum/contract_addresses.go) so changes in `@0x/contract-addresses` will not propagate there.

